PR #108, opened 2022-03-26
docs: Improved error message for schema validation failure for unsupported field type

JigarJoshi:


value/value.go
```diff
@@ -124,7 +124,7 @@ func NewValueUsingSchema(field *schema.Field, input *structpb.Value) (Value, err
 		return nil, status.Errorf(codes.InvalidArgument, "permissible type for '%s' is string only", field.FieldName)
 	}
 
-	return nil, status.Errorf(codes.InvalidArgument, "unsupported type '%T'", input.Kind)
+	return nil, status.Errorf(codes.InvalidArgument, "unsupported type for field name '%s'", field.Name())
```
efirs, line 127: Can it be `type %T for field name %s` ?
Meaning include both type and name.

efirs (approved):

value/value.go
JigarJoshi: discussed it offline and decided to go ahead with this message.

PR #111, opened 2022-03-27
Add go report card

efirs:


ovaistariq (approved):

query/filter/logical.go
```diff
@@ -67,7 +67,6 @@ func (a *AndFilter) Type() LogicalOP {
 	return AndOP
 }
 
-
 // Matches returns true if the input doc matches this filter.
```
himank, line 70: nit: do we need to remove line here?

himank (approved):

query/filter/logical.go
efirs: yes, this fixes gofmt.

PR #149, opened 2022-03-30
feat: tenant class to manage the namespaces, databases and collections

himank:
To support default or custom namespace. Have a database class to manage multiple collections.

Once we land this, the next part is to integrate this with the API layer.

### DIFF
--- a/server/metadata/encoding/dictionary.go
+++ b/server/metadata/encoding/dictionary.go
@@ -79,8 +79,8 @@ const (
 
 // subspaces
 const (
-	reservedSubspaceKey = "reserved"
-	encodingSubspaceKey = "encoding"
+	ReservedSubspaceKey = "reserved"
+	EncodingSubspaceKey = "encoding"
 )
 
 var (
@@ -95,17 +95,23 @@ var (
 type reservedSubspace struct {
 	sync.RWMutex
 
-	allocated map[uint32]string
+	allocated     map[uint32]string
+	namespaceToId map[string]uint32
 }
 
 func newReservedSubspace() *reservedSubspace {
 	return &reservedSubspace{
-		allocated: make(map[uint32]string),
+		allocated:     make(map[uint32]string),
+		namespaceToId: make(map[string]uint32),
 	}
 }
 
+func (r *reservedSubspace) getNamespaces() map[string]uint32 {
+	return r.namespaceToId
+}
+
 func (r *reservedSubspace) reload(ctx context.Context, tx transaction.Tx) error {
-	key := keys.NewKey(reservedSubspaceKey)
+	key := keys.NewKey(ReservedSubspaceKey)
 	it, err := tx.Read(ctx, key)
 	if err != nil {
 		return err
@@ -124,6 +130,7 @@ func (r *reservedSubspace) reload(ctx context.Context, tx transaction.Tx) error 
 
 		r.Lock()
 		r.allocated[ByteToUInt32(row.Value)] = allocatedTo.(string)
+		r.namespaceToId[allocatedTo.(string)] = ByteToUInt32(row.Value)
 		r.Unlock()
 	}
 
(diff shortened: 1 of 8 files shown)


server/metadata/encoding/dictionary.go
```diff
@@ -286,18 +301,123 @@ func (k *DictionaryEncoder) validCollectionId(id uint32) error {
...
+func (k *DictionaryEncoder) GetDatabases(ctx context.Context, tx transaction.Tx, namespaceId uint32) (map[string]uint32, error) {
+	databases := make(map[string]uint32)
+	it, err := tx.Read(ctx, keys.NewKey(EncodingSubspaceKey, encVersion, UInt32ToByte(namespaceId), dbKey))
+	if err != nil {
+		return nil, err
+	}
+
+	var v kv.KeyValue
+	for it.Next(&v) {
+		if it.Err() != nil {
```
efirs, line 313: This should be checked after the loop.
If `Next` returns `true` it's guaranteed that there is no error.

server/metadata/encoding/dictionary.go
```diff
@@ -286,18 +301,123 @@ func (k *DictionaryEncoder) validCollectionId(id uint32) error {
...
+			return nil, err
+		}
+		if len(v.Key) < 5 {
+			return nil, api.Errorf(codes.Internal, "not a valid key %v", v.Key)
+		}
+		switch len(v.Key) {
+		case 5:
```
efirs, line 320: This switch is not needed?
The len is checked above.

server/metadata/encoding/dictionary.go
```diff
@@ -286,18 +301,123 @@ func (k *DictionaryEncoder) validCollectionId(id uint32) error {
...
+func (k *DictionaryEncoder) GetCollections(ctx context.Context, tx transaction.Tx, namespaceId uint32, databaseId uint32) (map[string]uint32, error) {
+	var collections = make(map[string]uint32)
+	it, err := tx.Read(ctx, keys.NewKey(EncodingSubspaceKey, encVersion, UInt32ToByte(namespaceId), UInt32ToByte(databaseId)))
+	if err != nil {
+		return nil, err
+	}
+
+	var v kv.KeyValue
+	for it.Next(&v) {
+		if it.Err() != nil {
```
efirs, line 347: same as above

server/metadata/encoding/dictionary.go
```diff
@@ -286,18 +301,123 @@ func (k *DictionaryEncoder) validCollectionId(id uint32) error {
...
+			return nil, err
+		}
+		if len(v.Key) < 6 {
+			return nil, api.Errorf(codes.Internal, "not a valid key %v", v.Key)
+		}
+		switch len(v.Key) {
```
efirs, line 353: switch is not needed?

server/metadata/encoding/dictionary.go
```diff
@@ -286,18 +301,123 @@ func (k *DictionaryEncoder) validCollectionId(id uint32) error {
...
+func (k *DictionaryEncoder) GetIndexes(ctx context.Context, tx transaction.Tx, namespaceId uint32, databaseId uint32, collId uint32) (map[string]uint32, error) {
+	var indexes = make(map[string]uint32)
+	it, err := tx.Read(ctx, keys.NewKey(EncodingSubspaceKey, encVersion, UInt32ToByte(namespaceId), UInt32ToByte(databaseId), UInt32ToByte(collId)))
+	if err != nil {
+		return nil, err
+	}
+
+	var v kv.KeyValue
+	for it.Next(&v) {
+		if it.Err() != nil {
```
efirs, line 382: same as above

server/metadata/tenant_test.go
```diff
@@ -0,0 +1,216 @@
...
+		require.NoError(t, tx.Commit(context.TODO()))
+
+		_ = kv.DropTable(ctx, encoding.ReservedSubspaceKey)
+	})
+	t.Run("create_duplicate_tenant_error", func(t *testing.T) {
+		ctx := context.TODO()
+		_ = kv.DropTable(ctx, encoding.ReservedSubspaceKey)
+
+		m := NewTenantManager()
+		tx, err := tm.StartTxWithoutTracking(context.TODO())
```
efirs, line 86: nit: you can use ctx instead of context.TODO() everywhere in the tests.

server/metadata/encoding/dictionary.go
efirs: This also means that error case is not tested :)

server/metadata/tenant_test.go
himank: yeah, missed it.

server/metadata/encoding/dictionary.go
himank: done.

server/metadata/encoding/dictionary.go
himank: moved it :)

efirs (approved):

PR #165, opened 2022-04-05
refactor: refactoring query handling layer

himank:


store/kv/fdb.go
```diff
@@ -318,7 +319,7 @@ func (t *ftx) Insert(_ context.Context, table string, key Key, data []byte) erro
 		return err
 	}
 	if vv != nil {
-		return os.ErrExist
+		return api.Errorf(codes.AlreadyExists, "duplicate key value, violates unique primary key constraint")
```
efirs, line 322: I would not push dependency on API package that deep.

efirs: Can we just have kv.ErrAlreadyExists?

efirs (approved):

store/kv/fdb.go
himank: This is not permanent here. I think we need to define errors for kv package and then translate in the server package. Just keeping it as a holder.

himank: I'll also add a ToDo for this in follow up diff.

PR #167, opened 2022-04-07
feat: end-to-end schema management, support transactional schema changes. 

himank:


server/services/v1/query_runner.go
```diff
@@ -18,8 +18,12 @@ import (
 	"context"
 	"encoding/json"
 
+	"github.com/rs/zerolog/log"
```
efirs, line 21: goimports

efirs (approved):

PR #196, opened 2022-04-14
feat: support for array and object types

himank:


schema/schema_test.go
```diff
@@ -98,7 +101,82 @@ func TestCreateCollectionFromSchema(t *testing.T) {
...
+		},
+		"simple_object": {
+			"type": "object",
+			"properties": {
+				"name": { "type": "string" }
+			}
+		},
+		"product_items": {
+			"type": "array",
+			"items": {
```
efirs, line 140: Is it checked that if type is arrays then items must be set. Same if type is object then properties must be set.
Optionally we can check if properties and items is set for other types and reject, to help finding programming errors.

himank: good point, added more guards.

efirs (approved):

PR #206, opened 2022-04-20
feat: data wrapper to wrap the raw value in the underlying storage

himank:


internal/data.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (x *RawData) GetDataType() DataType {
+	return RawDataType
+}
+
+func Encode(data Data) ([]byte, error) {
+	var b []byte
+	enc := codec.NewEncoderBytes(&b, &bh)
+
+	switch data.GetDataType() {
```
efirs, line 76: What if we just pass interface to this function and:
```
switch t := data.(type) {
case *TableData:
```
So we don't need to implement any interfaces.

himank: if we don't implement the interface or enforce it then the problem would be that any object will fulfill the data interface and can be passed around.

efirs: But then encoding will fail.

himank: but that will be runtime, if we enforce then we can detect at compile time.

go.mod
```diff
@@ -69,6 +69,7 @@ require (
 	github.com/spf13/cast v1.4.1 // indirect
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/subosito/gotenv v1.2.0 // indirect
+	github.com/ugorji/go/codec v1.2.7 // indirect
 	github.com/valyala/fasthttp v1.34.0 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f // indirect
```
seeforschauer, line 74: would you please update oapi version from github.com/deepmap/oapi-codegen v1.9.1 version to v1.10.0? 

A release recently was https://github.com/deepmap/oapi-codegen/releases/tag/v1.10.1 :) 

v1.10.0 contains protobuf version update https://github.com/deepmap/oapi-codegen/pull/553

himank: sure :) probably will not couple that with this diff.

himank: done

Makefile
```diff
@@ -40,7 +41,14 @@ ${API_DIR}/client/${V}/%/http.go: ${PROTO_DIR}/%_openapi.yaml
 		-o ${API_DIR}/client/${V}/$(*F)/http.go \
 		${PROTO_DIR}/$(*F)_openapi.yaml
 
-generate: ${GEN_DIR}/api.pb.go ${GEN_DIR}/api.pb.gw.go ${GEN_DIR}/health.pb.go ${GEN_DIR}/health.pb.gw.go
+${DATA_PROTO_DIR}/%.pb.go ${DATA_PROTO_DIR}/%.pb.gw.go: ${DATA_PROTO_DIR}/%.proto
+	protoc -Iinternal \
+		--go_out=${DATA_PROTO_DIR} --go_opt=paths=source_relative \
+		--go-grpc_out=${DATA_PROTO_DIR} --go-grpc_opt=paths=source_relative \
```
efirs, line 47: We don't need go-grpc and grpc-gateway for now

server/metadata/encoding/dictionary.go
```diff
@@ -133,8 +134,8 @@ func (r *reservedSubspace) reload(ctx context.Context, tx transaction.Tx) error
 			return api.Errorf(codes.Internal, "unable to deduce the encoded key from fdb key %T", allocatedTo)
 		}
 
-		r.allocated[ByteToUInt32(row.Value)] = allocatedTo.(string)
-		r.namespaceToId[allocatedTo.(string)] = ByteToUInt32(row.Value)
+		r.allocated[row.Data.(*internal.DictEncData).EncValue] = allocatedTo.(string)
```
efirs, line 137: Type casting in multiple places doesn't look nice to me.
Can we create a helper method?
```
row.Data.ToDictEnc()
```
or 
```
GetDictEnc(row.Data)
```

efirs: And same for Table and Raw.

internal/data.go
```diff
@@ -0,0 +1,153 @@
...
+	var buf bytes.Buffer
+	err := buf.WriteByte(byte(data.GetDataType()))
+	if err != nil {
+		return nil, err
+	}
+	enc := codec.NewEncoder(&buf, &bh)
+
+	switch data.GetDataType() {
+	case TableDataType:
+		if err := enc.Encode(data.(*TableData)); err != nil {
```
efirs, line 101: If there is some bug, this may panic.
Can we `switch t := xxx.(type)` instead?

himank: sure

server/metadata/encoding/dictionary.go
himank: yeah agree - sure, let me think about it.

Makefile
himank: good point

efirs (approved):

PR #220, opened 2022-04-25
feat: added initial CDC implementation with stream API

mattayres:


cdc/listener.go
```diff
@@ -0,0 +1,102 @@
...
+		return errors.New("no publisher set")
+	}
+
+	key, err := l.pub.getNextKey()
+	if err != nil {
+		return err
+	}
+
+	data := internal.NewTableDataWithEncoding(json, internal.JsonEncoding)
+	bytes, err := internal.Encode(data)
```
himank, line 93: need to return an err from here.

cdc/manager.go
```diff
@@ -0,0 +1,64 @@
...
+	m.Lock()
+	defer m.Unlock()
+
+	if m.pubs[dbName] == nil {
+		m.pubs[dbName] = NewPublisher(dbName)
+	}
+	return m.pubs[dbName]
+}
+
+func (m *Manager) WrapContext(ctx context.Context) context.Context {
```
himank, line 47: have database name as well here and then return the listener from the corresponding publisher?

cdc/streamer.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/apple/foundationdb/bindings/go/src/fdb"
+	jsoniter "github.com/json-iterator/go"
+	"github.com/rs/zerolog/log"
+	"github.com/tigrisdata/tigrisdb/internal"
+	"github.com/tigrisdata/tigrisdb/server/config"
+)
+
+type Streamer struct {
+	db      fdb.Database
```
himank, line 28: will you be changing it to use kvstore in the next iteration?

cdc/streamer.go
```diff
@@ -0,0 +1,115 @@
...
+			return s.pub.beginKey, nil
+		}
+	})
+	if err != nil {
+		return err
+	}
+
+	s.lastKey = key.(fdb.Key)
+	s.Txs = make(chan Tx, s.cfg.StreamBuffer)
+	s.ticker = time.NewTicker(s.cfg.StreamInterval)
```
himank, line 59: once done, you need to close the ticker.

cdc/streamer.go
```diff
@@ -0,0 +1,115 @@
...
+	go func() {
+		for range s.ticker.C {
+			err := s.read()
```
himank, line 62: nit, combine it in one line `if err := s.read(); err != nil {}`

server/config/options.go
```diff
@@ -39,6 +40,13 @@ type AuthConfig struct {
 	LogOnly          bool
 }
 
+type CdcConfig struct {
```
himank, line 43: remove it from here? I think we want it to be in the outside config to avoid dependency issue?

server/services/v1/query_lifecycle.go
```diff
@@ -71,6 +76,8 @@ func (q *queryLifecycle) run(ctx context.Context, options *ReqOptions) (*Respons
 		log.Debug().Str("ns", tenant.String()).Msg("tenant found")
 	}
 
+	ctx = q.cdcMgr.WrapContext(ctx)
```
himank, line 79: do we need it here if we just have it in the runner?

cdc/listener.go
```diff
@@ -0,0 +1,102 @@
...
+	LKey  []byte `json:",omitempty"`
+	RKey  []byte `json:",omitempty"`
+	Data  []byte `json:",omitempty"`
+}
+
+func (tx *Tx) addOp(entry Op) {
+	tx.Ops = append(tx.Ops, entry)
+}
+
+type TxListener struct {
```
himank, line 42: if you do this way then we still have only one listener? Don't we need to have TxLister inside the Publisher?

cdc/manager.go
```diff
@@ -0,0 +1,64 @@
...
+	l := kv.Listener(nil)
+
+	if config.DefaultConfig.Cdc.Enabled {
+		l = &TxListener{tx: &Tx{}}
+	} else {
+		l = &kv.NoListener{}
+	}
+
+	return context.WithValue(ctx, kv.ListenerCtxKey{}, l)
```
himank, line 56: instead of using kv.ListernerCtxKey{}, can we use something like `dbname_listener` as key?

cdc/streamer.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/rs/zerolog/log"
+	"github.com/tigrisdata/tigrisdb/internal"
+	"github.com/tigrisdata/tigrisdb/server/config"
+)
+
+type Streamer struct {
+	db      fdb.Database
+	lastKey fdb.Key
+	cfg     config.CdcConfig
+	pub     *Publisher
```
himank, line 31: nit: do we need to have a publisher here? One option is to move out key-related methods in its own struct and then the streamer just can have access to that struct instead of the pub object.

cdc/listener.go
mattayres: fixed

cdc/manager.go
mattayres: caller doesn't know database name yet

cdc/streamer.go
mattayres: yeah, need to extend kvstore to support necessary functions

cdc/streamer.go
mattayres: doesn't `s.ticker.Stop()` used later handle this?

cdc/streamer.go
mattayres: k

cdc/streamer.go
himank: ah, yes - should we move it to defer? i.e. first line in the read method so that at any point if return ticker is closed.

server/config/options.go
mattayres: I was able to avoid a dependency issue by having `kv` no longer depend on `cdc`

cdc/streamer.go
himank: or do defer s.Close()

server/services/v1/query_lifecycle.go
mattayres: we need it here because the `tx.Commit(ctx)` call in the lifecycle needs to have the wrapped context

cdc/manager.go
himank: Should we have WrapContext inside the QueryRunner then you know the database name? And if it is not set i.e. if listener is not set by the caller then default to NoopListener.

cdc/listener.go
mattayres: a `TxListener` is created per context, when it's wrapped

server/config/options.go
himank: but with this server is importing the CDC package and the CDC package is importing the server package :) - maybe because it is in the config package that's why you're not seeing a dependency issue.

cdc/streamer.go
mattayres: that's really all the publisher is doing at the moment, handling key-related methods; it should probably be renamed

cdc/publisher.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/apple/foundationdb/bindings/go/src/fdb"
+	"github.com/apple/foundationdb/bindings/go/src/fdb/subspace"
+	"github.com/apple/foundationdb/bindings/go/src/fdb/tuple"
+	"github.com/tigrisdata/tigrisdb/server/config"
+	"github.com/tigrisdata/tigrisdb/store/kv"
+)
+
+type Publisher struct {
+	keySpace *PublisherKeySpace
+	listener *kv.Listener
```
himank, line 27: remove it.

cdc/streamer.go
```diff
@@ -0,0 +1,113 @@
...
+		return err
+	}
+
+	s.lastKey = key.(fdb.Key)
+	s.Txs = make(chan Tx, s.cfg.StreamBuffer)
+	s.ticker = time.NewTicker(s.cfg.StreamInterval)
+	go func() {
+		for range s.ticker.C {
+			if err := s.read(); err != nil {
+				log.Err(err).Msg("read failed")
```
himank, line 63: need to return from here after seeing the error.

himank (approved):

PR #229, opened 2022-04-27
fix: stage database object for explicit transactions

himank:


server/services/v1/query_lifecycle.go
```diff
@@ -16,7 +16,6 @@ package v1
 
 import (
 	"context"
-
```
efirs, line 19: goimports?

server/services/v1/query_runner.go
```diff
@@ -17,7 +17,6 @@ package v1
 import (
 	"bytes"
 	"context"
-
```
efirs, line 20: go imports?

efirs (approved):

PR #237, opened 2022-05-01
feat: Added getInfo handler to return server metadata

JigarJoshi:


server/services/v1/api.go
```diff
@@ -42,6 +43,8 @@ const (
 
 	documentPath        = collectionPath + "/documents"
 	documentPathPattern = documentPath + "/*"
+
+	infoPath = "/info"
```
himank, line 47: but why not call it version?

himank: what's the path look like?

JigarJoshi: `/api/v1/info` 

we had a long discussion on this one here https://github.com/tigrisdata/tigris-api/pull/46#discussion_r862020814

Makefile
```diff
@@ -20,7 +20,7 @@ all: server
 
 .PRECIOUS: ${PROTO_DIR}/%_openapi.yaml ${PROTO_DIR}/%.proto
 
-BUILD_PARAM=-tags=release -ldflags "-X 'main.Version=$(VERSION)' -X 'main.BuildHash=$(GIT_HASH)'" $(shell printenv BUILD_PARAM)
+BUILD_PARAM=-tags=release -ldflags "-X 'github.com/tigrisdata/tigris/util.Version=$(VERSION)' -X 'github.com/tigrisdata/tigris/util.BuildHash=$(GIT_HASH)'" $(shell printenv BUILD_PARAM)
```
efirs, line 23: I've removed GIT_HASH in client repo, I think it's redundant. There is enough information in VERSION itself.
It'll be symbolic name if tagged and symbolic+hash if not tagged.

server/services/v1/api.go
efirs: I think it's good idea to have more generic `info` where we can put any other server related metadata.

Makefile
JigarJoshi: removed `BuildHash`

server/services/v1/api.go
himank: Makes sense, let's go with info.

himank (approved):

PR #247, opened 2022-05-07
Test cleanup

efirs:
* Fix transaction rollback on failure
* Automatically retry implicit transactions
* Update client to newest and add high level client integration tests
* Removing logging from -v output
* Fix FDB connection when test on the hosts and server in the docker
* Various clean ups

### DIFF
--- a/store/kv/fdb.go
+++ b/store/kv/fdb.go
@@ -16,6 +16,7 @@ package kv
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 	"unsafe"
@@ -26,7 +27,6 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/tigrisdata/tigris/server/config"
 	ulog "github.com/tigrisdata/tigris/util/log"
-	"golang.org/x/xerrors"
 )
 
 const (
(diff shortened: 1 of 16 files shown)


himank (approved):

server/metadata/encoding/schema_test.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 func TestSchemaSubspace(t *testing.T) {
+	t.Skip("aaa")
```
himank, line 31: remove it

store/kv/fdb.go
```diff
@@ -481,15 +510,16 @@ func (t *ftx) Commit(ctx context.Context) error {
...
+
+			if err != nil && ep.Code == 1020 {
 				err = ErrConflictingTransaction
-			} else if err1 := t.tx.OnError(ep).Get(); err1 != nil {
-				err = err1
 			}
 		}
 
 		if err != nil {
+			t.tx.Cancel()
```
himank, line 522: do we need this cancel in fdbkv as well?

efirs: Yes, please check `txWithRetry` it has `defer tr.Cancel()`

himank (dismissed):

store/kv/fdb.go
himank: ah, right.

himank (approved):

PR #248, opened 2022-05-07
feat: added CDC event stream

mattayres:


server/metadata/tenant.go
```diff
@@ -112,10 +112,13 @@ type TenantManager struct {
 }
 
 func NewTenantManager() *TenantManager {
+	idToTenantMap := make(map[uint32]string)
+	idToTenantMap[DefaultNamespaceId] = DefaultNamespaceName
```
himank, line 116: is it causing some test failure? Otherwise, the reload method on the namespace manager should fill it during service start.

himank (dismissed):

server/services/v1/api.go
```diff
@@ -465,24 +465,27 @@ func (s *apiService) Stream(r *api.StreamRequest, stream api.Tigris_StreamServer
...
+					return api.Error(codes.Internal, "collection name decode failed")
+				}
 
-			if err := stream.Send(&api.StreamResponse{
-				Changes: changes,
-			}); ulog.E(err) {
-				return err
+				if r.Collection == "" || r.Collection == collection {
+					event := &api.StreamEvent{
+						Collection: collection,
```
himank, line 481: I left the comment on the API diff as well, collection will be part of the data? Because there could be different collections inside a transaction.

server/metadata/tenant.go
mattayres: It did break a test and I put a workaround in place to expect this default. I only added this because the collection name lookup only worked if the tenant was created in the same instantiation.

server/services/v1/api.go
mattayres: I don't believe so any more per my other comment.

himank (dismissed):

himank (dismissed):

himank (approved):

PR #253, opened 2022-05-12
Check if error is retriable at KV layer

efirs:


store/kv/fdb.go
```diff
@@ -127,10 +128,6 @@ func (d *fdbkv) txWithRetry(ctx context.Context, fn func(fdb.Transaction) (inter
 		if errors.As(err, &ep) {
 			// OnError returns nil if error is retryable
 			err = tr.OnError(ep).Get()
-
-			if err != nil && ep.Code == 1020 {
-				err = ErrConflictingTransaction
```
himank, line 132: we still need this error propagated back to API layer.

efirs: It'll retried till timeout or other error.

himank (dismissed):

himank (approved):

PR #259, opened 2022-05-18
feat: use timeout from request header, also use the context time for retries

himank:


server/services/v1/sessions.go
```diff
@@ -150,12 +152,24 @@ func (sessMgr *SessionManager) executeWithRetry(ctx context.Context, req *ReqOpt
...
+			if ok && time.Until(d) <= delta {
+				// if remaining is less than delta then probably not worth retrying
+				return
+			}
+			if !ok && time.Since(start) > (middleware.DefaultTimeout-delta) {
+				// this should not happen, adding a safeguard
+				return
+			}
+
+			log.Debug().Msgf("retrying transactions id: %s, since: %v", session.txCtx.Id, time.Since(start))
```
efirs, line 170: convert to:
```
log.Debug().Str("tx_id", session.txCtx.Id).Time("since", start).Msg("retrying transaction")
```

server/services/v1/sessions.go
```diff
@@ -150,12 +152,24 @@ func (sessMgr *SessionManager) executeWithRetry(ctx context.Context, req *ReqOpt
...
+			time.Sleep(time.Duration(rand.Intn(25)) * time.Millisecond)
```
efirs, line 171: Move 25ms to constants.

efirs (approved):

server/services/v1/sessions.go
himank: sure, will fix it in next diff.

PR #263, opened 2022-05-24
feat: auto-generate primary keys if not set by the user

himank:


server/services/v1/query_runner.go
```diff
@@ -150,49 +152,96 @@ func (runner *BaseQueryRunner) GetCollections(db *metadata.Database, collName st
...
+//  - error - if any error occurred
+func (runner *BaseQueryRunner) extractIndexParts(ctx context.Context, table []byte, userDefinedKeys []*schema.Field, doc []byte) ([]byte, []interface{}, []byte, error) {
 	var appendTo []interface{}
+	var keysReturnedInResp []byte
 	for _, v := range userDefinedKeys {
-		jsonVal, _, _, err := jsonparser.Get(doc, v.FieldName)
+		jsonVal, dtp, _, err := jsonparser.Get(doc, v.FieldName)
 		if err != nil {
-			return nil, api.Errorf(codes.InvalidArgument, errors.Wrapf(err, "missing index key column(s) '%s'", v.FieldName).Error())
+			if dtp == jsonparser.NotExist {
```
efirs, line 166:
```
if dtp != jsonparser.NotExist || !v.IsAutoGenerated() {
    return nil, nil, nil, api.Errorf(codes.InvalidArgument, errors.Wrapf(err, "parsing index key column(s) '%s'", v.FieldName).Error())
}
...
generate
```

test/v1/server/document_test.go
```diff
@@ -451,6 +460,183 @@ func (s *DocumentSuite) TestInsert_MultipleRows() {
...
+			"properties": Map{
+				"int_value": Map{
+					"type": "integer",
+				},
+				"pkey": Map{
+					"type":          "string",
+					"autoGenerated": true,
+				},
+			},
+			"primary_key": []interface{}{"pkey"},
```
efirs, line 477: From 1.18 it can be written as `[]any{"pkey"}`, we can start using it.

server/services/v1/key_generator.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return tsTime, tsNano, nil
+}
+
+// generateInt as it is used to generate int32 value, we are simply maintaining a counter. There is a contention to
+// generate a counter if it is concurrently getting executed but the generation should be fast then it is best to start
+// with this approach.
+func (g *generator) generateInt(ctx context.Context, tx transaction.Tx, table []byte) (int32, error) {
+	key := keys.NewKey([]byte(generatorSubspaceKey), table, int32IdKey)
+	it, err := tx.Read(ctx, key)
```
efirs, line 112: Looks like a point of contention.

server/services/v1/key_generator.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		var valueTS time.Time
+		var valueI64 int64
+		var valueI32 int32
+		switch field.Type() {
+		case schema.DateTimeType:
+			valueTS, _, err = g.generateTimestamp(ctx, tx, table)
+		case schema.Int64Type:
+			_, valueI64, err = g.generateTimestamp(ctx, tx, table)
```
efirs, line 69: Int32 is sequential and Int64 is random?

server/services/v1/key_generator.go
```diff
@@ -0,0 +1,131 @@
...
+
+// generateTimestamp stores values like this: "generator-<table>-timestamp_id-<tsNano>". The value is embedded in the
+// key to reduce the contention.
+// ToDo: cleanup old timestamps
+func (g *generator) generateTimestamp(ctx context.Context, tx transaction.Tx, table []byte) (time.Time, int64, error) {
+	tsTime := time.Now().UTC()
+	tsNano := time.Now().UTC().UnixNano()
+	key := keys.NewKey([]byte(generatorSubspaceKey), table, timestampIdKey, encoding.UInt32ToByte(uint32(tsNano)))
+
+	if err := tx.Insert(ctx, key, internal.NewTableData(encoding.UInt32ToByte(uint32(tsNano)))); err != nil {
```
efirs, line 100: Why do we need to save the timestamp?

lib/uuid/uuid.go
```diff
@@ -0,0 +1,11 @@
+package uuid
```
efirs, line 1: This file and package doesn't look very useful.

server/services/v1/key_generator.go
himank: that is correct, it is a good starting point for int32. Later we can think about something else.

server/services/v1/key_generator.go
himank: yeah to reduce the contention for int64, I am using timestamp nano.

lib/uuid/uuid.go
himank: this is just to streamline the use of uuid that we use the same google's package across the code.

test/v1/server/document_test.go
himank: sure

server/services/v1/key_generator.go
efirs: That's ok. There will be duplicate error when we actually insert the document.
There will be any kind of clock skews between workers, so this check doesn't look protect anything.

server/services/v1/key_generator.go
efirs: Can we just randomly genererate every type and regenerate if there is a duplicate key error when we insert the document?

efirs: Optimistic approach avoid any synchronization of metadata.

himank: how would that work in case user is using Replace API?

server/services/v1/key_generator.go
himank: What about Replace? How will we be handling this in that case?

himank: Clock-skews are fine in this case - all we need is a guarantee to not re-generate the same id again.

ovaistariq (dismissed):

schema/fields.go
```diff
@@ -231,3 +238,11 @@ func (f *Field) IsPrimaryKey() bool {
 
 	return *f.PrimaryKeyField
 }
+
+func (f *Field) IsAutoGenerated() bool {
+	if f.AutoGenerated == nil {
```
efirs, line 243: just `return f.AutoGenerated != nil && *f.AutoGenerated`

server/services/v1/query_runner.go
```diff
@@ -150,62 +153,129 @@ func (runner *BaseQueryRunner) GetCollections(db *metadata.Database, collName st
...
+			if doc, err = jsonparser.Set(doc, jsonVal, v.FieldName); err != nil {
+				return nil, nil, nil, false, err
+			}
+			autoGeneratedTypes = append(autoGeneratedTypes, v.DataType)
 		}
 
 		val, err := value.NewValue(v.DataType, jsonVal)
 		if err != nil {
-			return nil, err
+			return nil, nil, nil, false, err
```
efirs, line 190: Looks like to many return parameters, we need to think how to refactor this.

server/services/v1/query_runner.go
```diff
@@ -150,62 +153,129 @@ func (runner *BaseQueryRunner) GetCollections(db *metadata.Database, collName st
...
+	var autoGeneratedTypes []schema.FieldType
 	for _, v := range userDefinedKeys {
-		jsonVal, _, _, err := jsonparser.Get(doc, v.FieldName)
+		jsonVal, dtp, _, err := jsonparser.Get(doc, v.FieldName)
 		if err != nil {
-			return nil, api.Errorf(codes.InvalidArgument, errors.Wrapf(err, "missing index key column(s) '%s'", v.FieldName).Error())
+			if dtp != jsonparser.NotExist || !v.IsAutoGenerated() {
+				return nil, nil, nil, false, api.Errorf(codes.InvalidArgument, errors.Wrapf(err, "parsing index key column(s) '%s'", v.FieldName).Error())
+			}
+			if dtp == jsonparser.NotExist && !v.IsAutoGenerated() {
```
efirs, line 172: This need to be moved inside above `if` and be just `if !v.IsAutoGenerated`.

At this point `dtp == jsonparser.NotExist && v.IsAutoGenerated` is true, so the condition is always false.

server/services/v1/query_runner.go
```diff
@@ -150,62 +153,129 @@ func (runner *BaseQueryRunner) GetCollections(db *metadata.Database, collName st
...
+			keyValueResp = []byte(fmt.Sprintf(`"%s":"%s"`, v.FieldName, jsonVal))
+		} else {
+			// double quotes only needed on key, for auto-generated strings we are already adding it in generator
+			keyValueResp = []byte(fmt.Sprintf(`"%s":%s`, v.FieldName, jsonVal))
+		}
+		if len(keysReturnedInResp) > 0 {
+			keysReturnedInResp = append(keysReturnedInResp, []byte(`,`)...)
+			keysReturnedInResp = append(keysReturnedInResp, keyValueResp...)
+		} else {
+			keysReturnedInResp = keyValueResp
```
efirs, line 207: May be remove else and just unconditionally `keysReturnedInResp = append(keysReturnedInResp, keyValueResp...)`

server/services/v1/query_runner.go
```diff
@@ -150,62 +153,129 @@ func (runner *BaseQueryRunner) GetCollections(db *metadata.Database, collName st
...
 
 		var key keys.Key
-		if key, err = runner.encoder.EncodeKey(tenant.GetNamespace(), db, coll, coll.Indexes.PrimaryKey, indexParts); ulog.E(err) {
-			return nil, err
+		if key, err = runner.encoder.EncodeKey(encodedTable, coll.Indexes.PrimaryKey, indexParts); ulog.E(err) {
+			return nil, nil, err
 		}
 
 		tableData := internal.NewTableDataWithTS(ts, nil, doc)
+		if shouldUseInsert {
```
efirs, line 262: Why not use this variable directly in below if?

server/services/v1/query_runner.go
himank: that is correct - we can break down this method but it is mainly done for optimizations. For example, we are returning `keyValueResp ` from here so that we can simply reuse the memory assigned to the input document.

server/services/v1/query_runner.go
himank: We are only changing the method call if it is autoGenerate and field is int64/timestamp. Otherwise we need to honor what is passed in the API call. For example, API caller can pass insert true but `shouldUseInsert ` is false because the autoGenerate is not for this field. I can rename this variable to make it more clear.

efirs: got it.
Then just use it like `if insert || shouldUseInsert` {` in the next if.

himank: yeah, right!

efirs (approved):

PR #265, opened 2022-05-26
refac: use first byte as user table version during key encoding

himank:


server/metadata/key_encoder.go
```diff
@@ -15,13 +15,18 @@
...
+	"bytes"
 	api "github.com/tigrisdata/tigris/api/server/v1"
 	"github.com/tigrisdata/tigris/keys"
 	"github.com/tigrisdata/tigris/schema"
 	"github.com/tigrisdata/tigris/server/metadata/encoding"
 	"google.golang.org/grpc/codes"
 )
 
+var (
+	userTableKeyPrefix = []byte{'d', 'a', 't', 'a'}
```
efirs, line 27: maybe more readable like this: `[]byte("data")`

server/metadata/key_encoder.go
```diff
@@ -93,9 +99,16 @@ func (d *DictKeyEncoder) encodedIdxName(idx *schema.Index) []byte {
 }
 
 func (d *DictKeyEncoder) DecodeTableName(tableName []byte) (string, string, string, bool) {
-	nsId := encoding.ByteToUInt32(tableName[0:4])
-	dbId := encoding.ByteToUInt32(tableName[4:8])
-	collId := encoding.ByteToUInt32(tableName[8:12])
+	if len(tableName) < 16 {
```
efirs, line 102: `if len(tableName) < 16 || !bytes.HasPrefix(tableName, userTableKeyPrefix) {`

efirs (approved):

PR #273, opened 2022-06-02
chore: initial Docker file for fdb arm64 builds

rbarabas:
Adds a simplified FDB docker build process.

This enables us to build support images for platforms that are currently not supported by upstream and provide FDB docker images for M1 (arm64) processors.

Upstream uses CentOS 7 which does not yet support arm64 with the proper kernel page sizes so this build uses Ubuntu for the sake of simplicity.

This is just an initial commit. More layers will be added soon to optimize the build for deployment.


pboros:
This is great, but should this be in this repo or a separate one? At the end of the day, people probably won't build foundationdb, just use an image we provide.

rbarabas:
I moved this to another repository. Closing!

PR #275, opened 2022-06-03
refac: break listener into event listener and tx listener and manage txListener in server layer

himank:
This allows unifying the listeners and with this approach we just can have a single eventListener and that can be used by any pluggable TxListeners.


store/kv/listener.go
```diff
@@ -16,48 +16,87 @@ package kv
...
 
-func GetListener(ctx context.Context) Listener {
-	l, ok := ctx.Value(ListenerCtxKey{}).(Listener)
-	if l != nil && ok {
-		return l
-	} else {
-		return &NoListener{}
+func GetEventListener(ctx context.Context) EventListener {
+	a := ctx.Value(EventListenerCtxKey{})
+	fmt.Printf("GetEventListener %T%v\n", a, a)
```
mattayres, line 94: Probably not meant to be left here?

server/services/v1/tx_listener.go
```diff
@@ -0,0 +1,43 @@
...
+	// OnPostCommit is called after committing the transaction. This means the transaction is complete and any post
+	// activity can happen as part of this implementation.
+	OnPostCommit(context.Context, kv.EventListener) error
+	// OnRollback is called when the user transaction is rolled back.
+	OnRollback(context.Context, kv.EventListener)
+}
+
+type NoopTxListener struct{}
+
+func (l *NoopTxListener) OnCommit(context.Context, transaction.Tx, kv.EventListener) error {
```
mattayres, line 37: Maybe rename this to `OnPreCommit` now that there's an `OnPostCommit`.

store/kv/fdb.go
```diff
@@ -306,6 +313,10 @@ func (b *fbatch) SetVersionstampedValue(_ context.Context, _ []byte, _ []byte) e
 	return fmt.Errorf("batch doesn't support setting versionstamped value")
 }
 
+func (b *fbatch) SetVersionstampedKey(_ context.Context, _ []byte, _ []byte) error {
+	return fmt.Errorf("batch doesn't support setting versionstamped key")
```
mattayres, line 317: Should this return an error?

mattayres (approved):

store/kv/fdb.go
himank: realized after adding it in fdb layer that fdb is not returning an error for this. Makes sense, will remove it.

server/services/v1/tx_listener.go
himank: sg

store/kv/fdb.go
efirs: Maybe we can just remove batch in one of the next diffs, it's unused anyway.

efirs (approved):

PR #277, opened 2022-06-07
Pass transaction context in the headers

efirs:


api/server/v1/tx.go
```diff
@@ -14,17 +14,56 @@
...
+import (
+	"context"
+
+	"github.com/grpc-ecosystem/go-grpc-middleware/util/metautils"
+	"google.golang.org/grpc"
+	"google.golang.org/protobuf/proto"
+)
+
+var (
+	HeaderPrefix = "Tigris-"
```
himank, line 26: should we add `header.go` in api/server/v1 then? I am thinking then we can also move the timeout header here as well?

server/midddleware/pprof.go
```diff
@@ -21,7 +21,7 @@ import (
 	"google.golang.org/grpc"
 )
 
-func PprofUnaryServerInterceptor() grpc.UnaryServerInterceptor {
+func pprofUnaryServerInterceptor() grpc.UnaryServerInterceptor {
```
himank, line 24: any reason of changing this?

efirs: It's only used inside the the package, better to have only necessary scope.

efirs: It is. See above:
```
-		PprofUnaryServerInterceptor(),
+		pprofUnaryServerInterceptor(),
```
It's just to signify that the function is not used outside of `middleware` package.

himank: that is fine but main point is then just change the other as well that are inside this package. Mainly - `TimeoutUnaryServerInterceptor`, `TxCtxUnaryServerInterceptor`

himank (approved):

server/midddleware/pprof.go
efirs: Yes, this will be changed in the next diff.

PR #278, opened 2022-06-07
Transaction requests forwarder

efirs:


server/midddleware/forwarder.go
```diff
@@ -0,0 +1,88 @@
...
+
+var (
+	UserAgent = "tigris-forwarder/" + util.Version
+)
+
+var clients map[string]*grpc.ClientConn
+
+// TODO: Sweep unused connections periodically
+func getClient(ctx context.Context, origin string) (*grpc.ClientConn, error) {
+	if c, ok := clients[origin]; ok {
```
himank, line 23: Scope is global for the client map, you need to access it using lock as ForwardRequest will be called concurrently. Why not have a forwarder here? `type Forwarder struct {}` and have map inside it? Also, if we have explicit forwarder then it would be easy to have its own scope and metrics.

server/midddleware/forwarder.go
```diff
@@ -0,0 +1,88 @@
...
+
+	clients[origin] = conn
+
+	return conn, nil
+}
+
+func proxyDirector(ctx context.Context, _ string) (context.Context, *grpc.ClientConn, error) {
+	txCtx := api.GetTransaction(ctx, nil)
+
+	client, err := getClient(ctx, txCtx.GetOrigin())
```
himank, line 47: return `ctx, getClient(ctx, txCtx.GetOrigin())`

server/midddleware/forwarder.go
efirs: Good catch. Yes, will synchronize.

server/midddleware/forwarder.go
efirs: It doesn't look work this way.

himank (approved):

PR #280, opened 2022-06-07
feat: tigris search backend

himank:


query/filter/comparison.go
```diff
@@ -22,8 +22,11 @@ import (
 )
 
 const (
-	EQ = "$eq"
-	GT = "$gt"
+	EQ  = "$eq"
+	GT  = "$gt"
+	LT  = "$lt"
+	GTE = "$gte"
+	LTE = "$lte"
```
adilansari, line 29: need a `$ne` not equals matcher as well.

query/filter/comparison.go
```diff
@@ -111,3 +126,73 @@ func (g *GreaterThanMatcher) String() string {
...
+type GreaterThanEqMatcher struct {
+	Value value.Value
+}
+
+func (g *GreaterThanEqMatcher) GetValue() value.Value {
+	return g.Value
+}
+
+func (g *GreaterThanEqMatcher) Matches(input value.Value) bool {
+	fmt.Printf("input %v\n", input)
```
adilansari, line 140: printf on purpose?

test/docker/docker-compose.yml
```diff
@@ -25,6 +25,16 @@ services:
     ports:
       - "4500:4500"
 
+#docker run -p 8108:8108 -v/tmp/data:/data typesense/typesense:0.23.0 --data-dir /data --api-key=Hu52dwsas2AdxdE
+  tigris_search:
+    container_name: tigris_search
+    image: typesense/typesense:0.23.0
+    environment:
+      - TYPESENSE_DATA_DIR=/tmp
+      - TYPESENSE_API_KEY=Hu52dwsas2AdxdE
```
adilansari, line 34: should we find an alternate way to distribute this?

query/filter/comparison.go
```diff
@@ -22,8 +22,11 @@ import (
 )
 
 const (
-	EQ = "$eq"
-	GT = "$gt"
+	EQ  = "$eq"
```
JigarJoshi, line 25: should we add `$in` too as discussed earlier?

```
{
 '$in': {
  'state': ['CA', 'FL']
  }
}
```

himank: yes, let's do it iteratively. I need to test it on TS as well.

test/docker/docker-compose.yml
himank: This is for local dev, in prod we need a different key. We need to generate it and the pass it in the container.

query/filter/comparison.go
himank: good catch.

query/filter/comparison.go
himank: will do it iteratively, need to test on TS as well.

rbarabas:
Outside of the outstanding issues and what the others found, it's very nice!

query/filter/selector.go
```diff
@@ -46,6 +46,22 @@ func (s *Selector) Matches(doc []byte) bool {
 	return false
 }
 
+func (s *Selector) ToSearchFilter() string {
+	switch s.Matcher.Type() {
+	case EQ:
+		return fmt.Sprintf("%s:=%v", s.Field, s.Matcher.GetValue().AsInterface())
```
efirs, line 52: `op = ":="` in every case and
return `fmt.Sprintf("%s%s%v", s.Field, op, s.Matcher.GetValue().AsInterface())`
at the end.
also return "" in the default case

store/search/ts.go
```diff
@@ -0,0 +1,118 @@
...
+	client *typesense.Client
+}
+
+type IndexDocumentsOptions struct {
+	Action    string
+	BatchSize int
+}
+
+func (s *storeImpl) convertToInternalError(err error) error {
+	if e, ok := err.(*typesense.HTTPError); ok {
```
efirs, line 38:
```
var ep typesense.HTTPError
if errors.As(err, &ep) {
....
}
```

store/search/ts.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return err
+}
+
+func (s *storeImpl) DeleteDocuments(_ context.Context, table string, key string) error {
+	_, err := s.client.Collection(table).Document(key).Delete()
+	return s.convertToInternalError(err)
+}
+
+func (s *storeImpl) IndexDocuments(_ context.Context, table string, documents io.Reader, options IndexDocumentsOptions) (err error) {
```
efirs, line 56: I think the closure is redundant and we should handle the errors:
```
func (s *storeImpl) IndexDocuments(_ context.Context, table string, documents io.Reader, options IndexDocumentsOptions) error {
	reader, err := s.client.Collection(table).Documents().ImportJsonl(documents, &tsApi.ImportDocumentsParams{
		Action:    &options.Action,
		BatchSize: &options.BatchSize,
	})
    if err != nil {
        return err
    }
    defer func() { ulog.E(reader.Close()) }()

	type resp struct {
		Code     int
		Document string
		Error    string
		Success  bool
	}

	var r resp
	res, err := io.ReadAll(reader)
    if err != nil {
        return err
    }
	if err := jsoniter.Unmarshal(res, &r); err != nil {
        return err
    }
	if len(r.Error) > 0 {
		return fmt.Errorf(r.Error)
	}

	return nil
}
```

store/search/ts.go
```diff
@@ -0,0 +1,118 @@
...
+				err = fmt.Errorf(r.Error)
+			}
+			closer.Close()
+		}
+	}()
+	return err
+}
+
+func (s *storeImpl) Search(_ context.Context, table string, filterBy string) ([]tsApi.SearchResult, error) {
+	q := "*"
```
efirs, line 83: Is this always will be "*"? Maybe define a constant then.

store/search/ts.go
```diff
@@ -0,0 +1,118 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return res.Results, nil
+}
+
+func (s *storeImpl) CreateCollection(_ context.Context, schema *tsApi.CollectionSchema) error {
+	_, err := s.client.Collections().Create(schema)
```
efirs, line 103: _, err := s.client.Collections().Create(schema)
return s.convertToInternalError(err)

store/search/ts.go
```diff
@@ -0,0 +1,118 @@
...
+	if err == nil {
+		return nil
+	}
+
+	return s.convertToInternalError(err)
+}
+
+func (s *storeImpl) DropCollection(_ context.Context, table string) error {
+	_, err := s.client.Collection(table).Delete()
```
efirs, line 112: same as above

test/docker/docker-compose.yml
efirs: Should we make clear that it's not intended for production by just using something like 'ts_test_key1'?

store/search/search.go
```diff
@@ -0,0 +1,56 @@
...
+	IndexDocuments(ctx context.Context, table string, documents io.Reader, options IndexDocumentsOptions) error
+	DeleteDocuments(ctx context.Context, table string, key string) error
+	Search(ctx context.Context, table string, filterBy string) ([]tsApi.SearchResult, error)
+}
+
+func NewStore(config *config.SearchConfig) (Store, error) {
+	client := typesense.NewClient(
+		typesense.WithServer(fmt.Sprintf("http://%s:%d", config.GetHost(), config.Port)),
+		typesense.WithAPIKey(config.AuthKey))
+	return &storeImpl{
```
efirs, line 39: I think short initializers like this better to format as a single line, it'll more readable:
```
return &storeImpl{client: client}, nil
```

query/filter/logical.go
```diff
@@ -81,6 +81,17 @@ func (a *AndFilter) GetFilters() []Filter {
 	return a.filter
 }
 
+func (a *AndFilter) ToSearchFilter() string {
+	var str string
+	for i, f := range a.filter {
+		str += f.ToSearchFilter()
```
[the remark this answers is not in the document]
himank: OR doesn't work as-is. It needs more work because we need to create a new JSON request.

server/config/options.go
```diff
@@ -67,9 +68,31 @@ var DefaultConfig = Config{
 		StreamBatch:    100,
 		StreamBuffer:   200,
 	},
+	Search: SearchConfig{
+		Host:         "0.0.0.0",
+		Port:         8108,
+		ReadEnabled:  true,
+		WriteEnabled: true,
+		AuthKey:      "Hu52dwsas2AdxdE",
```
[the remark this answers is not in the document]
himank: This is just for local dev and it is public key provided by TS to play-around. Once we have a way to generate a key then it will be removed from here.

server/metadata/tenant.go
```diff
@@ -524,18 +536,24 @@ func (tenant *Tenant) CreateCollection(ctx context.Context, tx transaction.Tx, d
...
+		}
+	}
+
 	return nil
 }
 
 // DropCollection is to drop a collection and its associated indexes. It removes the "created" entry from the encoding
 // subspace and adds a "dropped" entry for the same collection key.
-func (tenant *Tenant) DropCollection(ctx context.Context, tx transaction.Tx, db *Database, collectionName string) error {
+func (tenant *Tenant) DropCollection(ctx context.Context, tx transaction.Tx, db *Database, collectionName string, searchStore search.Store) error {
```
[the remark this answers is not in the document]
himank: It is only passed for DropCollection/CreateCollection. I think it is fine rather than tenant depending on it.

store/search/ts.go
himank: this needs more thinking based on search API, placeholder for now.

store/search/ts.go
himank: yeah closure is not needed here.

store/search/errors.go
```diff
@@ -0,0 +1,58 @@
...
+func NewSearchError(httpCode int, code ErrCode, msg string, args ...interface{}) error {
+	return Error{httpCode: httpCode, code: code, msg: fmt.Sprintf(msg, args...)}
+}
+
+func (se Error) Error() string {
+	return se.msg
+}
+
+func IsSearchError(err error) bool {
+	switch err.(type) {
```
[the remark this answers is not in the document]
himank: never tried it, let me check.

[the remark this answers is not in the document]
himank: no, it checks for error, I need to check for the type.

server/services/v1/row_reader.go
```diff
@@ -0,0 +1,172 @@
...
+
+func MakeSearchRowReaderUsingFilter(ctx context.Context, table string, filters []filter.Filter, store search.Store) (*SearchRowReader, error) {
+	return MakeSearchRowReader(ctx, table, nil, filters, store)
+}
+
+func (s *SearchRowReader) NextRow(row *Row) bool {
+	for {
+		document, more := s.result.Hits.GetDocument(s.idx)
+		if !more {
+			break
```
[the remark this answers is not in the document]
himank: no, I need increment index.

server/services/v1/row_reader.go
```diff
@@ -0,0 +1,172 @@
...
+		data, err := json.Marshal(*document)
+		if err != nil {
+			s.err = err
+			return false
+		}
+		row.Key = []byte((*document)[searchID].(string))
+		row.Data = &internal.TableData{
+			RawData: data,
+		}
+		break
```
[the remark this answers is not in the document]
himank: ^

server/services/v1/row_reader.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (s *SearchRowReader) NextRow(row *Row) bool {
+	for {
+		document, more := s.result.Hits.GetDocument(s.idx)
+		if !more {
+			break
+		}
+
+		if document == nil {
```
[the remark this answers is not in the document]
himank: I am not sure yet, but hits has document and other metadata as well. I will remove it once I understand the response more.

server/services/v1/row_reader.go
```diff
@@ -0,0 +1,172 @@
...
+			s.err = err
+			return false
+		}
+		row.Key = []byte((*document)[searchID].(string))
+		row.Data = &internal.TableData{
+			RawData: data,
+		}
+		break
+	}
+	hasNext := s.result.Hits.HasMoreHits(s.idx)
```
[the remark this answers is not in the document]
himank: this is added because document may be nil. So once we are sure that document can't be nil then I'll remove this and above check on the document.

store/search/ts.go
himank: `As` is panicking, reverting it back to check explicitly for type.

test/docker/docker-compose.yml
himank: We need to set this key inside container which will be matched again the key that we are sending in the request. How will this work `ts_test_key1 `?

store/search/errors.go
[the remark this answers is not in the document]
efirs: Above does exactly that, but shorter than the switch.

server/services/v1/search_indexer.go
```diff
@@ -0,0 +1,157 @@
...
+		if collection == nil {
+			continue
+		}
+
+		searchKey, err := CreateSearchKey(event.Table, event.Key)
+		if err != nil {
+			return err
+		}
+
+		if event.Op == kv.DeleteEvent {
```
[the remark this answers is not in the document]
efirs: ok

server/services/v1/row_reader.go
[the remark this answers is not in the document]
efirs: `!more` mean there is no more documents, why `idx` need to be incremented?

store/search/ts.go
efirs: Interesting, ok.

store/search/ts.go
```diff
@@ -0,0 +1,122 @@
...
+
+	defer func() { ulog.E(closer.Close()) }()
+
+	type resp struct {
+		Code     int
+		Document string
+		Error    string
+		Success  bool
+	}
+	if closer != nil {
```
efirs, line 75: This check is redundant.

store/search/ts.go
```diff
@@ -0,0 +1,122 @@
...
+	return err
+}
+
+func (s *storeImpl) DeleteDocuments(_ context.Context, table string, key string) error {
+	_, err := s.client.Collection(table).Document(key).Delete()
+	return s.convertToInternalError(err)
+}
+
+func (s *storeImpl) IndexDocuments(_ context.Context, table string, documents io.Reader, options IndexDocumentsOptions) (err error) {
+	var closer io.ReadCloser
```
efirs, line 58: `reader` is a better name for it.

test/docker/docker-compose.yml
efirs: We set it to `ts_test_key1` in both places.
Here for the `typesense` container and in our config for the requests we are sending.

store/search/ts.go
himank: is it guaranteed to be non-nil?

test/docker/docker-compose.yml
himank: I tried it and it didn't work for me. There was some panic. I will debug it later.

server/services/v1/row_reader.go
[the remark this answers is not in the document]
himank: there is a check below if document is nil then we move forward.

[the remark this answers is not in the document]
efirs: so you are saying that (inline):
```
     for {   
        document, more := s.result.Hits.GetDocument(s.idx)
		if !more {  // this return that there is no more documents
			break
		}
     }
     hasNext := s.result.Hits.HasMoreHits(s.idx) // and this returns that there is more documents?
```
looks like something is not right here. The hasNext returns true and `row` object is not populated.

store/search/ts.go
efirs: if err != nil then yes. And it's already used in the `defer` :)

himank: but readCloser here is nothing but response.Body which can be nil if there is some unhandled edge cases in their client.

server/services/v1/row_reader.go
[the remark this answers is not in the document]
himank: why will hasNext returns true and the row object is not populated? The method guarantees that if it returned true then row object is populated that's why there is if check at line 84.

store/search/ts.go
efirs: ok. but that's unexpected.

server/services/v1/row_reader.go
```diff
@@ -0,0 +1,172 @@
...
+		idx:    0,
+		result: s,
+	}, nil
+}
+
+func MakeSearchRowReaderUsingFilter(ctx context.Context, table string, filters []filter.Filter, store search.Store) (*SearchRowReader, error) {
+	return MakeSearchRowReader(ctx, table, nil, filters, store)
+}
+
+func (s *SearchRowReader) NextRow(row *Row) bool {
```
efirs, line 77: This method is error prone, we need to refactor it in the future.

efirs (approved):

server/services/v1/row_reader.go
himank: why?

PR #281, opened 2022-06-08
Autogeneration of the fields fixed

efirs:
* Server was returning unquoted string if user pass non zero value of the
  autogenerated field
* Autogenerate if the value passed is "zero" value of its type


server/services/v1/query_runner.go
```diff
@@ -152,6 +154,22 @@ func (runner *BaseQueryRunner) GetCollections(db *metadata.Database, collName st
 	return collection, nil
 }
 
+func IsNull(tp schema.FieldType, val []byte) bool {
+	switch tp {
+	case schema.Int32Type:
+		return string(val) == "0"
```
himank, line 160: return `val == []byte(`0`)` won't work?

server/services/v1/query_runner.go
```diff
@@ -152,6 +154,22 @@ func (runner *BaseQueryRunner) GetCollections(db *metadata.Database, collName st
...
+	case schema.Int64Type:
+		return string(val) == "0"
+	case schema.UUIDType:
+		return string(val) == uuid.UUID{}.String()
+	case schema.DateTimeType:
+		return string(val) == time.Time{}.Format(time.RFC3339Nano)
```
himank, line 166: do we need format here? just time.Time{}?

himank (dismissed):

server/services/v1/query_runner.go
efirs: We need to, time is a raw string in the json document here.

himank (approved):

PR #283, opened 2022-06-09
chore: unify make targets for architectures

rbarabas:
Unifies make targets for the various architectures and upgrades FDB to 7.1.7 to support M1 natively.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -12,24 +12,33 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.18-bullseye AS build
+FROM ubuntu:jammy-20220531 AS build
 
-RUN apt-get update && apt-get install -y --no-install-recommends apt-transport-https sudo
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    gcc \
+    git \
+    golang \
+    make \
+    sudo
 
 RUN mkdir /build
 
 #Download deps once, during docker build. Rebuild only on go.mod change
+ENV PATH="${PATH}:/root/go/bin"
 COPY scripts/install_build_deps.sh /build
-RUN sh /build/install_build_deps.sh
 COPY go.mod /build
 WORKDIR /build
 RUN go mod download
-
 COPY . /build
-
+RUN sh /build/install_build_deps.sh
 RUN --mount=type=cache,target=/root/.cache/go-build rm -f server/service && make bins
 
-FROM debian:bullseye-slim
+
+FROM ubuntu:jammy-20220531 AS server
 
 # Remove apt configuration
 RUN rm -rf /etc/apt/*
--- a/scripts/install_build_deps.sh
+++ b/scripts/install_build_deps.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Copyright 2022 Tigris Data, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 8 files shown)


scripts/install_build_deps.sh
```diff
@@ -13,34 +14,83 @@
...
+case "${OS}" in
+	"Darwin")
+		FDB_PACKAGE_NAME="FoundationDB-${FDB_VERSION}_${ARCH}.pkg"
+		FDB_PACKAGE_PATH="$(mktemp -d)/${FDB_PACKAGE_NAME}"
+		curl --create-dirs -Lo "$FDB_PACKAGE_PATH" "https://tigrisdata-pub.s3.us-west-2.amazonaws.com/${FDB_PACKAGE_NAME}"
+		echo "$FDB_SHA  $FDB_PACKAGE_PATH" | shasum -a 512 -c
+		sudo installer -pkg "$FDB_PACKAGE_PATH" -target /
+		;;
+	"Linux")
+		sudo apt-get install -y protobuf-compiler
```
pboros, line 89: Duplicate protobuf compiler installation, also done above.

rbarabas: Great catch, fixed! Thanks!

docker/Dockerfile
```diff
@@ -12,9 +12,18 @@
...
+FROM ubuntu:jammy-20220531 AS build
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    gcc \
+    git \
+    golang \
```
efirs, line 24: What is the golang version bundled with `ubuntu:jammy-20220531`? We need to make sure it's at least 1.18.

docker/Dockerfile
```diff
@@ -12,9 +12,18 @@
...
-
-RUN apt-get update && apt-get install -y --no-install-recommends apt-transport-https sudo
+FROM ubuntu:jammy-20220531 AS build
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    gcc \
```
efirs, line 22: Do we need to install gcc?

docker/Dockerfile
```diff
@@ -24,12 +33,11 @@ RUN sh /build/install_build_deps.sh
 COPY go.mod /build
 WORKDIR /build
 RUN go mod download
-
 COPY . /build
+#RUN --mount=type=cache,target=/root/.cache/go-build rm -f server/service && make bins
+RUN rm -f server/service && make bins
```
efirs, line 38: mounting cache here, significantly speeds up the build

rbarabas: Thanks for the feedback, make sense, assumed so! I'm trying to figure out what's braking build in that it cannot find the protoc helper and tried to disable the BUILDKIT which helped me to get around the error locally. Unfortunately, as we can see it did not solve the build issue here so I'll be reverting this change soon.

docker/Dockerfile
rbarabas: Yeah, one of the builds failed without it. I don't have the exact location but if we remove it we should see where it fails in `make {run,test,local_test}`!

scripts/install_build_deps.sh
```diff
@@ -1,4 +1,5 @@
-#!/bin/bash
+#!/usr/bin/env bash
+set -ex
```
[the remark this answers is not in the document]
rbarabas: Sure, let's move it there!

docker/Dockerfile
rbarabas: 1.18.1

docker/Dockerfile
rbarabas: I added it back, didn't really help here.

himank:
Very cool!

himank (approved):

PR #287, opened 2022-06-10
refactor: GRPC metrics using tally and can be tagged properly

pboros:
GRPC metrics are using tally, this way they can be prefixed and tagged properly.


server/midddleware/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func newGrpcCallData(grpcMethod string, grpcType string) grpcCallData {
+	return grpcCallData{grpcMethod: grpcMethod, grpcType: grpcType}
+}
+
+func (g *grpcCallData) getTags() map[string]string {
+	fullMethodParts := strings.Split(g.grpcMethod, "/")
+	grpcService := fullMethodParts[1]
+	grpcMethod := fullMethodParts[2]
```
efirs, line 37: just use it directly below, no need in new variable

server/metrics/metrics.go
```diff
@@ -1,16 +1,27 @@
...
-func init() {
-	GRPCMetrics.EnableHandlingTimeHistogram()
-	PrometheusRegistry.MustRegister(GRPCMetrics)
+func InitializeMetrics() io.Closer {
+	var closer io.Closer
+	registry := prom.NewRegistry()
+	Reporter = promreporter.NewReporter(promreporter.Options{Registerer: registry})
+	Root, closer = tally.NewRootScope(tally.ScopeOptions{
+		Prefix:         "tigris",
+		Tags:           map[string]string{},
```
himank, line 22: do we need tags initialized in the root scope?

server/midddleware/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+func (g *grpcCallData) getTimeHistogram() tally.Histogram {
+	tags := g.getTags()
+	return metrics.Root.Tagged(tags).Histogram("grpc_server_handling_time_bucket", tally.DefaultBuckets)
+}
+
+func UnaryMetricsServerInterceptor() func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		callData := newGrpcCallData(info.FullMethod, "unary")
+		timeHistogram := callData.getTimeHistogram()
+		stopWatch := timeHistogram.Start()
```
efirs, line 83: just `stopWatch := callData.getTimeHistogram().Start()`

server/midddleware/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+		stopWatch.Stop()
+		return resp, err
+	}
+}
+func StreamMetricsServerInterceptor() grpc.StreamServerInterceptor {
+	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+		callData := newGrpcCallData(info.FullMethod, "stream")
+		timeHistogram := callData.getTimeHistogram()
+		stopWatch := timeHistogram.Start()
```
efirs, line 100: same as above

server/midddleware/metrics_test.go
```diff
@@ -0,0 +1,44 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package middleware
+
+import (
+	"github.com/tigrisdata/tigris/server/metrics"
```
himank, line 18: goimports?

server/midddleware/metrics.go
efirs: or even better you can do Go magic `defer callData.getTimeHistogram().Start().Stop()`
so you can get rid of both `timeHistorgram` and `stopWatch`.

server/midddleware/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+type grpcCallData struct {
+	grpcMethod string
+	grpcType   string
+}
+
+func newGrpcCallData(grpcMethod string, grpcType string) grpcCallData {
+	return grpcCallData{grpcMethod: grpcMethod, grpcType: grpcType}
+}
+
+func (g *grpcCallData) getTags() map[string]string {
```
himank, line 34: So as tags are expensive to create with each request should we think about defining some of the known grpc method tags already at the top? Mainly for `Insert/Replace/Update/Delete/Read`. If tag is already defined above if use it otherwise we create it in this method. This reduce tag creation for common APIs. @efirs @pboros ?

server/midddleware/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+	grpcMethod := fullMethodParts[2]
+	return map[string]string{
+		"grpc_method":  grpcMethod,
+		"grpc_service": grpcService,
+		"grpc_type":    g.grpcType,
+	}
+}
+
+func (g *grpcCallData) getGrpcCounter(name string) tally.Counter {
+	tags := g.getTags()
```
himank, line 44: if we do the above then all the metrics reporting would be cheap.

server/midddleware/metrics.go
```diff
@@ -0,0 +1,113 @@
...
+	return metrics.Root.Tagged(tags).Counter(name)
+}
+
+func (g *grpcCallData) increaseGrpcCounter(counter tally.Counter, value int64) {
+	counter.Inc(value)
+}
+
+func (g *grpcCallData) receiveMessage() {
+	counter := g.getGrpcCounter("grpc_server_msg_received_total")
```
himank, line 53: as the metric string is known, is it possible to just create this counter once? Same applies to `handleMessage` and `errorMessage`.

server/midddleware/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"context"
+	"github.com/tigrisdata/tigris/server/metrics"
+	"github.com/uber-go/tally"
+	"google.golang.org/grpc"
+	"strings"
+)
+
+type grpcCallData struct {
```
himank, line 25: if this is mainly used for metrics reporting should we just call it `grpcReqMetrics` or something similar.

server/midddleware/metrics.go
pboros: fixed

server/midddleware/metrics.go
pboros: fixed with the second approach

server/midddleware/metrics.go
pboros: Nice! I like this one.

server/midddleware/metrics_test.go
pboros: fixed

server/midddleware/metrics.go
pboros: We had a discussion about this, in the next diff, I will work towards making the counters and the tags both static (create them at the time of creating the services).

server/midddleware/metrics.go
pboros: We will pre-create what we can in the next iteration.

server/metrics/metrics.go
pboros: We will probably have them like version, etc.

server/midddleware/metrics.go
pboros: I renamed it.

server/midddleware/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package middleware
+
+import (
```
himank, line 17: goimports here as well :)

server/midddleware/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+	"google.golang.org/grpc"
+	"strings"
+)
+
+type grpcReqMetrics struct {
+	grpcMethod string
+	grpcType   string
+}
+
+func newGrpcCallData(grpcMethod string, grpcType string) grpcReqMetrics {
```
himank, line 30: nit: `newGrpcReqData`

server/midddleware/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (g *grpcReqMetrics) getTimeHistogram() tally.Histogram {
+	tags := g.getTags()
+	return metrics.Root.Tagged(tags).Histogram("grpc_server_handling_time_bucket", tally.DefaultBuckets)
+}
+
+func UnaryMetricsServerInterceptor() func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		callData := newGrpcCallData(info.FullMethod, "unary")
```
himank, line 79: callData --> reqMetrics

server/midddleware/metrics.go
pboros: fixed

server/midddleware/metrics.go
pboros: fixed all of these

himank (approved):

PR #293, opened 2022-06-13
feat: initialize GRPC metrics on startup

pboros:
Initialize all GRPC metrics with tags at server startup (on GRPC server creation).

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -1,3 +1,17 @@
+// Copyright 2022 Tigris Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package metrics
 
 import (
@@ -9,9 +23,26 @@ import (
 	promreporter "github.com/uber-go/tally/prometheus"
 )
 
+type ServerRequestCounter struct {
+	Name    string
+	Tags    map[string]string
+	Counter tally.Counter
+}
+
+type ServerRequestHistogram struct {
+	Name      string
+	Tags      map[string]string
+	Histogram tally.Histogram
+}
+
 var (
 	Root     tally.Scope
 	Reporter promreporter.Reporter
+	// Both counters and histograms are initialized during server startup in muxer.go
+	// method name and counter name
+	ServerRequestCounters map[string]map[string]*ServerRequestCounter
+	// method name and histogram name
+	ServerRequestHistograms map[string]map[string]*ServerRequestHistogram
 )
 
 func InitializeMetrics() io.Closer {
@@ -24,5 +55,7 @@ func InitializeMetrics() io.Closer {
 		CachedReporter: Reporter,
 		Separator:      promreporter.DefaultSeparator,
 	}, 1*time.Second)
+	ServerRequestCounters = make(map[string]map[string]*ServerRequestCounter)
+	ServerRequestHistograms = make(map[string]map[string]*ServerRequestHistogram)
 	return closer
 }
--- a/server/metrics/requests.go
+++ b/server/metrics/requests.go
@@ -1,0 +1,131 @@
+// Copyright 2022 Tigris Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"fmt"
+
+	"github.com/uber-go/tally"
+	"google.golang.org/grpc"
+)
+
+const (
+	ServerRequestsReceivedTotal      = "server_requests_received_total"
+	ServerRequestsHandledTotal       = "server_requests_handled_total"
+	ServerRequestsErrorTotal         = "server_requests_error_total"
+	ServerRequestsOkTotal            = "server_requests_ok_total"
+	ServerRequestsHandlingTimeBucket = "server_requests_handling_time_bucket"
+)
+
+var (
+	ServerRequestsCounterNames = []string{
+		ServerRequestsReceivedTotal,
+		ServerRequestsHandledTotal,
+		ServerRequestsErrorTotal,
+		ServerRequestsOkTotal,
+	}
+	ServerRequestsHistogramNames = []string{
+		ServerRequestsHandlingTimeBucket,
+	}
+)
+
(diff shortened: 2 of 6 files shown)


server/metrics/grpc.go
```diff
@@ -0,0 +1,131 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"fmt"
```
himank, line 18: goimports on all the files pls.

server/metrics/grpc.go
```diff
@@ -0,0 +1,131 @@
...
+	tags := endpointMetadata.getTags()
+
+	for _, histogramName := range getGrpcHistogramNames() {
+		histogram := GrpcHistogram{
+			Name:      histogramName,
+			Tags:      tags,
+			Histogram: Root.Tagged(tags).Histogram(histogramName, tally.DefaultBuckets),
+		}
+		_, ok := GrpcHistograms[fullMethodName]
+		if !ok {
```
himank, line 107: Can this list have duplicates? Do you need this check?

server/metrics/metrics.go
```diff
@@ -9,9 +23,25 @@ import (
...
+	Histogram tally.Histogram
+}
+
 var (
 	Root     tally.Scope
 	Reporter promreporter.Reporter
+	// method name and counter name
+	GrpcCounters map[string]map[string]*GrpcCounter
+	// method name and histogram name
+	GrpcHistograms map[string]map[string]*GrpcHistogram
```
himank, line 44: nit: also add a comment on these maps that these are only initialized once during startup to avoid the need for synchronization.

server/metrics/metrics.go
```diff
@@ -9,9 +23,25 @@ import (
...
+	Name      string
+	Tags      map[string]string
+	Histogram tally.Histogram
+}
+
 var (
 	Root     tally.Scope
 	Reporter promreporter.Reporter
+	// method name and counter name
+	GrpcCounters map[string]map[string]*GrpcCounter
```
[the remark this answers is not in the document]
himank: For FDB metrics, we can just have some metrics initialized in the fdbkv/ftx object present in the store package.

server/muxer/muxer.go
```diff
@@ -73,7 +74,10 @@ func (m *Muxer) Start(host string, port int16) error {
 	cm := cmux.New(l)
 	for _, s := range m.servers {
 		_ = s.Start(cm)
+		// Initialize the metrics for each GRPC service
+		if s.GetType() == "GRPC" {
```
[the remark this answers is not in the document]
rbarabas: https://github.com/tigrisdata/tigris/blob/56482770a3be26dbbc1a96b0a8fc854b63f0b5f8/server/types/types.go#L25-L26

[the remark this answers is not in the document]
rbarabas:
```
if s.GetType() == types.GRPCServer {
    ...
 }
 ```

File: server/metrics/grpc.go (hunk 1 above)
Comment (pboros) [reply server/metrics/grpc.go:18]: done

File: server/metrics/grpc.go (hunk 2 above)
Comment (pboros) [reply server/metrics/grpc.go:107]: It's not for duplicates, but for keys that do not exist yet in the nested map. The metrics are by method name and then by metric name, so access from the interceptor is cheap. With this removed, initialization of the nested map will panic because it will try to initialize 2 levels at once.

File: server/metrics/metrics.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (pboros) [reply server/metrics/metrics.go:42]: Sounds good.

File: server/muxer/muxer.go (hunk 5 above)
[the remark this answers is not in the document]
Comment (pboros) [reply server/muxer/muxer.go:78]: Good catch, thanks!

File: server/metrics/grpc.go [hunk 6]
```diff
@@ -0,0 +1,131 @@
...
+		"grpc_service": g.grpcServiceName,
+		"grpc_type":    g.grpcTypeName,
+	}
+}
+
+func (g *GrpcEndPointMetadata) getFullMethod() string {
+	return fmt.Sprintf("/%s/%s", g.grpcServiceName, g.grpcMethodName)
+}
+
+func getGrpcCounterNames() []string {
```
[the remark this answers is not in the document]
pboros: Did that, also made the values constants.

server/metrics/metrics.go
pboros: Good point, added

pboros:
Renamed GRPC metrics to request metrics, because HTTP is counted as well. Some example metrics:
```
tigris_server_requests_handling_time_bucket_bucket{grpc_method="ListDatabases",grpc_service="tigrisdata.v1.Tigris",grpc_type="unary",le="0"} 0
tigris_server_requests_handling_time_bucket_bucket{grpc_method="ListDatabases",grpc_service="tigrisdata.v1.Tigris",grpc_type="unary",le="0.01"} 2
tigris_server_requests_handling_time_bucket_bucket{grpc_method="ListDatabases",grpc_service="tigrisdata.v1.Tigris",grpc_type="unary",le="0.025"} 2

tigris_server_requests_receive_total{grpc_method="ListDatabases",grpc_service="tigrisdata.v1.Tigris",grpc_type="unary"} 2
```

server/metrics/requests.go
```diff
@@ -0,0 +1,133 @@
...
+	grpcTypeName    string
+}
+
+func NewGrpcEndpointMetadata(grpcServiceName string, grpcMethodName string, grpcTypeName string) GrpcEndPointMetadata {
+	return GrpcEndPointMetadata{grpcServiceName: grpcServiceName, grpcMethodName: grpcMethodName, grpcTypeName: grpcTypeName}
+}
+
+func (g *GrpcEndPointMetadata) getTags() map[string]string {
+	return map[string]string{
+		"grpc_method":  g.grpcMethodName,
```
efirs, line 56: I would call it just "method", "service", "type". It's relevant for both HTTP and GRPC

server/metrics/requests.go
```diff
@@ -0,0 +1,133 @@
...
+	tags := endpointMetadata.getTags()
+
+	for _, counterName := range ServerRequestsCounterNames {
+		counter := ServerRequestCounter{
+			Name:    counterName,
+			Tags:    tags,
+			Counter: Root.Tagged(tags).Counter(counterName),
+		}
+		_, ok := ServerRequestCounters[fullMethodName]
+		if !ok {
```
efirs, line 89: This two lines can be combined:
```
if _, ok := ServerRequestCounters[fullMethodName]; !ok {
...
```

server/metrics/requests.go
```diff
@@ -0,0 +1,133 @@
...
+	tags := endpointMetadata.getTags()
+
+	for _, histogramName := range ServerRequestsHistogramNames {
+		histogram := ServerRequestHistogram{
+			Name:      histogramName,
+			Tags:      tags,
+			Histogram: Root.Tagged(tags).Histogram(histogramName, tally.DefaultBuckets),
+		}
+		_, ok := ServerRequestHistograms[fullMethodName]
+		if !ok {
```
efirs, line 110: same a above

server/metrics/requests.go
pboros: Good point, thanks!

server/metrics/requests.go
pboros: Great catch!

pboros: Actually if I combine these 2 line it panics, because that part of the map is not initialized yet.

efirs (dismissed):

server/metrics/requests.go
pboros: Scratch that, it's combined now.

efirs (approved):

PR #294, opened 2022-06-13
feat: using binary64 for floating point filtering, add internal fields as reserved, fix metadata in response

himank:


internal/data.go
```diff
@@ -42,11 +42,23 @@ const (
...
 )
 
+// CreateNewTimestamp is a method used to construct timestamp from unixNano. In search backend we store internal
+// timestamps as unixNano.
+func CreateNewTimestamp(nano int64) *Timestamp {
+	ts := time.Unix(0, nano).UTC()
+	return &Timestamp{
+		Seconds:     ts.Unix(),
+		Nanoseconds: int64(ts.Nanosecond()),
+		UnixNano:    ts.UnixNano(),
```
efirs, line 52: This looks like a duplication of the above two fields.
We can just have a method `UnixNano() int64 { return t.Seconds * time.Nanosecond + t.Nanoseconds}`

value/value.go
```diff
@@ -122,11 +120,32 @@ func (i *IntValue) String() string {
...
+
+	i := &DoubleValue{
+		Double:   val,
+		asString: raw,
+		bin64Enc: math.Float64bits(val),
+	}
+	return i, nil
 }
 
 func (d *DoubleValue) CompareTo(v Value) (int, error) {
```
efirs, line 151: Do we save this DoubleValue in the search engine?
How ordering will work?

internal/data.go
himank: sure

value/value.go
himank: it is bin32 encoding so ordering will work but as it is bin32 encoding it is limited and does not work well for big floats.

efirs (approved):

PR #302, opened 2022-06-16
feat: honor id from the user model, use id as primary key if not set

himank:


schema/schema.go
```diff
@@ -156,14 +156,19 @@ func addPrimaryKeyIfMissing(reqSchema jsoniter.RawMessage) (jsoniter.RawMessage,
...
+		// primary key exists, no need to do anything.
 		return reqSchema, nil
 	}
 
 	schema[PrimaryKeySchemaK] = []string{AutoPrimaryKeyF}
 	if p, ok := schema["properties"]; ok {
-		p.(map[string]interface{})[AutoPrimaryKeyF] = map[string]string{
-			"type":   jsonSpecString,
-			"format": jsonSpecFormatUUID,
+		if _, ok = p.(map[string]interface{})[AutoPrimaryKeyF]; !ok {
```
efirs, line 165: I think unprotected type casting is very error prone. Can we use the library definition of the JSON schema or define simple structure like we do in the client:

```// Field represents JSON schema object
type Field struct {
	Type   string            `json:"type,omitempty"`
	Format string            `json:"format,omitempty"`
	Tags   []string          `json:"tags,omitempty"`
	Desc   string            `json:"description,omitempty"`
	Fields map[string]*Field `json:"properties,omitempty"`
	Items  *Field            `json:"items,omitempty"`

	AutoGenerate bool `json:"autoGenerate,omitempty"`
}

// Schema is top level JSON schema object
type Schema struct {
	Name       string            `json:"title,omitempty"`
	Desc       string            `json:"description,omitempty"`
	Fields     map[string]*Field `json:"properties,omitempty"`
	PrimaryKey []string          `json:"primary_key,omitempty"`
}```

efirs (dismissed):

schema/schema.go
himank: We can just add a casting check, we don't need to define the structure of the fields again. Check `fields.go` where we unmarshal the schema. But we don't want to do it at this point because there are intermediary steps to validate the schema. 

Also, doing it this way allows us to have the same logic, by just appending the `id` in the schema at the start when we receive the request. All other logic remains the same. Otherwise, we need to duplicate the work.

efirs: But we are unmarshalling entire schema and marshalling it again here.
Cand just the primary key to the structured schema in schema.go:Build ?

himank: We are Un-marshaling it but we are not changing the schema passed to us. Un-marshaling into the field means we need to keep the field exactly as the original schema which is not needed here.

I didn't understand second part of your comment.

ovaistariq (approved):

schema/schema.go
efirs: Can we just add the primary key to the structured schema in schema.go:Build ?

himank: That will not be sufficient. We need to add the field as well in properties.

efirs: Right, properties already structured in the Build too.

efirs (approved):

PR #309, opened 2022-06-20
feat: metrics for fdb accesses

pboros:
First version of counters and response time histograms for fdb accesses.


pboros:
I will add tests tomorrow, but it's ready for some initial feedback.

server/cdc/publisher.go
```diff
@@ -63,13 +64,17 @@ func NewPublisher(dbName string) *Publisher {
 }
 
 func (p *Publisher) NewStreamer(kvStore kv.KeyValueStore) (*Streamer, error) {
+	intDb, err := kvStore.GetInternalDatabase()
+	if err != nil {
+		ulog.E(err)
```
efirs, line 69: No return on error, is it intentional?

If not, it better be:

```
	if ulog.E(err) {
		return nil, err
    }
```

server/cdc/publisher.go
```diff
@@ -63,13 +64,17 @@ func NewPublisher(dbName string) *Publisher {
...
+	}
 	s := Streamer{
 		keySpace: p.keySpace,
-		db:       kvStore.GetInternalDatabase().(fdb.Database),
+		db:       intDb.(fdb.Database),
 		cfg:      config.DefaultConfig.Cdc,
 	}
 
-	err := s.start()
+	err = s.start()
```
efirs, line 77: combine with next line:
```
if err = s.start(); err != nil {
...
```

server/main.go
```diff
@@ -41,7 +41,8 @@ func main() {
 
 	log.Info().Str("version", util.Version).Msgf("Starting server")
 
-	kvStore, err := kv.NewKeyValueStore(&config.DefaultConfig.FoundationDB)
+	//kvStore, err := kv.NewKeyValueStore(&config.DefaultConfig.FoundationDB)
```
efirs, line 44: Remove the comment in the final version.

server/metrics/fdb.go
```diff
@@ -0,0 +1,78 @@
...
+
+package metrics
+
+import (
+	"github.com/uber-go/tally"
+	"strconv"
+)
+
+var (
+	FdbErrorRequests    tally.Scope
```
efirs, line 23: Unexport, if it won't be used outside of this package.

server/midddleware/metrics.go
```diff
@@ -62,10 +62,10 @@ func countSpecificErrorMessage(fullMethod string, methodType, errSource string,
 func UnaryMetricsServerInterceptor() func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 		methodType := "unary"
+		resp, err := handler(ctx, req)
 		tags := metrics.GetPreinitializedTagsFromFullMethod(info.FullMethod, methodType)
 		defer metrics.RequestsRespTime.Tagged(tags).Histogram("histogram", tally.DefaultBuckets).Start().Stop()
```
efirs, line 67: This should be called before `resp, err := handler(ctx, req)` otherwise we are missing call latency.

store/kv/kv.go
```diff
@@ -76,6 +85,103 @@ func NewKeyValueStore(cfg *config.FoundationDBConfig) (KeyValueStore, error) {
...
+			fdbkv: kv,
+		},
+	}, nil
+}
+
+func (m *KeyValueStoreImplWithMetrics) measure(name string, f func() error) {
+	tags := metrics.GetFdbReqTags(name, false)
+	defer metrics.FdbRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets).Start().Stop()
+	var fdberr fdb.Error
+	err := f()
```
efirs, line 104: Maybe reorder to reduce branching for readability
```
err := f()
if err == nil {
	metrics.FdbRequests.Tagged(tags).Counter("ok").Inc(1)
    return
}

var fdberr fdb.Error
if !errors.As(err, &fdberr) {
    metrics.FdbErrorRequests.Tagged(tags).Counter("unknown").Inc(1)
    return
}
		
errTags := metrics.GetFdbReqSpecificErrorTags(name, strconv.Itoa(fdberr.Code), false)
metrics.FdbErrorRequests.Tagged(errTags).Counter("specific").Inc(1)
	
```

store/kv/kv.go
```diff
@@ -167,14 +332,132 @@ func (k *KeyValueStoreImpl) BeginTx(ctx context.Context) (Tx, error) {
...
+	var err error
+	m.measure("BeginTx", func() error {
+		btx, err = m.kv.BeginTx(ctx)
+		return err
+	})
+
+	return &TxImplWithMetrics{
+		btx,
+	}, err
+
```
efirs, line 347: remove empty line

store/kv/kv.go
```diff
@@ -167,14 +332,132 @@ func (k *KeyValueStoreImpl) BeginTx(ctx context.Context) (Tx, error) {
...
 	*ftx
 }
 
+type TxImplWithMetrics struct {
+	tx Tx
+}
+
+func (m *TxImplWithMetrics) measure(name string, f func() error) {
+	tags := metrics.GetFdbReqTags(name, true)
+	defer metrics.FdbRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets).Start().Stop()
```
efirs, line 374: the difference with non-tx version is just tags.
so you can just create a helper method `measureLow(name, f, isTx)` and use it in both places.
```
func (m *TxImplWithMetrics) measure(name string, f func() error) {
	measureLow(name, f, true)
}
```
```
func (m *KeyValueStoreImplWithMetrics) measure(name string, f func() error) {
	measureLow(name, f, false)
}
```

store/kv/kv.go
```diff
@@ -114,6 +249,16 @@ func (k *KeyValueStoreImpl) ReadRange(ctx context.Context, table []byte, lkey Ke
 	}, nil
 }
 
+func (m *KeyValueStoreImplWithMetrics) ReadRange(ctx context.Context, table []byte, lkey Key, rkey Key) (Iterator, error) {
```
efirs, line 252: Should we use named parameters and naked return everywhere?
```
func (m *KeyValueStoreImplWithMetrics) ReadRange(ctx context.Context, table []byte, lkey Key, rkey Key) (it Iterator, err error) {
	m.measure("ReadRange", func() error {
		it, err = m.kv.ReadRange(ctx, table, lkey, rkey)
		return err
	})
	return
}
```

server/metrics/fdb.go
pboros: This is used in kv, measure() increases these counters.

server/main.go
pboros: done

server/cdc/publisher.go
pboros: Great catch, I was focusing on adding err, but it's always nil for `GetInternalDatabase()`. I changed this.

server/cdc/publisher.go
pboros: fixed

store/kv/kv.go
pboros: Good catch!

server/midddleware/metrics.go
pboros: Right, this is fixed.

store/kv/kv.go
pboros: This is great, also much easier for testing. changed it.

store/kv/kv.go
pboros: Changed like this everywhere.

efirs (approved):

PR #310, opened 2022-06-21
refactor: renamed stream to events and made it collection specific

mattayres:


himank (approved):

server/services/v1/api.go
```diff
@@ -398,7 +398,7 @@ func (s *apiService) GetInfo(_ context.Context, _ *api.GetInfoRequest) (*api.Get
 	}, nil
 }
 
-func (s *apiService) Stream(r *api.EventsRequest, stream api.Tigris_EventsServer) error {
+func (s *apiService) Events(r *api.EventsRequest, stream api.Tigris_EventsServer) error {
```
himank, line 401: Are you also thinking to restrict it to a collection?

himank (approved):

server/services/v1/api.go
mattayres: Isn't that handled implicitly since the endpoint only exists under `/collections` now?

PR #311, opened 2022-06-22
feat: allow collection updates for backward compatible changes

himank:


schema/collection.go
```diff
@@ -22,6 +22,8 @@ import (
 	"math"
 	"strconv"
 
+	"github.com/tigrisdata/tigris/lib/set"
```
efirs, line 25: goimports

server/config/options.go
```diff
@@ -73,6 +73,7 @@ var DefaultConfig = Config{
 		Port:         8108,
 		ReadEnabled:  true,
 		WriteEnabled: true,
+		AuthKey:      "ts_test_key",
```
himank, line 76: it is needed if the server is not running inside docker.

schema/fields.go
```diff
@@ -299,3 +346,18 @@ func (f *Field) IsPrimaryKey() bool {
 func (f *Field) IsAutoGenerated() bool {
 	return f.AutoGenerated != nil && *f.AutoGenerated
 }
+
+func (f *Field) IsCompatible(f1 *Field) error {
+	if f.DataType != f1.DataType {
+		return api.Errorf(api.Code_INVALID_ARGUMENT, "data type mismatch for field %q", f.FieldName)
+	}
+
+	if f.IsPrimaryKey() && !f1.IsPrimaryKey() {
```
efirs, line 355: `f.IsPrimaryKey() != f1.IsPrimaryKey()` ?

server/config/options.go
efirs: Instead of doing this you can just run `make local_run`

schema/fields.go
himank: ah, right - that will work.

server/config/options.go
himank: is that setting environment variables locally?

efirs: It uses `config/server.dev.yaml` to start the binary:
```
local_run: server
	$(DOCKER_COMPOSE) up --no-build --detach tigris_search tigris_fdb
	TIGRIS_ENVIRONMENT=dev ./server/service
```
After `make local_run` you can even attach to and debug the server from the IDE.

himank: but then I need to attach to remote process? I won't be able to simply run from the IDE, is that right?

efirs: You can run from IDE if you set `TIGRIS_ENVIRONMENT=dev` in the IDE.

efirs: When you attach, it's local process on you host.
You can set breakpoints and debug as usual, not much different from running from the IDE.

himank: So TIGRIS_ENVIRONMENT=dev still making sense to me but otherwise why introduce one more thing for the developer? All we need is to load the config object from the yaml based on the env.

himank: I have removed it but curl is not working for me when I am running the server using `make local_run`. Is it working for you?

efirs: It should. Let's check tomorrow, why it's not.

efirs (approved):

PR #313, opened 2022-06-22
feat: adding facets feature for search, honoring page from the request

himank:


api/server/v1/marshaler.go
```diff
@@ -451,11 +432,61 @@ func (x *SearchResponse) MarshalJSON() ([]byte, error) {
...
 	}{
-		Data: x.Data,
-		Meta: x.Meta,
+		Data:     x.Data,
+		Metadata: CreateMDFromSearchMD(x.Metadata),
 	}
 	return json.Marshal(resp)
 }
+
+type SearchHitMetadata struct {
```
efirs, line 444: Looks identical to the below, why do we need this one?

server/services/v1/query_runner.go
```diff
@@ -566,11 +583,23 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 
 		var row Row
 		for rowReader.Next(ctx, &row) {
+			var createdAt, updatedAt *timestamppb.Timestamp
+			if row.Data.CreatedAt != nil {
```
efirs, line 587: row.Data could have helpers GetCreatedAt(), GetUpdatedAt() so we don't need to check everytime and use it directly below.

server/services/v1/query_runner.go
```diff
@@ -546,14 +547,30 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
...
 		SearchFields(searchFields).
-		PageSize(defaultPerPage).
+		Facets(facets).
+		PageSize(pageSize).
 		Filter(wrappedF).
 		Build()
 
-	rowReader, err := NewSearchReader(ctx, runner.searchStore, collection, searchQ)
+	var rowReader *SearchRowReader
+	if runner.req.Page != 0 {
```
efirs, line 569: So first page index is 1?

server/services/v1/search_indexer.go
```diff
@@ -194,13 +194,13 @@ func PackSearchFields(data *internal.TableData, collection *schema.DefaultCollec
 	return jsoniter.Marshal(decData)
 }
 
-func UnpackSearchFields(doc *map[string]interface{}, collection *schema.DefaultCollection) (string, error) {
+func UnpackSearchFields(doc *map[string]interface{}, collection *schema.DefaultCollection) (string, *internal.TableData, error) {
```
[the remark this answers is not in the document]
himank: I hate it as well, all the TS responses are returning pointers to the corresponding responses, currently passing it as-is, but I am planning to change it.

server/services/v1/query_runner.go
himank: it is 1. I was thinking to make it 0, the only thing is clients need to make sure to use a pointer otherwise default value would be 0.

efirs (approved):

PR #314, opened 2022-06-22
Forwarder fixed and tests added

efirs:


Makefile
```diff
@@ -60,7 +60,7 @@ local_test: generate
 	go test $(TEST_PARAM) ./...
 
 run: clean generate
-	$(DOCKER_COMPOSE) up --build --detach tigris_server
+	$(DOCKER_COMPOSE) up --build --detach tigris_server2
```
himank, line 63: Do we need this change?

server/midddleware/forwarder.go
```diff
@@ -24,6 +30,35 @@ type Forwarder struct {
 
 var forwarder = Forwarder{}
 
+func requestToResponse(method string) (proto.Message, proto.Message) {
+	log.Debug().Str("method", method).Msg("requestToResponse")
+	switch strings.TrimPrefix(method, "/tigrisdata.v1.Tigris/") {
+	case "Insert":
```
himank, line 36: I think we have this defined somewhere?

server/midddleware/forwarder.go
```diff
@@ -86,9 +123,98 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
 		}
 
 		return handler(ctx, req)
 	}
 }
+
+func proxyHandler(_ interface{}, serverStream grpc.ServerStream) error {
+	fullMethodName, ok := grpc.MethodFromServerStream(serverStream)
+	if !ok {
+		return status.Errorf(codes.Internal, "lowLevelServerStream not exists in context")
```
himank, line 138: Don't we need to return tigris error from here?

server/midddleware/forwarder.go
```diff
@@ -86,9 +123,98 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+			}
+		case c2sErr := <-c2sErrChan:
+			serverStream.SetTrailer(clientStream.Trailer())
+			if c2sErr != io.EOF {
+				return c2sErr
+			}
+			return nil
+		}
+	}
+	return status.Errorf(codes.Internal, "gRPC proxying should never reach this stage.")
```
himank, line 174: ^

server/midddleware/forwarder.go
```diff
@@ -86,9 +123,98 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+			}
+		}
+	}()
+	return ret
+}
+
+func forwardServerToClient(src grpc.ServerStream, req proto.Message, dst grpc.ClientStream) chan error {
+	ret := make(chan error, 1)
+	go func() {
+		for i := 0; ; i++ {
```
himank, line 208: Why not just for {}

server/midddleware/forwarder.go
```diff
@@ -86,9 +123,98 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+	s2cErrChan := forwardServerToClient(serverStream, req, clientStream)
+	c2sErrChan := forwardClientToServer(clientStream, resp, serverStream)
+
+	for i := 0; i < 2; i++ {
+		select {
+		case s2cErr := <-s2cErrChan:
+			if s2cErr == io.EOF {
+				_ = clientStream.CloseSend()
+			} else {
+				clientCancel()
```
himank, line 163: Don't we need to always cancel?

server/midddleware/forwarder.go
```diff
@@ -86,9 +123,98 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+				return status.Errorf(codes.Internal, "failed proxying s2c: %v", s2cErr)
+			}
+		case c2sErr := <-c2sErrChan:
```
himank, line 166: So we don't need to clientCancel() in this case?

server/midddleware/forwarder.go
efirs: I've tried to look it up. Let me try once again.

server/midddleware/forwarder.go
efirs: Let me change. But it actually doesn't matter. So as standard codes are mapped one to one to Tigris codes.

server/midddleware/forwarder.go
efirs: I is used to send headers on first message.

server/midddleware/forwarder.go
himank: Should we define it? To avoid it getting out of sync.

server/midddleware/forwarder.go
himank: actually, you already have a defer above, do you need this?

server/midddleware/forwarder.go
himank: where is that?

efirs: ah, its in other function. removing this one.

server/midddleware/forwarder.go
efirs: good catch. removing.

server/midddleware/forwarder.go
```diff
@@ -86,9 +120,97 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+
+	clientStream, err := grpc.NewClientStream(clientCtx, &grpc.StreamDesc{ServerStreams: true, ClientStreams: true}, backendConn, fullMethodName)
+	if err != nil {
+		return err
+	}
+
+	s2cErrChan := forwardServerToClient(serverStream, req, clientStream)
+	c2sErrChan := forwardClientToServer(clientStream, resp, serverStream)
+
+	for i := 0; i < 2; i++ {
```
himank, line 154: is this needed because server chan can return `io.EoF` and we need to wait for `c2sErrChan` to return something?

server/midddleware/forwarder.go
```diff
@@ -86,9 +120,97 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+			}
+		}
+	}()
+	return ret
+}
+
+func forwardServerToClient(src grpc.ServerStream, req proto.Message, dst grpc.ClientStream) chan error {
+	ret := make(chan error, 1)
+	go func() {
+		for {
```
himank, line 204: is this for loop because to handle reads? Where there will be multiple back-forth messages streamed?

server/midddleware/forwarder.go
```diff
@@ -24,6 +28,34 @@ type Forwarder struct {
...
+	case "Replace":
+		return &api.ReplaceRequest{}, &api.ReplaceResponse{}
+	case "Update":
+		return &api.UpdateRequest{}, &api.UpdateResponse{}
+	case "Delete":
+		return &api.DeleteRequest{}, &api.DeleteResponse{}
+	case "Read":
+		return &api.ReadRequest{}, &api.ReadResponse{}
+	case "Stream":
+		return &api.StreamRequest{}, &api.StreamResponse{}
```
himank, line 44: I have fixed it, please rebase.

server/midddleware/forwarder.go
efirs: This is to handle BiDi streams. This is to send messages downstream.
ClientToServer is to send read messages back upstream.

server/midddleware/forwarder.go
efirs: Yes, this read from both channels. It doesn't actually matter which one we are closing first. Refactored, please check again.

server/midddleware/forwarder.go
```diff
@@ -86,9 +120,92 @@ func forwarderUnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+
+	clientStream, err := grpc.NewClientStream(clientCtx, &grpc.StreamDesc{ServerStreams: true, ClientStreams: true}, backendConn, fullMethodName)
+	if err != nil {
+		return err
+	}
+
+	downErr := <-forwardDownstream(serverStream, req, clientStream)
+	_ = clientStream.CloseSend()
+
+	upErr := <-forwardUpstream(clientStream, resp, serverStream)
```
himank, line 154: Line 151(downErr := <- ..) is blocking? Which means if there is an upErr we won't see it because thread will be blocked at downErr till there is an error from downStream?

efirs: so you think better to have select here?

himank: Yeah, then with either error we can return.

efirs: Actually this is totally wrong. Let me think how to manage this.

Makefile
rbarabas, line 63: Should we rename these instances instead of using a generic numeric index? Is it possible to name them in a way that reflects their purpose?

server/midddleware/forwarder.go
rbarabas: Are we talking about the prefix stored as a constant somewhere? If so, I agree.

query/filter/filter.go
```diff
@@ -81,7 +81,7 @@ func NewWrappedFilter(filters []Filter) *WrappedFilter {
 }
 
 func All(reqFilter []byte) bool {
-	return bytes.Equal(reqFilter, filterAll)
+	return len(reqFilter) == 0 || bytes.Equal(reqFilter, filterAll)
```
rbarabas, line 84: Nice!

server/services/v1/query_runner.go
```diff
@@ -218,20 +218,17 @@ func (runner *BaseQueryRunner) buildKeysUsingFilter(tenant *metadata.Tenant, db
...
-		}
 		return runner.encoder.EncodeKey(encodedTable, primaryKeyIndex, indexParts)
 	}))
-	iKeys, err := kb.Build(filters, coll.Indexes.PrimaryKey.Fields)
-	if err != nil {
-		return nil, err
-	}
 
-	return iKeys, nil
+	return kb.Build(filters, coll.Indexes.PrimaryKey.Fields)
```
rbarabas, line 231: Nice!

server/services/v1/query_runner.go
```diff
@@ -218,20 +218,17 @@ func (runner *BaseQueryRunner) buildKeysUsingFilter(tenant *metadata.Tenant, db
 		return nil, err
 	}
 
+	encodedTable, err := runner.encoder.EncodeTableName(tenant.GetNamespace(), db, coll)
```
rbarabas, line 221: Nice!

test/config/config.go
```diff
@@ -26,3 +26,12 @@ func GetBaseURL() string {
 	}
 	return "http://localhost:8081"
 }
+
+func GetBaseURL2() string {
```
rbarabas, line 30: I'm not sure about this function. What are your thoughts about putting the whole environment setup logic in a single `GetBaseURL()` function instead?

test/v1/server/server_test.go
```diff
@@ -29,3 +51,85 @@ func info(t *testing.T) *httpexpect.Response {
 	return e.GET(config.GetBaseURL() + "/info").
 		Expect()
 }
+
+func TestTxForwarder(t *testing.T) {
+	e1 := expectLow(t, config.GetBaseURL())
```
rbarabas, line 56: Could we just use a single `config.GetBaseURL()` call and dedup the code?

test/v1/server/document_test.go
```diff
@@ -45,13 +45,17 @@ type DocumentSuite struct {
 	database   string
 }
 
-func expect(s httpexpect.LoggerReporter) *httpexpect.Expect {
+func expectLow(s httpexpect.LoggerReporter, url string) *httpexpect.Expect {
```
rbarabas, line 48: Wonder if a generic `GetBaseURL()` pattern might eliminate the need of having to implement this function.

Makefile
efirs: This just second instance of tigris_server (tigris_server too :)
It's not forwarder specific it can be used for any kind of distributed testing.

server/midddleware/forwarder.go
efirs: Yeah, prefix and method names. We'll define constants in later diff, currently it used here only.

test/v1/server/document_test.go
efirs: Sound like a good idea, I'll fix it in one of the next iterations.

test/v1/server/server_test.go
efirs: Yeah, will optimize it when will work on the next test.

PR #315, opened 2022-06-23
feat: Added basic restriction for collection field names

JigarJoshi:
- basic restriction for collection field names (language keywords not allowed and fixed pattern)
- field name can start with `[a-zA-Z_$]` and then contain zero or more `[a-zA-Z0-9_$]` 

The motivation here is to future proof various programming language's models. 

TODO

 - we can allow user to disable programming language keyword restriction on field names optionally (not by default) in case if they know they are never going to work with some programming language, there is no point of restricting them for that scenario.
 - Add similar restriction for collection and database name.
 - Fine tune these restriction to make more tighter as we learn iteratively.

### DIFF
--- a/schema/fields_test.go
+++ b/schema/fields_test.go
@@ -15,6 +15,7 @@
 package schema
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -61,7 +62,7 @@ func TestFieldBuilder_Build(t *testing.T) {
 			},
 		}
 		for _, c := range cases {
-			_, err := c.builder.Build()
+			_, err := c.builder.Build(false)
 			require.Equal(t, c.expError, err)
 		}
 	})
@@ -98,7 +99,36 @@ func TestFieldBuilder_Build(t *testing.T) {
 	})
 	t.Run("test reserved fields", func(t *testing.T) {
 		builder := &FieldBuilder{FieldName: "created_at", Type: "date-time"}
-		_, err := builder.Build()
+		_, err := builder.Build(false)
 		require.Error(t, err)
+	})
+
+	t.Run("test programming language keywords as field name", func(t *testing.T) {
+		keywords := []string{"abstract", "integer", "yield"}
+		for _, keyword := range keywords {
+			_, err := (&FieldBuilder{FieldName: keyword, Type: "string"}).Build(false) // one time builder, thrown away after test concluded
+			require.Equal(t, err, api.Errorf(api.Code_INVALID_ARGUMENT,
+				fmt.Sprintf("Invalid collection field name, It contains language keyword for fieldName = '%s'", keyword)))
+		}
+	})
+
+	t.Run("test invalid field name pattern", func(t *testing.T) {
+		invalidFieldNames := []string{"0id", "0ID", "("}
+		for _, invalidFieldName := range invalidFieldNames {
+			_, err := (&FieldBuilder{FieldName: invalidFieldName, Type: "string"}).Build(false) // one time builder, thrown away after test concluded
+			require.Equal(t, err,
+				api.Errorf(api.Code_INVALID_ARGUMENT,
+					"Invalid collection field name, field name can only contain [a-zA-Z0-9_$] and it can only start"+
+						" with [a-zA-Z_$] for fieldName = '%s'",
+					invalidFieldName))
+		}
+	})
+
+	t.Run("test valid field name pattern", func(t *testing.T) {
+		validFieldNames := []string{"a1", "$a1", "$_a1", "$_", "A1", "Z1"}
+		for _, validFieldName := range validFieldNames {
+			_, err := (&FieldBuilder{FieldName: validFieldName, Type: "string"}).Build(false) // one time builder, thrown away after test concluded
+			require.NoError(t, err)
+		}
 	})
 }
--- a/schema/fields.go
+++ b/schema/fields.go
@@ -15,6 +15,7 @@
 package schema
 
 import (
+	"regexp"
 	"strings"
 
 	jsoniter "github.com/json-iterator/go"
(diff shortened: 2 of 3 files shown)


adilansari (approved):

schema/rules.go
```diff
@@ -1,19 +1,27 @@
...
...l": true, "not": true, "notnull": true, "nuint": true, "null": true, "number": true, "object": true, "of": true, "on": true, "operator": true, "or": true, "orderby": true, "out": true, "override": true, "package": true, "params": true, "partial": true, "pass": true, "print": true, "private": true, "protected": true, "public": true, "raise": true, "range": true, "readonly": true, "record": true, "ref": true, "remove": true, "require": true, "require_once": true, "return": true, "sbyte": true, "sealed": true, "select": true, "set": true, "short": true, "sizeof": true, "stackalloc": true, "static": true, "strictfp": true, "string": true, "struct": true, "super": true, "switch": true, "symbol": true, "synchronized": true, "this": true, "throw": true, "throws": true, "trait": true, "transient": true, "true": true, "try": true, "type": true, "typealias": true, "typeof": true, "uint": true, "ulong": true, "unchecked": true, "unmanaged": true, "unsafe": true, "unset": true, "use": true, "ushort": true, "using": true, "val": true, "value": true, "var": true, "virtual": true, "void": true, "volatile": true, "when": true, "where": true, "while": true, "with": true, "xor": true, "yield": true}
```
efirs, line 16: Looks like very long line. Can we break it for readability?
Not word per line, but by some rule:
  * 80-100 characters per line
  * Letter per line
  * 5-10 words per line
  * Line per language

efirs: Also, consider using set.New() our lib object from `"github.com/tigrisdata/tigris/lib/set"`

schema/rules.go
```diff
@@ -1,19 +1,27 @@
...
+	ValidFieldNamePattern         = regexp.MustCompile(`^[a-zA-Z_$][a-zA-Z0-9_$]*$`)
 )
 
 var (
 	validators = []Validator{
 		&IndexSchemaValidator{},
 		&FieldSchemaValidator{},
+		&FieldNameValidator{},
```
himank, line 24: Actually, these validators are added for backward compatibility checks which means this code path won't execute when we receive a new schema. Add this check in the `Validate` method on the `fieldBuilder`. This `Validate` method is where we check if the field is valid and if we should accept the schema.

schema/rules.go
JigarJoshi: Thanks for pointing to set API.

schema/rules.go
JigarJoshi: Thanks for catching this Himank. After looking at the field.go - the build() method is more appropriate to place field name validation. Wdyt?

himank: Either place works, build or validate.

himank (approved):

schema/fields_test.go
```diff
@@ -101,4 +101,59 @@ func TestFieldBuilder_Build(t *testing.T) {
...
+
+		builder3 := &FieldBuilder{FieldName: "yield", Type: "string"}
+		_, err3 := builder3.Build()
+		require.Equal(t, err3, api.Errorf(api.Code_INVALID_ARGUMENT, "Invalid collection field name, It contains language keyword for fieldName = 'yield'"))
+	})
+
+	t.Run("test invalid field name pattern", func(t *testing.T) {
+		builder1 := &FieldBuilder{FieldName: "0id", Type: "string"}
+		_, err1 := builder1.Build()
+		require.Equal(t, err1, api.Errorf(api.Code_INVALID_ARGUMENT, "Invalid collection field name, field name can only contain [a-zA-Z0-9_$] and it can only start with [a-zA-Z_$] for fieldName = '0id'"))
```
rbarabas, line 122: What are your thoughts on incorporate `MsgFieldNameInvalidPattern` in the second argument of `api.Errorf()`?

schema/fields_test.go
```diff
@@ -101,4 +101,59 @@ func TestFieldBuilder_Build(t *testing.T) {
...
+		_, err2 := builder2.Build()
+		require.Equal(t, err2, api.Errorf(api.Code_INVALID_ARGUMENT, "Invalid collection field name, field name can only contain [a-zA-Z0-9_$] and it can only start with [a-zA-Z_$] for fieldName = '0ID'"))
+
+		builder3 := &FieldBuilder{FieldName: "(", Type: "string"}
+		_, err3 := builder3.Build()
+		require.Equal(t, err3, api.Errorf(api.Code_INVALID_ARGUMENT, "Invalid collection field name, field name can only contain [a-zA-Z0-9_$] and it can only start with [a-zA-Z_$] for fieldName = '('"))
+	})
+
+	t.Run("test valid field name pattern", func(t *testing.T) {
+		builder1 := &FieldBuilder{FieldName: "a1", Type: "string"}
```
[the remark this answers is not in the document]
JigarJoshi: sure updated it.

schema/fields_test.go
```diff
@@ -101,4 +101,59 @@ func TestFieldBuilder_Build(t *testing.T) {
 		_, err := builder.Build()
 		require.Error(t, err)
 	})
+
+	t.Run("test programming language keywords as field name", func(t *testing.T) {
+		builder1 := &FieldBuilder{FieldName: "abstract", Type: "string"}
+		_, err1 := builder1.Build()
+		require.Equal(t, err1, api.Errorf(api.Code_INVALID_ARGUMENT, "Invalid collection field name, It contains language keyword for fieldName = 'abstract'"))
```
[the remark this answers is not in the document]
JigarJoshi: sure updated it. I personally like to duplicate messages in test so that if we change message in future with the intention to change it at only one place and if it is being used at more than one place, our test will give us a light punch.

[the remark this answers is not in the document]
JigarJoshi: Thanks @rbarabas this is much concise.

schema/rules.go
JigarJoshi: updated. thanks!

schema/fields_test.go
JigarJoshi: updated.

himank (approved):

PR #316, opened 2022-06-24
Local development image build

efirs:
* Add docker file for building monocontainer development image.
* Used docker compose plugin for local builds
* Makefile improvements:
    * Only rebuild and restart changed containers
    * Cache tests results (only run changed tests)
  (all caches, volumes still can be clean and all containers restart by `make clean`)


Makefile
```diff
@@ -10,7 +10,7 @@ DATA_PROTO_DIR=internal
 
 # Needed to be able to build amd64 binaries on MacOS M1
 DOCKER_DIR=test/docker
-DOCKER_COMPOSE=COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose -f ${DOCKER_DIR}/docker-compose.yml
+DOCKER_COMPOSE=COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker compose -f ${DOCKER_DIR}/docker-compose.yml
```
rbarabas, line 13: Should we add a version check for docker to ensure clients clearly see what's going on?

rbarabas: Also, do we need to amed the `.github/workflows/go-test.yaml` GitHub workflow as well?

rbarabas (approved):
LGTM. Left minor remarks, probably the go-test workflow the more significant but neither of them a blocker.

Makefile
efirs: I think Compose V2 should be included in most recent installations.

I'll fix go-test.yaml (Good catch, thanks!)

.github/workflows/push-docker-local-image.yaml
```diff
@@ -0,0 +1,76 @@
...
+    types: [published]
+  workflow_call:
+    secrets:
+      GH_DOCKER_ACCESS_USER:
+        required: true
+      GH_DOCKER_ACCESS_TOKEN:
+        required: true
+
+env:
+  DOCKER_REPOSITORY: "tigrisdata/tigris"
```
ovaistariq, line 19: We probably want to push to a different repo tigrisdata/tigris-allinone or tigris-local or something like that.

docker/Dockerfile.local
```diff
@@ -0,0 +1,90 @@
...
+
+
+FROM ubuntu:jammy-20220531 AS server
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+	curl \
+	tree
+
+ADD https://dl.typesense.org/releases/0.23.0/typesense-server-0.23.0-amd64.deb /tmp
```
ovaistariq, line 48: We would need to install platform specific packages here and for FDB below?

test/docker/docker-compose.yml
```diff
@@ -46,14 +46,14 @@ services:
...
-        source: fdbdata
+        source: dbdata
         target: /etc/foundationdb/
     ports:
       - "8081:8081"
     command: >
       bash -c 'fdbcli --exec "configure new single memory" && /server/service'
     depends_on:
-      - tigris_fdb
+      - tigris_db
```
ovaistariq, line 56: Why rename it?

docker/Dockerfile.local
efirs: Right.

test/docker/docker-compose.yml
efirs: To be consistent with `tigris_search`

.github/workflows/push-docker-local-image.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build and push Docker images
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: docker/Dockerfile.local
+          platforms: linux/amd64
```
rbarabas, line 73: Do we intend to support arm64 here as well?

Makefile
```diff
@@ -48,28 +48,28 @@ docker_compose_build:
...
 docker_test: generate
-	$(DOCKER_COMPOSE) up --build --abort-on-container-exit --exit-code-from tigris_test tigris_test
+	$(DOCKER_COMPOSE) up --build tigris_test tigris_test
 
 docker_test_no_build:
-	$(DOCKER_COMPOSE) up --no-build --abort-on-container-exit --exit-code-from tigris_test tigris_test
+	$(DOCKER_COMPOSE) up --no-build tigris_test tigris_test
 
-test: clean docker_test
+test: docker_test
```
rbarabas, line 56: I think it makes sense to avoid a clean each time we run these. Having said that, should we mention this behavior change in the README?

Makefile
```diff
@@ -48,28 +48,28 @@ docker_compose_build:
...
 
-run: clean generate
+run: generate
 	$(DOCKER_COMPOSE) up --build --detach tigris_server2
 
 local_run: server
-	$(DOCKER_COMPOSE) up --no-build --detach tigris_search tigris_fdb
-	fdbcli -C ./test/config/fdb.cluster --exec "configure new single memory"
+	$(DOCKER_COMPOSE) up --no-build --detach tigris_search tigris_db
+	fdbcli -C ./test/config/fdb.cluster --exec "configure new single memory" || true
```
rbarabas, line 67: What happens when there is an error during `fdbcli` execution? Will this potentially hide it?

rbarabas:
> * Add docker file for building monocontainer development image.
> * Used docker compose plugin for local builds
> * Makefile improvements:
>   
>   * Only rebuild and restart changed containers
>   * Cache tests results (only run changed tests)
>     (all caches, volumes still can be clean and all containers restart by `make clean`)

I have given the *high level idea* of using a mono-container some thought. Is the main purpose of this change to make client checkout process simpler? 

However, I'm wondering if we have weighed the following drawbacks it may yield:
- duplicating build/maintenance effort: instead of leveraging vendor provided builds, we  duplicate FDB build effort and need to amend ours each time there is a change in upstream build
- build magnification: each time any of our component change, this image will need to be rebuilt. The more components we get the more frequent this becomes. Potential for bottleneck?
- different build/installation path: build/deployment path for local install is going to diverge with this change. Our setup is currently pretty simple but it will grow over time, increasing both size and complexity. Potential for differences in troubleshooting.

Makefile
efirs: I'll remove it. The intention was to not fail if database already exists.

scripts/install_build_deps.sh
```diff
@@ -33,15 +34,15 @@ case "${OS}-${ARCH}" in
 		;;
 	"Linux-aarch64")
 		BINARIES="apt-get curl go"
-		FDB_SHA=8173952f0aa8dabfc7da9cb23b8eff4de08831a02a3fce846fc7f996f0d3bed33588caf6f7d837ffde71ecf4b512e5e17e3af7e4c52acf838f10b3131656274e
+		FDB_SHA=c994ebb01a660cff9ef699a0e38482a561679db04c02c256efae25ba687cf903
```
ovaistariq, line 37: formatting?

scripts/install_build_deps.sh
```diff
@@ -33,15 +34,15 @@ case "${OS}-${ARCH}" in
...
 		;;
 	"Linux-arm64")
 		BINARIES="apt-get curl go"
-		FDB_SHA=8173952f0aa8dabfc7da9cb23b8eff4de08831a02a3fce846fc7f996f0d3bed33588caf6f7d837ffde71ecf4b512e5e17e3af7e4c52acf838f10b3131656274e
+		FDB_SHA=c994ebb01a660cff9ef699a0e38482a561679db04c02c256efae25ba687cf903
```
ovaistariq, line 41: formatting?

scripts/install_build_deps.sh
```diff
@@ -86,10 +87,11 @@ case "${OS}" in
 		sudo installer -pkg "$FDB_PACKAGE_PATH" -target /
 		;;
 	"Linux")
-		FDB_PACKAGE_NAME="FoundationDB-${FDB_VERSION}_${ARCH}.deb"
+		FDB_PACKAGE_NAME="foundationdb-clients_${FDB_VERSION}-1_${ARCH}.deb"
 		FDB_PACKAGE_PATH="$(mktemp -p /tmp/ -u)/${FDB_PACKAGE_NAME}"
-		curl --create-dirs -Lo "$FDB_PACKAGE_PATH" "https://tigrisdata-pub.s3.us-west-2.amazonaws.com/${FDB_PACKAGE_NAME}"
-		echo "$FDB_SHA  $FDB_PACKAGE_PATH" | sha512sum -c
+		curl --create-dirs -Lo "$FDB_PACKAGE_PATH" "https://tigrisdata-pub.s3.us-west-2.amazonaws.com/ubuntu/focal/${FDB_PACKAGE_NAME}"
```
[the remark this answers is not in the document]
efirs: Right. Fixed regular docker build to use 20.04 as well.

rbarabas (approved):
Looks good to me! Should we unify the rest of the images and make all of them use `20.04`?

efirs:
> Looks good to me! Should we unify the rest of the images and make all of them use `20.04`?

We actually will be able to move back to latest Ubuntu, once all dependencies support it.

rbarabas (approved):

PR #319, opened 2022-06-28
feat: Add Datadog tracer

rbarabas:
Instruments server with tracing per [Datadog documentation](https://docs.datadoghq.com/tracing/setup_overview/setup/go/?tabs=containers).

Will inject trace_id into logs with a later PR, I first want to get the gist of the tracing out there and validate with the infrastructure before chiseling it further.

There is probably a nicer way to compile the configuration strings, I didn't want to do the equivalent of a rouge eval() or iterate through properties of the struct but maybe latter is the answer?

Local initialization message with expected error due to me not running Datadog on my laptop:

```
2022/06/29 08:00:04 Datadog Tracer v1.38.1 WARN: DIAGNOSTICS Unable to reach agent intake: Post "http://localhost:8126/v0.4/traces": dial tcp [::1]:8126: connect: connection refused
2022/06/29 08:00:04 Datadog Tracer v1.38.1 INFO: DATADOG TRACER CONFIGURATION {"date":"2022-06-29T08:00:04-04:00","os_name":"darwin","os_version":"12.4","version":"v1.38.1","lang":"Go","lang_version":"go1.18.3","env":"","service":"___Build___Run_Server","agent_url":"http://localhost:8126/v0.4/traces","agent_error":"Post \"http://localhost:8126/v0.4/traces\": dial tcp [::1]:8126: connect: connection refused","debug":false,"analytics_enabled":false,"sample_rate":"NaN","sampling_rules":null,"sampling_rules_error":"","service_mappings":null,"tags":{"runtime-id":"8e1f97d7-6a56-4416-b5ba-d53fc4e448bf"},"runtime_metrics_enabled":false,"health_metrics_enabled":false,"profiler_code_hotspots_enabled":true,"profiler_endpoints_enabled":true,"dd_version":"","architecture":"arm64","global_service":"","lambda_mode":"false","appsec":false,"agent_features":{"DropP0s":false,"Stats":false,"StatsdPort":0}}
```


server/main.go
```diff
@@ -24,15 +24,29 @@ import (
...
 	ulog "github.com/tigrisdata/tigris/util/log"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
 )
 
+func initTracer() {
+	if !config.DefaultConfig.Tracing.Enabled {
+		return
+	}
+	tracer.Start(
+		tracer.WithUDS("/var/run/datadog/apm.socket"),
```
himank, line 35: should we move this to config as well?

server/main.go
```diff
@@ -24,15 +24,29 @@ import (
...
+	)
+	defer tracer.Stop()
```
himank, line 37: this won't work because as the function return this tracer will stop, we need to return this tracer from this method and add this defer in main.

rbarabas: I surmised that might not. Ok, let me do that!

server/main.go
rbarabas: I was contemplating about adding further options but decided to delay adding more code until I validate this works as expected. Giving it a second though it is very little work so you're right, might as well add it!

himank (approved):

PR #326, opened 2022-07-01
feat: Added CreateNamespace admin API

JigarJoshi:


server/services/v1/admin.go
```diff
@@ -0,0 +1,78 @@
...
+func newAdminService(tenantMgr *metadata.TenantManager, txMgr *transaction.Manager) *adminService {
+	return &adminService{
+		tenantMgr: tenantMgr,
+		txMgr:     txMgr,
+	}
+}
+
+func (a *adminService) CreateNamespace(ctx context.Context, req *api.CreateNamespaceRequest) (*api.CreateNamespaceResponse, error) {
+	namespace := metadata.NewTenantNamespace(req.Name, uint32(req.Id))
+	_, err := a.tenantMgr.CreateTenant(ctx, a.txMgr, namespace)
```
himank, line 54: why not start a transaction here instead of in CreateTenant and pass txObject?

server/metadata/tenant.go
```diff
@@ -166,6 +166,39 @@ func (m *TenantManager) CreateOrGetTenant(ctx context.Context, txMgr *transactio
...
+	if ulog.E(e) {
+		return nil, e
+	}
+
+	defer func() {
+		if err == nil {
+			if err = tx.Commit(ctx); err == nil {
+				// commit succeed, so we can safely cache it now, for other workers it may happen as part of the
+				// first call in query lifecycle
+				m.tenants[namespace.Name()] = tenant
```
himank, line 191: Okay, we need to cache then keep the transaction logic inside this method, ignore my comment below.

server/metadata/tenant.go
```diff
@@ -166,6 +166,39 @@ func (m *TenantManager) CreateOrGetTenant(ctx context.Context, txMgr *transactio
 	return m.createOrGetTenantInternal(ctx, tx, namespace)
 }
 
+// CreateTenant is a thread safe implementation of creating a new tenant. It returns the error if it already exists.
+func (m *TenantManager) CreateTenant(ctx context.Context, txMgr *transaction.Manager, namespace Namespace) (tenant *Tenant, err error) {
```
himank, line 170: do we need to return the tenant object from here?

server/metadata/tenant.go
```diff
@@ -166,6 +166,39 @@ func (m *TenantManager) CreateOrGetTenant(ctx context.Context, txMgr *transactio
...
+	m.Lock()
+	defer m.Unlock()
+
+	var ok bool
+	if tenant, ok = m.tenants[namespace.Name()]; ok {
+		if tenant != nil {
```
himank, line 176: if it is present in the map and the tenant is nil then it is a bug, use the tenant object directly. If it happens then it will help us to catch the bug.

server/metadata/tenant.go
himank: Actually, we don't need to cache it. Let's keep it simple. CreateTenant is only responsible for creating a tenant if it doesn't exist. Then you also don't need transaction logic inside this method and can move the transaction logic outside.

server/metadata/tenant.go
himank, line 170: also add a test for this method in tenant_test.go

server/metadata/tenant.go
efirs: So this condition could be just: `if tenant = m.tenants[namespace.Name()]; tenant != nil {
// conflict`

server/metadata/tenant.go
rbarabas: Could it be that the tenant is returned because it's potentially mutated during the call (namespace or other details added)?

server/services/v1/admin.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &api.CreateNamespaceResponse{
```
rbarabas, line 58: This type doesn't resolve for me, either, FYI. Not sure if we want to update the submodule references twice, once during development and once when API was merged and we are ready to commit this back to main?

server/services/v1/admin.go
rbarabas: +1

server/services/v1/admin.go
```diff
@@ -0,0 +1,78 @@
...
+		return err
+	}
+	router.HandleFunc(adminPath+namespacePathPattern, func(w http.ResponseWriter, r *http.Request) {
+		mux.ServeHTTP(w, r)
+	})
+	return nil
+}
+
+func (h *adminService) RegisterGRPC(grpc *grpc.Server) error {
+	api.RegisterAdminServer(grpc, h)
```
rbarabas, line 76: Another REF/declaration issue

server/metadata/tenant.go
JigarJoshi: ok moved the caching part out of this method.

server/services/v1/admin.go
JigarJoshi: updated it to start transaction from handler.

server/metadata/tenant.go
JigarJoshi: updated the signature to not return tenant.

server/metadata/tenant.go
JigarJoshi: updated it.

server/metadata/tenant.go
JigarJoshi: done.

server/services/v1/admin.go
```diff
@@ -0,0 +1,78 @@
...
+const (
+	adminPath = "/admin/" + version
+)
+const (
+	namespacePath        = "/namespaces"
+	namespacePathPattern = namespacePath + "/*"
+)
+
+type adminService struct {
+	api.UnimplementedAdminServer
```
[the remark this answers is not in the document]
JigarJoshi: did you try `make generate`? if it still persists, try invalidating IDE cache.

server/services/v1/admin.go
JigarJoshi: did you try make generate? if it still persists, try invalidating IDE cache.

server/services/v1/admin.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return &api.CreateNamespaceResponse{
+		Status:  "CREATED",
+		Message: "Namespace created, with id=" + fmt.Sprint(req.Id) + ", and name=" + req.Name,
+	}, nil
+}
+
+func (h *adminService) RegisterHTTP(router chi.Router, _ *inprocgrpc.Channel) error {
+	mux := runtime.NewServeMux(runtime.WithMarshalerOption(string(JSON), &runtime.JSONBuiltin{}))
+	if err := api.RegisterAdminHandlerServer(context.TODO(), mux, h); err != nil {
```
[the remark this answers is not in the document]
JigarJoshi: same as above.

server/services/v1/admin.go
JigarJoshi: same as above.

server/metadata/tenant.go
```diff
@@ -166,6 +166,24 @@ func (m *TenantManager) CreateOrGetTenant(ctx context.Context, txMgr *transactio
 	return m.createOrGetTenantInternal(ctx, tx, namespace)
 }
 
+// CreateTenant is a thread safe implementation of creating a new tenant. It returns the error if it already exists.
+func (m *TenantManager) CreateTenant(ctx context.Context, tx transaction.Tx, namespace Namespace) (err error) {
```
himank, line 170: no need to have a named return value here.

server/metadata/tenant.go
```diff
@@ -166,6 +166,24 @@ func (m *TenantManager) CreateOrGetTenant(ctx context.Context, txMgr *transactio
...
+
+	if id, found := namespaces[namespace.Name()]; found {
+		return api.Errorf(api.Code_CONFLICT, "namespace with same name already exists with id '%s'", fmt.Sprint(id))
+	}
+	for name, id := range namespaces {
+		if id == namespace.Id() {
+			return api.Errorf(api.Code_CONFLICT, "namespace with same id already exists with name '%s'", name)
+		}
+	}
+	_, err = m.createOrGetTenantInternal(ctx, tx, namespace)
```
himank, line 183: as you are not returning the tenant don't call this as this method is also creating a tenant object which is used to cache it. Replace this by executing the version bump and encoding the tenant information like this,
```
log.Debug().Str("tenant", namespace.Name()).Msg("tenant not found, creating")

// bump the version first
if err := m.versionH.Increment(ctx, tx); ulog.E(err) {
	return nil, err
}

if err := m.encoder.ReserveNamespace(ctx, tx, namespace.Name(), namespace.Id()); ulog.E(err) {
	return nil, err
}
```

JigarJoshi: updated.

server/metadata/tenant.go
JigarJoshi: fixed.

himank (approved):

PR #327, opened 2022-07-01
Added machinery to set outgoing headers. Allow CORS by default

efirs:


server/midddleware/auth.go
```diff
@@ -27,6 +27,12 @@ import (
 	"github.com/tigrisdata/tigris/server/config"
 )
 
+type key string
```
himank, line 30: do we need to have this? or can we just have `tokenKey string = "token"`

efirs: Help says it's better to define a type for the context key.
```
// The provided key must be comparable and should not be of type
// string or any other built-in type to avoid collisions between
// packages using context. Users of WithValue should define their own
// types for keys.
```

himank: if you are defining because of this then the usual pattern is to not use string but define it like this,
```
type TokenCtxKey struct{}
```
and then you set it.

himank (dismissed):

server/midddleware/auth.go
efirs: Fixed.

himank (approved):

PR #329, opened 2022-07-01
feat: Support querying nested fields

himank:


schema/collection.go
```diff
@@ -51,9 +52,11 @@ type DefaultCollection struct {
 	Validator *jsonschema.Schema
 	// JSON schema
 	Schema jsoniter.RawMessage
-
 	// search schema
 	Search *tsApi.CollectionSchema
+	// QueryableFields are similar to fields but these are flattened form of fields i.e. for an object there may be more
```
rbarabas, line 57: Nit: since you are comparing the types, this should probably say "are similar to Fields", with an upper case.

Consider: "QueryableFields are similar to Fields but they represent them in a flattened form. In other words, for an object there...

schema/collection.go
```diff
@@ -73,17 +76,18 @@ func NewDefaultCollection(name string, id uint32, schVer int, fields []*Field, i
 	// flag. Later probably we can relax it. Starting with strict validation is better than not validating extra keys.
```
rbarabas, line 76: Nit: Comment could be improved. 

// Allows the caller to pass additional properties. This feature is currently disabled as it is better to be strict than to not validate and extra key. This constraint will most likely be relaxed later.

rbarabas (commented):
Where is the function `PackSearchField` used? Should it be ported to use `FieldType` or if not used, removed?

rbarabas (approved):
LGTM. Left minor remarks, nits only.

himank:
> LGTM. Left minor remarks, nits only.



> Where is the function `PackSearchField` used? Should it be ported to use `FieldType` or if not used, removed?

Some cleanup remaining, good catch. Let me remove it and fix all the comments :)

server/services/v1/search_indexer.go
```diff
@@ -239,3 +242,40 @@ func UnpackAndSetMD(doc map[string]interface{}, tableData *internal.TableData) {
...
+			}
+		default:
+			resp[k] = v
+		}
+	}
+
+	return resp
+}
+
+func UnFlattenObjects(flat map[string]any) map[string]any {
```
efirs, line 263: I think we can do it more efficient, and avoid dependency on `mergo`:
```
result := make(map[string]any)

for k, v := range flat {
   keys := strings.Split(k, schema.ObjFlattenDelimiter)
   m := result // top level map always exists
   for i := 1; i < len(keys); i++ { // create submaps if necessary
      if _, ok := m[keys[i-1]; !ok {
         m[keys[i-1] = make(map[string]any)
      }
      m = map[keys[i-1]]
   }
   // here m contains last submap
   m[keys[len(keys)-1] = v
}
```

himank: let me change it.

server/services/v1/search_indexer.go
```diff
@@ -239,3 +240,45 @@ func UnpackAndSetMD(doc map[string]interface{}, tableData *internal.TableData) {
...
+			flattenObjects(k, vType, resp)
+		default:
+			resp[k] = v
+		}
+	}
+
+	return resp
+}
+
+func flattenObjects(key string, obj map[string]any, resp map[string]any) {
```
efirs, line 258: If you add this:
```
if key != "" {
   key += schema.ObjFlattenDelimiter
}
```
you can replace below key to be `key+k` and call this function in the above directly, no need in additional loop.

efirs (approved):

PR #331, opened 2022-07-01
Refactor tracing

rbarabas:
Contains the following changes:
- Turns tracing and profiling OFF by default
- Lets you explicitly enable tracing and profiling both through ENV variables and the YAML configuration
- Moves tracing into its own package under `server/`.
- Ensures proper tagging is configurable for profiling

[Used debug images for validation](https://hub.docker.com/repository/docker/tigrisdata/tigrisdebug/tags?page=1&ordering=last_updated).

*** After deployment *** 

Relevant env:
```
tigris@tigris-server-7c586ccfc9-9cdn4:/server$ env | grep ^TIGRIS_SERVER
TIGRIS_SERVER_TRACING_AGENT_ADDR=status.hostIP:8126
TIGRIS_SERVER_TRACING_ENABLED=true
TIGRIS_SERVER_PROFILING_ENABLED=true
TIGRIS_ENVIRONMENT=dev
```

Initialization logs from the `dev` cluster deployment (see corresponding infra PR):
```
2022/07/01 21:09:01 Datadog Tracer v1.38.1 INFO: DATADOG TRACER CONFIGURATION {"date":"2022-07-01T21:09:01Z","os_name":"Linux (Unknown Distribution)","os_version":"20.04.4 LTS (Focal Fossa)","version":"v1.38.1","lang":"Go","lang_version":"go1.18.3","env":"local","service":"tigris-server","agent_url":"http://10.0.10.147:8126/v0.4/traces","agent_error":"","debug":false,"analytics_enabled":false,"sample_rate":"NaN","sampling_rules":null,"sampling_rules_error":"","service_mappings":null,"tags":{"runtime-id":"66e32c72-f7bd-4670-b468-58fcfaff69a6"},"runtime_metrics_enabled":false,"health_metrics_enabled":false,"profiler_code_hotspots_enabled":true,"profiler_endpoints_enabled":true,"dd_version":"99af934","architecture":"amd64","global_service":"tigris-server","lambda_mode":"false","appsec":false,"agent_features":{"DropP0s":true,"Stats":true,"StatsdPort":0}}
```

```
2022/07/01 21:09:01 Datadog Tracer v1.38.1 INFO: Profiler configuration: {"date":"2022-07-01T21:09:01Z","os_name":"Linux (Unknown Distribution)","os_version":"20.04.4 LTS (Focal Fossa)","version":"v1.38.1","lang":"Go","lang_version":"go1.18.3","hostname":"tigris-server-7c5fc8b577-m2sl5","delta_profiles":true,"service":"tigris-server","env":"local","target_url":"http://10.0.10.147:8126/profiling/v1/input","agentless":false,"tags":["pid:1","profiler_version:v1.38.1","runtime_version:1.18.3","runtime_compiler:gc","runtime_arch:amd64","runtime_os:linux","runtime-id:66e32c72-f7bd-4670-b468-58fcfaff69a6","version:99af934"],"profile_period":"1m0s","enabled_profiles":["metrics","cpu","heap"],"cpu_duration":"1m0s","cpu_profile_rate":0,"block_profile_rate":100000000,"mutex_profile_fraction":10,"max_goroutines_wait":1000,"upload_timeout":"10s"}
```


server/config/options.go
```diff
@@ -17,6 +17,8 @@ package config
 import (
 	"time"
 
+	"github.com/tigrisdata/tigris/util"
+
```
ovaistariq, line 21: Remove new line between the imports

server/config/options.go
```diff
@@ -93,13 +102,18 @@ var DefaultConfig = Config{
 		WriteEnabled: true,
 	},
 	Tracing: TracingConfig{
-		Enabled: true,
+		Enabled: false,
```
ovaistariq, line 105: So we are keeping it disabled by default?

.github/workflows/push-docker-image-debug.yaml
```diff
@@ -0,0 +1,73 @@
+name: publish docker image
```
ovaistariq, line 1: Why do we need this?

server/config/options.go
rbarabas: Yes as the profiler would log recurring errors on developers laptops. Those laptops would likely have no opentracing or datadog tracing sink installed.

.github/workflows/push-docker-image-debug.yaml
rbarabas: To iterate with concepts like this much quicker when we need to validate how the codebase interacts with deployments.

rbarabas: I just found an issue with the deployed debug image that would have ended up as a separate PR / review / main / merge loop otherwise!

server/config/options.go
rbarabas: Is there a linter / formatter where we could codify our preferences for imports? I got the opposite recommendation yesterday! :- ) I'm using `goimports` to format this file btw, not sure if that applies to whitespaces or not.

.github/workflows/push-docker-image-debug.yaml
rbarabas: I intend to introduce this file as part of a separate PR though, I just needed this idea to quickly iterate through images while testing. Removed from the PR!

server/config/options.go
rbarabas: @efirs do you have a preference for this?

server/config/options.go
```diff
@@ -51,19 +54,25 @@ type CdcConfig struct {
...
 
 type ProfilingConfig struct {
-	Enabled         bool
-	EnableCPU       bool
-	EnableHeap      bool
-	EnableBlock     bool
-	EnableMutex     bool
-	EnableGoroutine bool
+	Enabled         bool `mapstructure:"enabled" yaml:"enabled" json:"enabled"`
+	EnableCPU       bool `mapstructure:"enable_cpu" yaml:"enable_cpu" json:"enable_cpu"`
```
rbarabas, line 65: I'm not sure if these are necessary when there are no significant changes to the pattern?

server/config/options.go
ovaistariq: We use goimports. But it shouldn't have put a newline here. Newline is needed between new groups of imports.

efirs: Yeah, combine them.
Basically we only want to separate Go's "builtin" packages into separate section. "time" in this case.

efirs: goimports will automatically sort imports in the same group.

server/config/options.go
```diff
@@ -51,19 +54,25 @@ type CdcConfig struct {
 }
 
 type TracingConfig struct {
-	Enabled           bool
-	WithUDS           string
-	WithAgentAddr     string
-	WithDogStatsdAddr string
+	Enabled           bool   `mapstructure:"enabled" yaml:"enabled" json:"enabled"`
```
ovaistariq, line 57: What will be the use of mapstructure tag here?

server/tracing/tracing.go
```diff
@@ -0,0 +1,52 @@
...
+
+func InitTracer(config *config.Config) error {
+	if !config.Profiling.Enabled {
+		return nil
+	}
+
+	tracer.Start(getTracingOptions(config)...)
+	defer tracer.Stop()
+
+	if config.Profiling.Enabled {
```
efirs, line 44: you don't need this `if` it has been take care of on line 37

server/tracing/tracing.go
```diff
@@ -0,0 +1,52 @@
...
+	return opts
+}
+
+func InitTracer(config *config.Config) error {
+	if !config.Profiling.Enabled {
+		return nil
+	}
+
+	tracer.Start(getTracingOptions(config)...)
+	defer tracer.Stop()
```
efirs, line 42: How about these tracer and profiler defers?
you probably need to pass them up to the main.
`return func() { trace.Stop(); profiler.Stop() }, nil`
and InitTracer function would return `(func(), error)`.

server/config/options.go
rbarabas: My understanding is that this enables the key to be used as an environment variable with `spf13/cobra` and `AutoEnv()`.

server/tracing/tracing.go
efirs: actually `func() { ulog.E(trace.Stop()); ulog.E(profiler.Stop()) }`
if `Stop`s return error

server/tracing/tracing.go
rbarabas: You actually caught a 🐞 here. L37 should say `config.Tracing.Enabled`! Great catch!

server/tracing/tracing.go
rbarabas: Thanks for the tip!

server/config/options.go
himank: I don't think we need all these knobs here if it is enabled then in most of the cases we want to profile different things like CPU, Heap, etc. Just have one config key Enabled.

server/tracing/tracing.go
himank: yeah, this tracer needs to be returned so that it can be stopped in main.

server/tracing/tracing.go
efirs: It looks correct to me. if not enabled return.

server/config/options.go
```diff
@@ -51,19 +53,25 @@ type CdcConfig struct {
...
-	Enabled         bool
-	EnableCPU       bool
-	EnableHeap      bool
-	EnableBlock     bool
-	EnableMutex     bool
-	EnableGoroutine bool
+	Enabled         bool `mapstructure:"enabled" yaml:"enabled" json:"enabled"`
+	EnableCPU       bool `mapstructure:"enable_cpu" yaml:"enable_cpu" json:"enable_cpu"`
+	EnableHeap      bool `mapstructure:"enable_heap" yaml:"enable_heap" json:"enable_heap"`
+	EnableBlock     bool `mapstructure:"enable_block" yaml:"enable_block" json:"enable_block"`
```
himank, line 66: should we remove these keys? and just have Enabled?

server/main.go
```diff
@@ -21,48 +21,28 @@ import (
...
-	tracer.Start(getTracingOptions()...)
-	defer tracer.Stop()
+	log.Info().Msgf("Environment: %v\n", config.GetEnvironment())
 
-	if config.DefaultConfig.Profiling.Enabled {
-		if err := profiler.Start(); err != nil {
-			ulog.E(err)
-		}
-		defer profiler.Stop()
+	deferTracerAndProfiler, err := tracing.InitTracer(&config.DefaultConfig)
```
himank, line 41: nit, call it closer

server/config/options.go
rbarabas: I would keep them as options for configuration. Some of these are high overhead but could come handy during troubleshooting.

himank: Sure, but in most cases, we always want these five to be enabled.

rbarabas: For Production deployments I'd only keep CPU and HEAP enabled, these are what I'd call our safe defaults for performance reasons. For development / local deployments, I'd enable all five, due to the overhead they incur. This means we could disable the switches for CPU and HEAP as they will be in most deployments and keep the "extras" configurable if you think that's a better approach.

himank: So there are two parts to it I think, 

- Go Pprof: https://pkg.go.dev/net/http/pprof
- Datadog Profiler

I am assuming when you say expensive it is the datadog profiling? If yes, that makes sense. Is that also enabling go pprof by default? or do we need to do something to enable it? For Go pprof, we can enable all and for datadog profiler, we can go with these config knobs.

himank (approved):

server/config/options.go
rbarabas: Ah, I believe I now see where you are coming from. Thanks for the extra context! 

Yes, what I said was applicable only for Datadog profiling. 

To confirm if I got your point correctly, let me ask back if I understood you right! Are we suggesting that the tracing and profiling configuration should be generic options and we should figure out what is applicable for the particular deployment? In other words, if we enable CPU profiling, it could enable Datadog CPU Profiling, `go pprof` CPU profiling and other frameworks?

server/config/config.go
```diff
@@ -44,31 +44,35 @@ var envEnv = "tigris_environment"
...
 
 func GetEnvironment() string {
 	return environment
 }
 
 func LoadEnvironment() {
-	env := os.Getenv(envEnv)
-	if env == "" {
-		env = os.Getenv(strings.ToUpper(envEnv))
+	resolutionOrder := []string{
```
ovaistariq, line 59: Nice, now even reading the code I can figure out how things are loaded.

server/config/options.go
himank: So the Go's pprof is cheap as the handler only gets executed when the request is made for the profiling. I would suggest keeping these options only for the datadog related profiling(which is what you have now) and regarding the go proof once we add it will be enabled for profiling.

PR #334, opened 2022-07-05
feat: Add workflow for building debug images

rbarabas:
Adds a new workflow that will allow us to publish ad-hoc debug images for deployment validation and integration tests.

This workflow will create images if the PR is labeled with `debug image`. The resultant image will be published to `tigrisdata/tigris-debug`.

Before label:

Screen Shot 2022-07-05 at 6 48 09 PM

After label:

Screen Shot 2022-07-05 at 6 49 02 PM

See [DockerHub](https://hub.docker.com/repository/docker/tigrisdata/tigrisdebug) for an example image that was created during #331, triggered using the branch name.

### DIFF
--- a/.github/workflows/publish-debug-image.yaml
+++ b/.github/workflows/publish-debug-image.yaml
@@ -1,0 +1,71 @@
+name: publish debug image
+
+on:
+  pull_request:
+    types:
+      - labeled
+  push:
+  workflow_call:
+    secrets:
+      GH_DOCKER_ACCESS_USER:
+        required: true
+      GH_DOCKER_ACCESS_TOKEN:
+        required: true
+
+env:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-debug-image.yaml
```diff
@@ -0,0 +1,71 @@
...
+  push:
+  workflow_call:
+    secrets:
+      GH_DOCKER_ACCESS_USER:
+        required: true
+      GH_DOCKER_ACCESS_TOKEN:
+        required: true
+
+env:
+  DOCKER_REPOSITORY: "tigrisdata/tigrisdebug"
```
efirs, line 16: maybe `tigrisdata/tigris-debug`

PR #335, opened 2022-07-06
feat: ability to specify config as an argument

rbarabas:
Allows one to specify the configuration file to use with server. This makes it easy to specify custom configuration files for unique deployment scenarios.

***NOTE*** I removed an the old pflag.String() cruft that I believe was not in use. Please let me know if that's not the case! 

***SYNOPSIS***
```
$ ./server/service --help
Usage of ./server/service:
  -c, --config string   server configuration file
pflag: help requested

$ ./server/service -c config/server.dev.yaml
```

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -20,6 +20,8 @@ import (
 	"os/exec"
 	"strings"
 
+	ulog "github.com/tigrisdata/tigris/util/log"
+
 	"github.com/spf13/pflag"
 
 	"github.com/davecgh/go-spew/spew"
@@ -35,6 +37,8 @@ var configPath = []string{
 	"./config/",
 	"./",
 }
+
+var configFile string
 
 // envPrefix is used by viper to detect environment variables that should be used.
 // viper will automatically uppercase this and append _ to it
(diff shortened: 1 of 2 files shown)


ovaistariq (approved):

server/config/config.go
```diff
@@ -70,16 +72,23 @@ func LoadEnvironment() {
...
-	LoadEnvironment()
-
-	log.Info().Msgf("TIGRIS_ENVIRONMENT=%s", GetEnvironment())
-
-	viper.SetConfigName(name)
-	viper.SetConfigType("yaml")
+func LoadConfig(config interface{}) {
+	pflag.StringVarP(&configFile, "config", "c", "", "server configuration file")
+	if err := viper.BindPFlags(pflag.CommandLine); err != nil {
+		log.Err(err).Msg("bind flags")
```
efirs, line 78: do we ignore error here?
if so we can just put `ulog.E(viper.BindPFlags(pflag.CommandLine))` instead of it.

rbarabas: Updated! Thanks for the advice!

pboros (approved):

PR #342, opened 2022-07-08
feat: Add namespace based authz

JigarJoshi:
TODO:

- [x] setup context for stream server interceptor
- [ ] make the cache reload when namespace is created
- [x] add tests.
- [x] check if there is anything needs to be done for forwarders
- [x] fix tests
- [x] fix lint warnings

### DIFF
--- a/server/metadata/tenant.go
+++ b/server/metadata/tenant.go
@@ -167,30 +167,37 @@ func (m *TenantManager) CreateOrGetTenant(ctx context.Context, txMgr *transactio
 }
 
 // CreateTenant is a thread safe implementation of creating a new tenant. It returns the error if it already exists.
-func (m *TenantManager) CreateTenant(ctx context.Context, tx transaction.Tx, namespace Namespace) error {
+func (m *TenantManager) CreateTenant(ctx context.Context, tx transaction.Tx, namespace Namespace) (Namespace, error) {
 	m.Lock()
 	defer m.Unlock()
 	namespaces, err := m.encoder.GetNamespaces(ctx, tx)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	if id, found := namespaces[namespace.Name()]; found {
-		return api.Errorf(api.Code_CONFLICT, "namespace with same name already exists with id '%s'", fmt.Sprint(id))
+		return nil, api.Errorf(api.Code_CONFLICT, "namespace with same name already exists with id '%s'", fmt.Sprint(id))
 	}
 	for name, id := range namespaces {
 		if id == namespace.Id() {
-			return api.Errorf(api.Code_CONFLICT, "namespace with same id already exists with name '%s'", name)
+			return nil, api.Errorf(api.Code_CONFLICT, "namespace with same id already exists with name '%s'", name)
 		}
 	}
-	if err := m.versionH.Increment(ctx, tx); ulog.E(err) {
-		return err
-	}
-
 	if err := m.encoder.ReserveNamespace(ctx, tx, namespace.Name(), namespace.Id()); ulog.E(err) {
-		return err
+		return nil, err
 	}
+	if err := m.versionH.Increment(ctx, tx); ulog.E(err) {
+		return nil, err
+	}
+	return namespace, nil
+}
 
+func (m *TenantManager) GetNamespace(namespaceName string) Namespace {
+	m.RLock()
+	defer m.RUnlock()
+	if tenant, found := m.tenants[namespaceName]; found {
+		return tenant.namespace
+	}
 	return nil
 }
 
--- a/server/midddleware/auth.go
+++ b/server/midddleware/auth.go
@@ -33,7 +33,7 @@ var (
 	headerAuthorize = "authorization"
 )
 
-type Organization struct {
+type Namespace struct {
 	Code string `json:"code"`
 }
 
@@ -42,22 +42,13 @@ type User struct {
 }
 
 type CustomClaim struct {
-	Roles []string     `json:"https://tigris-db-api/r"`
-	Org   Organization `json:"https://tigris-db-api/o"`
-	User  User         `json:"https://tigris-db-api/u"`
+	Namespace Namespace `json:"https://tigris-db-api/n"`
+	User      User      `json:"https://tigris-db-api/u"`
 }
 
-func (c CustomClaim) Validate(ctx context.Context) error {
-	// org code claim verification
-	orgName, err := getOrganizationName(ctx)
-	if err != nil {
-		return err
-	}
-	if len(c.Org.Code) == 0 {
-		return api.Errorf(api.Code_PERMISSION_DENIED, "empty organization code in token")
-	}
-	if orgName != c.Org.Code {
-		return api.Errorf(api.Code_PERMISSION_DENIED, "your token is not valid for this URL")
+func (c CustomClaim) Validate(_ context.Context) error {
+	if len(c.Namespace.Code) == 0 {
+		return api.Errorf(api.Code_PERMISSION_DENIED, "empty namespace code in token")
 	}
 	return nil
 }
@@ -115,32 +106,21 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
 		return ctx, err
 	}
 
-	validToken, err := jwtValidator.ValidateToken(ctx, tkn)
+	validatedToken, err := jwtValidator.ValidateToken(ctx, tkn)
 	if err != nil {
 		return ctx, api.Errorf(api.Code_UNAUTHENTICATED, err.Error())
 	}
 
-	log.Debug().Msg("Valid token received")
-	return context.WithValue(ctx, TokenCtxkey{}, validToken), nil
-}
-
-func getOrganizationName(ctx context.Context) (string, error) {
-	host := api.GetHeader(ctx, ":authority")
-	if host == "" {
-		host = api.GetHeader(ctx, "host")
+	if validatedClaims, ok := validatedToken.(*validator.ValidatedClaims); ok {
+		if customClaims, ok := validatedClaims.CustomClaims.(*CustomClaim); ok {
+			log.Debug().Msg("Valid token received")
+			token := &AccessToken{
+				Namespace: customClaims.Namespace.Code,
+				Sub:       validatedClaims.RegisteredClaims.Subject,
+			}
+			return setAccessToken(ctx, token), nil
+		}
 	}
-	// <project>-<org-name>.<env>.tigrisdata.cloud
-	parts := strings.Split(host, ".")
-	if len(parts) < 3 {
-		return "", api.Errorf(api.Code_FAILED_PRECONDITION, "hostname is not as per expected scheme")
-	}
-	subParts := strings.Split(parts[0], "-")
-	if len(subParts) > 2 {
-		return "", api.Errorf(api.Code_FAILED_PRECONDITION, "hostname is not as per expected scheme")
-	}
-	organizationName := subParts[0]
-	if len(subParts) > 1 {
-		organizationName = subParts[1]
-	}
-	return organizationName, nil
+	// this should never happen.
+	return ctx, api.Errorf(api.Code_UNAUTHENTICATED, "You are not authorized to perform this action")
 }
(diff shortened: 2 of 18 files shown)


server/metadata/access_token.go
```diff
@@ -0,0 +1,8 @@
+package metadata
```
himank, line 1: do we need it here? should we move it to server package?

server/metadata/namespace.go
```diff
@@ -0,0 +1,38 @@
+package metadata
```
himank, line 1: So instead of having it specific to the namespace, should we just have a generic class in server package `RequestMetadata` or `ContextMetadata` etc and that will have a bunch of methods to extract the values of different keys in the context that is needed in the server package. This way nothing changes in the metadata package and once you get the namespace name then sessions.go will simply follow the same process that it is doing now.

server/metadata/namespace.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+	"fmt"
+
+	api "github.com/tigrisdata/tigris/api/server/v1"
+	"github.com/tigrisdata/tigris/server/transaction"
+)
+
+// NamespaceExtractor - extract the namespace from context
+type NamespaceExtractor interface {
```
himank, line 12: this interface if needed can also live in server package which has one implementation in the above class ^

server/midddleware/middleware.go
```diff
@@ -26,17 +26,26 @@ import (
...
 
-func Get(config *config.Config) (grpc.UnaryServerInterceptor, grpc.StreamServerInterceptor) {
+func Get(config *config.Config, tenantMgr *metadata.TenantManager, txMgr *transaction.Manager) (grpc.UnaryServerInterceptor, grpc.StreamServerInterceptor) {
 	jwtValidator := GetJWTValidator(config)
 	// inline closure to access the state of jwtValidator
 	authFunction := func(ctx context.Context) (context.Context, error) {
 		return AuthFunction(ctx, jwtValidator, config)
 	}
 
+	namespaceInitializer := NamespaceInitializer{
```
himank, line 42: is it an `Initializer`? or is it more like a `setter`? should we call it `namespaceSetter`

server/metadata/namespace.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (tokenNamespaceExtractor *AccessTokenNamespaceExtractor) Extract(ctx context.Context) (Namespace, error) {
+	// read token
+	value := ctx.Value(AccessTokenCtxKey{})
+	if value != nil {
+		if token, ok := value.(*AccessToken); ok {
+			// check from cache
+			namespace := tokenNamespaceExtractor.TenantManager.getNamespace(token.Namespace)
+			fmt.Println("Namespace from token = " + token.Namespace)
```
himank, line 31: remove it once done with debugging.

server/metadata/namespace.go
```diff
@@ -0,0 +1,38 @@
...
+
+type NamespaceCtxKey struct {
+}
+
+type AccessTokenNamespaceExtractor struct {
+	TenantManager      *TenantManager
+	Transactionmanager *transaction.Manager
+}
+
+func (tokenNamespaceExtractor *AccessTokenNamespaceExtractor) Extract(ctx context.Context) (Namespace, error) {
```
himank, line 24: Once we move the class outside of this package then `Extract` should only be responsible for extracting the namespace name. In sessions.go, you'll then use this name to ask the tenant if it exists(if we need to do strict checking) and returns an error if it doesn't exist.

server/metadata/access_token.go
JigarJoshi: you mean next to main.go?  https://github.com/tigrisdata/tigris/tree/main/server

himank: Sorry i meant in service package

server/metadata/namespace.go
JigarJoshi: there are admin apis, which will be at cluster level and namespace will be nil for those, do you recommend setting default namespace for those in context from middleware. This way we never expect nil namespace in context in sessions.go?

himank: That makes sense, we can set the default from outside. Then we can do strict check in the sessions.go

server/metadata/tenant.go
```diff
@@ -167,30 +167,35 @@ func (m *TenantManager) CreateOrGetTenant(ctx context.Context, txMgr *transactio
...
 	if err := m.versionH.Increment(ctx, tx); ulog.E(err) {
-		return err
+		return nil, err
 	}
+	return namespace, nil
+}
 
-	if err := m.encoder.ReserveNamespace(ctx, tx, namespace.Name(), namespace.Id()); ulog.E(err) {
-		return err
+func (m *TenantManager) GetNamespace(namespaceName string) Namespace {
```
himank, line 195: As it is exposed we need to take read lock.

server/midddleware/auth.go
```diff
@@ -115,32 +107,20 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
...
-	if len(subParts) > 1 {
-		organizationName = subParts[1]
+	if validatedClaims, ok := validatedToken.(*validator.ValidatedClaims); ok {
+		if customClaims, ok := validatedClaims.CustomClaims.(*CustomClaim); ok {
+			log.Debug().Msg("Valid token received")
+			token := v1.AccessToken{
+				Namespace: customClaims.Namespace.Code,
+				Sub:       validatedClaims.RegisteredClaims.Subject,
+			}
+			result := v1.SetAccessToken(ctx, token)
```
himank, line 121: nit: `return v1.SetAccessToken(ctx, token), nil`

server/midddleware/auth.go
```diff
@@ -115,32 +107,20 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
...
+			return result, nil
+		}
 	}
-	return organizationName, nil
+	return ctx, api.Errorf(api.Code_UNAUTHENTICATED, err.Error())
```
himank, line 125: there is no error at this point i.e. err obj is nil. Don't we need to send some custom message here?

server/midddleware/namespace.go
```diff
@@ -0,0 +1,43 @@
+package middleware
+
+import (
+	"context"
+
+	middleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	api "github.com/tigrisdata/tigris/api/server/v1"
+	"github.com/tigrisdata/tigris/server/metadata"
+	"github.com/tigrisdata/tigris/server/services/v1"
```
himank, line 9: middleware shouldn't be dependent on the services/v1 package

server/services/v1/request_metadata.go
```diff
@@ -0,0 +1,101 @@
...
+	var result = ctx
+	if err != nil && requestMetadata == nil {
+		requestMetadata = &RequestMetadata{}
+		result = context.WithValue(ctx, RequestMetadataCtxKey{}, requestMetadata)
+	}
+	requestMetadata.accessToken = token
+	return result
+}
+
+func SetNamespace(ctx context.Context, namespace metadata.Namespace) context.Context {
```
himank, line 56: so any Set on the context object shouldn't be here as that is done by middleware, only have getters in this file.

server/services/v1/request_metadata.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type RequestMetadata struct {
+	accessToken AccessToken
+	namespace   metadata.Namespace
+}
+
+// NamespaceExtractor - extract the namespace from context
+type NamespaceExtractor interface {
+	Extract(ctx context.Context) (metadata.Namespace, error)
```
himank, line 26: Do we need to return the object, or should we just Set/Get the namespace name? That should be enough for us to get the tenant in sessionMgr.

server/metadata/namespace.go
JigarJoshi: added a wrapper RequestMetadata type.

server/midddleware/auth.go
JigarJoshi: at this point, the claims in token are not validataed. so user is still unauthorized. Good point, let me update message.

server/midddleware/request_metadata.go
```diff
@@ -0,0 +1,101 @@
...
+
+func GetRequestMetadata(ctx context.Context) (*RequestMetadata, error) {
+	// read token
+	value := ctx.Value(RequestMetadataCtxKey{})
+	if value != nil {
+		if requestMetadata, ok := value.(*RequestMetadata); ok {
+			return requestMetadata, nil
+		}
+	}
+	return nil, api.Errorf(api.Code_NOT_FOUND, "request metadata not found 1")
```
himank, line 42: any reason of adding `1` in the error string?

server/midddleware/request_metadata.go
```diff
@@ -0,0 +1,101 @@
...
+}
+func GetAccessToken(ctx context.Context) (*AccessToken, error) {
+	// read token
+	value := ctx.Value(RequestMetadataCtxKey{})
+	if value != nil {
+		if requestMetadata, ok := value.(*RequestMetadata); ok {
+			return &requestMetadata.accessToken, nil
+		}
+	}
+	return nil, api.Errorf(api.Code_NOT_FOUND, "request metadata not found 2")
```
himank, line 74: If we need to return an error from here then it should be that `access token not found`.

server/midddleware/request_metadata.go
```diff
@@ -0,0 +1,101 @@
...
+}
+func GetNamespace(ctx context.Context) (metadata.Namespace, error) {
+	// read token
+	value := ctx.Value(RequestMetadataCtxKey{})
+	if value != nil {
+		if requestMetadata, ok := value.(*RequestMetadata); ok {
+			return requestMetadata.namespace, nil
+		}
+	}
+	return nil, api.Errorf(api.Code_NOT_FOUND, "request metadata not found 3")
```
himank, line 84: `namespace not found in request token`

server/midddleware/auth.go
himank: that is fine but my point was there is no JWT error in the method at this point because there is a check-in line 110 and based on that you're already returning. At this point, the error needed to be a custom msg.

himank (dismissed):

server/midddleware/request_metadata.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (tokenNamespaceExtractor *AccessTokenNamespaceExtractor) Extract(ctx context.Context) (metadata.Namespace, error) {
+	// read token
+	token, err := GetAccessToken(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if token != nil {
+		namespace := tokenNamespaceExtractor.TenantManager.GetNamespace(token.Namespace)
```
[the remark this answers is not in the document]
JigarJoshi: makes sense. fixed it.

server/midddleware/request_metadata.go
JigarJoshi: fixed. (it was left over from debugging)

server/midddleware/request_metadata.go
JigarJoshi: fixed. (it was left over from debugging)

server/services/v1/request_metadata.go
JigarJoshi: updated.

server/services/v1/request_metadata.go
JigarJoshi: updated.

server/midddleware/namespace.go
JigarJoshi: fixed.

server/midddleware/auth.go
JigarJoshi: fixed.

server/metadata/tenant.go
JigarJoshi: added read lock.

server/metadata/namespace.go
JigarJoshi: 👍

server/metadata/namespace.go
JigarJoshi: fixed.

server/midddleware/auth.go
JigarJoshi: yes this line should never be invoked. updated message and comment.

himank (approved):

JigarJoshi:
Currently the namespace creation doesn't refresh the cache, so server still says namespace doesn't exist. This will be fixed in later PR.

PR #343, opened 2022-07-08
fix: optimize checking of versions during dml requests

himank:


server/main.go
```diff
@@ -26,14 +26,15 @@ import (
...
+	"runtime"
 )
 
 func main() {
 	config.LoadConfig(&config.DefaultConfig)
-
 	ulog.Configure(config.DefaultConfig.Log)
 
 	log.Info().Msgf("Environment: %v\n", config.GetEnvironment())
+	log.Info().Msgf("running with number of CPUs: %v", runtime.NumCPU())
```
efirs, line 37: Just `Number of CPUs:`

server/metadata/tenant.go
```diff
@@ -374,22 +377,31 @@ func (tenant *Tenant) ReloadUsingOutsideVersion(ctx context.Context, tx transact
...
-	tenant.Lock()
-	defer tenant.Unlock()
-
 	currentVersion, err := tenant.versionH.Read(ctx, tx)
 	if err != nil {
 		return err
 	}
 
-	if bytes.Equal(currentVersion, tenant.version) {
+	if !tenant.shouldReload(currentVersion) {
```
efirs, line 385: What if the version changes after this check? We don't need to reload?

store/kv/fdb.go
```diff
@@ -109,20 +109,23 @@ func (d *fdbkv) txWithRetry(ctx context.Context, fn func(fdb.Transaction) (inter
 	for {
 		tr, err := d.db.CreateTransaction()
 		if err != nil {
+			tr.Cancel()
 			return nil, err
 		}
-		defer tr.Cancel()
```
efirs, line 114: What is the issue with defer?

himank: don't we need to cancel the `transaction` if we are not returning in the first attempt? `transaction` is getting created inside for loop but defer will only cancel the last one.

server/metadata/tenant.go
himank: if the version changed after this check then we are good because that means a conflicting transaction has moved forward and this transaction will not see that schema change which is fine.

store/kv/fdb.go
efirs: Ah, right. This was a bug.

efirs: Too many cancels doesn't look nice.
May be we can move the body of the loop the separate function and defer there.

server/config/options.go
```diff
@@ -78,7 +78,7 @@ type TagsConfig struct {
 
 var DefaultConfig = Config{
 	Log: log.LogConfig{
-		Level: "trace",
+		Level: "info",
```
pboros, line 81: Is the default change on purpose or maybe leftover?

efirs (approved):

server/midddleware/auth.go
```diff
@@ -103,7 +103,7 @@ func GetJWTValidator(config *config.Config) *validator.Validator {
 func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config *config.Config) (ctxResult context.Context, err error) {
 	defer func() {
 		if err != nil {
-			log.Warn().Bool("log_only?", config.Auth.LogOnly).Str("error", err.Error()).Err(err).Msg("could not validate token")
+			//log.Warn().Bool("log_only?", config.Auth.LogOnly).Str("error", err.Error()).Err(err).Msg("could not validate token")
```
pboros, line 106: Is this on purpose?

server/midddleware/middleware.go
```diff
@@ -64,7 +64,7 @@ func Get(config *config.Config) (grpc.UnaryServerInterceptor, grpc.StreamServerI
 		grpc_ratelimit.UnaryServerInterceptor(&RateLimiter{}),
 		grpc_auth.UnaryServerInterceptor(authFunction),
 		grpctrace.UnaryServerInterceptor(grpctrace.WithServiceName(config.Tags.Service)),
-		grpc_logging.UnaryServerInterceptor(grpc_zerolog.InterceptorLogger(log.Logger)),
+		//grpc_logging.UnaryServerInterceptor(grpc_zerolog.InterceptorLogger(log.Logger)),
```
pboros, line 67: Should this be put back?

store/kv/noop_store.go
```diff
@@ -0,0 +1,59 @@
...
+type NoopTx struct {
+	*NoopKV
+}
+
+func (n *NoopTx) Commit(context.Context) error   { return nil }
+func (n *NoopTx) Rollback(context.Context) error { return nil }
+func (n *NoopTx) IsRetriable() bool              { return false }
+
+// NoopKVStore is a noop store, useful if we need to profile/debug only compute and not with the storage,
+type NoopKVStore struct {
```
pboros, line 22: This is awesome! Will make future experiments easier.

pboros (approved):

server/midddleware/middleware.go
himank: I think probably we don't want this because this is essentially doing 2 logs per request. We need to figure out how to probably only sample request logs.

server/midddleware/auth.go
himank: unless we hook it up with auth this is always logging on each request so commenting for now.

server/config/options.go
himank: So do we need `trace` by default? I have no concern about changing it back. The clusters must be already setting it to info so this is just for local?

pboros: Right, the clusters are using a production configuration, we kept the default configuration for the local run of a single developer.

store/kv/fdb.go
himank: let me take a stab at it.

server/config/options.go
himank: let me keep it to info for now as we are doing a lot of perf testing, later if needed we can change it to trace again.

server/main.go
```diff
@@ -26,14 +26,15 @@ import (
...
+	"runtime"
 )
 
 func main() {
 	config.LoadConfig(&config.DefaultConfig)
-
 	ulog.Configure(config.DefaultConfig.Log)
 
 	log.Info().Msgf("Environment: %v\n", config.GetEnvironment())
+	log.Info().Msgf("Number of CPUs: %v", runtime.NumCPU())
```
JigarJoshi, line 37: is this docker aware?

JigarJoshi (approved):

server/main.go
himank: yes, it is.

server/midddleware/middleware.go
efirs: We can do this conditionally and enable in the dev/debug environment.

PR #344, opened 2022-07-08
feat: WIP configurable metrics

pboros:
Draft for debugging, do not merge.


server/midddleware/metrics_test.go
```diff
@@ -33,39 +34,40 @@ func TestGrpcMetrics(t *testing.T) {
 	}
 	fullMethodName := "/tigrisdata.v1.Tigris/TestMethod"
 
-	metrics.InitServerRequestMetrics(svcName, methodInfo)
+	if config.DefaultConfig.Metrics.GrpcRequestCounters {
```
efirs, line 37: Don't we want test it unconditionally?

pboros: No, because the initialisation has the condition as well. If the initial structures are not created in tally, the test will panic. I will improve the tests to test the case when it's turned off as well.

efirs: We can just set `config.DefaultConfig.Metrics.GrpcRequestCounters = true` in the beginning of the test?

pboros: It's the same panic. I made the indentation nicer with checking this first.

pboros: Scratch that, it works.

rbarabas (approved):

PR #347, opened 2022-07-11
fix: sync metric naming and tagging with datadog conventions

pboros:
Fix metric naming convention with datadog.


server/metrics/metrics.go
```diff
@@ -38,8 +38,11 @@ var (
 )
 
 func GetGlobalTags() map[string]string {
-	// TODO: add version here
-	return map[string]string{}
+	return map[string]string{
+		"service": config.DefaultConfig.Tags.Service,
+		"env":     config.GetEnvironment(),
```
rbarabas, line 43: We could probably just switch this back to `Tags` to only set it once (in Tags).

server/metrics/requests.go
```diff
@@ -34,17 +34,17 @@ func newRequestEndpointMetadata(serviceName string, methodInfo grpc.MethodInfo)
 
 func (g *RequestEndpointMetadata) GetPreInitializedTags() map[string]string {
 	return map[string]string{
-		"tigris_server_request_method":       g.methodInfo.Name,
-		"tigris_server_request_service_name": g.serviceName,
+		"method": g.methodInfo.Name,
+		//"service": g.serviceName,
```
rbarabas, line 38: Should we add this back?

server/metrics/metrics.go
pboros: I changed it.

rbarabas (approved):

PR #352, opened 2022-07-12
fix: add grpc service name to grpc metrics

pboros:
Load the environment from tags, add grpc_service to grpc request metric tags.

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -40,7 +40,7 @@ var (
 func GetGlobalTags() map[string]string {
 	return map[string]string{
 		"service": config.DefaultConfig.Tags.Service,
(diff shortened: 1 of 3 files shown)


server/metrics/metrics.go
```diff
@@ -40,7 +40,7 @@ var (
 func GetGlobalTags() map[string]string {
 	return map[string]string{
 		"service": config.DefaultConfig.Tags.Service,
-		"env":     config.GetEnvironment(),
+		"env":     config.DefaultConfig.Tags.Environment,
```
rbarabas, line 43: I'm actually retracting my stance here with https://github.com/tigrisdata/tigris/pull/349/files.

rbarabas (approved):

PR #353, opened 2022-07-12
feat: namespace tag in grpc request metrics

pboros:
Add namespace as part of the tags. When the metrics are initialised at server startup, those are using `default_namespace`.

Example:
```
server_requests_ok{env="",grpc_service="tigrisdata.v1.Tigris",method="ListDatabases",namespace="default_namespace",service="tigris-server",version="065426c"} 15
```

### DIFF
--- a/server/metrics/requests.go
+++ b/server/metrics/requests.go
@@ -36,6 +36,7 @@ func (g *RequestEndpointMetadata) GetPreInitializedTags() map[string]string {
 	return map[string]string{
 		"method":       g.methodInfo.Name,
 		"grpc_service": g.serviceName,
+		"namespace":    "default_namespace",
 	}
 }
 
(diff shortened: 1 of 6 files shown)


rbarabas (approved):

server/muxer/muxer.go
```diff
@@ -16,13 +16,13 @@ package muxer
 
 import (
 	"fmt"
+	"github.com/tigrisdata/tigris/server/metrics"
```
himank, line 19: goimports.

server/metrics/requests.go
```diff
@@ -36,6 +36,7 @@ func (g *RequestEndpointMetadata) GetPreInitializedTags() map[string]string {
 	return map[string]string{
 		"method":       g.methodInfo.Name,
 		"grpc_service": g.serviceName,
+		"namespace":    "default_namespace",
```
himank, line 39: nit: not for this diff, but probably we can define this constant somewhere(may be in types.go), currently it is defined in the metadata package.

PR #355, opened 2022-07-12
fix: disable log

himank:


server/midddleware/middleware.go
```diff
@@ -101,7 +101,7 @@ func Get(config *config.Config, tenantMgr *metadata.TenantManager, txMgr *transa
 		grpc_auth.UnaryServerInterceptor(authFunction),
 		namespaceInitializer.NamespaceSetterUnaryServerInterceptor(),
 		grpctrace.UnaryServerInterceptor(grpctrace.WithServiceName(util.Service)),
-		grpc_logging.UnaryServerInterceptor(grpc_zerolog.InterceptorLogger(taggedLogger)),
+		//grpc_logging.UnaryServerInterceptor(grpc_zerolog.InterceptorLogger(taggedLogger)),
```
rbarabas, line 104: Hm, we still kinda need this to troubleshoot log tracing on the deployment side (wether the right flags are set etc). Should we make this configurable with a switch in the interim? 

Btw the plan is to make the sample rate configurable tomorrow, along with it, this logging.

himank: This diff is mainly because I am doing the benchmarking, feel free to take your time to make it configurable. I am not landing it.

PR #358, opened 2022-07-12
feat: Implemented ListNamespace admin API

JigarJoshi:


server/services/v1/admin.go
```diff
@@ -74,6 +74,30 @@ func (a *adminService) CreateNamespace(ctx context.Context, req *api.CreateNames
 	}
 }
 
+func (a *adminService) ListNamespaces(ctx context.Context, _ *api.ListNamespacesRequest) (*api.ListNamespacesResponse, error) {
+	namespaces, err := a.tenantMgr.ListNamespaces(ctx, a.txMgr)
+	if err != nil {
+		return nil, err
+	}
+	if namespaces != nil {
```
himank, line 82: avoid else by simply short-circuiting to `if namespaces == nil {return ...}`

server/metadata/tenant.go
```diff
@@ -201,12 +201,28 @@ func (m *TenantManager) GetNamespace(ctx context.Context, namespaceName string,
...
+	}
+	for i := range namespaces {
+		if namespaces[i].Name() == namespaceName {
+			return namespaces[i], nil
+		}
+	}
+	return nil, api.Errorf(api.Code_NOT_FOUND, "Namespace not found")
+}
+
+func (m *TenantManager) ListNamespaces(ctx context.Context, txMgr *transaction.Manager) ([]Namespace, error) {
```
himank, line 223: as you're controlling the transaction for the create namespace in the service layer, should we do the same for ListNamespace? The GetNamespace will create the transaction and pass it to ListNamespace.

server/metadata/tenant.go
```diff
@@ -201,12 +201,28 @@ func (m *TenantManager) GetNamespace(ctx context.Context, namespaceName string,
 	m.RUnlock()
 
 	m.Lock()
-	defer m.Unlock()
 	if tenant, found := m.tenants[namespaceName]; found {
 		return tenant.namespace, nil
 	}
-	// this will never create new namespace
-	// this is for reading namespaces from storage into cache
+	m.Unlock()
```
himank, line 207: this won't work, you need to release the lock after populating the map.

server/services/v1/admin.go
JigarJoshi: trimmed.

server/metadata/tenant.go
JigarJoshi: made it consistent.

server/metadata/tenant.go
JigarJoshi: moved after it is done reading. Please review this one more time.

server/metadata/tenant.go
```diff
@@ -191,41 +191,82 @@ func (m *TenantManager) CreateTenant(ctx context.Context, tx transaction.Tx, nam
 	}
 	return namespace, nil
 }
-
-func (m *TenantManager) GetNamespace(ctx context.Context, namespaceName string, txMgr *transaction.Manager) (Namespace, error) {
-	m.RLock()
+func (m *TenantManager) getTenantFromCache(namespaceName string) (tenant *Tenant) {
 	if tenant, found := m.tenants[namespaceName]; found {
```
himank, line 195: take read lock

JigarJoshi: fixed. thanks!

PR #359, opened 2022-07-12
fix: Squashed minor search api bugs

adilansari:


api/server/v1/marshaler.go
```diff
@@ -441,6 +441,36 @@ func (x *SearchHit) MarshalJSON() ([]byte, error) {
 	return json.Marshal(resp)
 }
 
+func (x *SearchMetadata) MarshalJSON() ([]byte, error) {
+	resp := struct {
+		Found      int64 `json:"found"`
```
adilansari, line 446: Found could be 0, hence, not omitting

server/services/v1/query_runner.go
```diff
@@ -537,6 +537,7 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 	}
 
 	facets, err := qsearch.UnmarshalFacet(runner.req.Facet)
+	// test if facet field is actually a schema field
```
adilansari, line 540: Will include this in another iteration, wanted to get server in sync with latest API

server/services/v1/search_reader.go
```diff
@@ -18,16 +18,17 @@ import (
 	"context"
 
 	jsoniter "github.com/json-iterator/go"
+	tsApi "github.com/typesense/typesense-go/typesense/api"
+
```
himank, line 22: remove the line and run `go imports`.

store/search/ts.go
```diff
@@ -16,15 +16,19 @@ package search
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 
+	"github.com/rs/zerolog/log"
+
```
himank, line 25: ^

test/docker/docker-compose.yml
```diff
@@ -86,7 +86,6 @@ services:
       # instead of test/config/fdb.cluster file when running on the host
       - TIGRIS_ENVIRONMENT=test
       - TEST_PARAM
-    container_name: tigris_test
```
himank, line 89: is this intentional?

adilansari: Already set above

server/services/v1/query_runner.go
```diff
@@ -591,17 +580,31 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
...
 			})
-			totalInPage++
 
 			if len(resp.Hits) == pageSize {
 				break
 			}
 		}
-		if len(resp.Hits) == 0 {
+		if rowReader.err != nil {
+			return nil, ctx, rowReader.err
```
himank, line 589: We may have buffered the events in the `Hits` if we return an error from here then we won't be able to send those hits?

server/services/v1/query_runner.go
```diff
@@ -591,17 +580,31 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
...
+			Found:      rowReader.getTotalFound(),
+			TotalPages: totalPages,
+			Page: &api.Page{
+				Current: pageNo,
+				Size:    int32(searchQ.PageSize),
+			},
+		}
+
+		// break the stream after sending at least 1 response
+		if len(resp.Hits) == 0 && pageNo != int32(defaultPageNo) {
```
himank, line 604: Shouldn't the condition here be OR and also what if the user passed the pageNo as `1` then also we need to stop?

server/services/v1/search_reader.go
adilansari: apparently this is after goimports, it is grouping imports by package. Is there a particular configuration that we follow for goimports?

himank: So we remove the extra lines and then run `go imports`. This way there are only two groups of imports.

PR #364, opened 2022-07-13
feat: Admin authz and unknown namespace

JigarJoshi:
- Allow to perform admin operations for configured admin namespaces users. (This will later on extended to finer grain role based access control system)
- For administration operations set the namespace to "unknown"


server/config/options.go
```diff
@@ -102,10 +104,11 @@ var DefaultConfig = Config{
 	},
 	Auth: AuthConfig{
 		IssuerURL:                "https://tigrisdata-dev.us.auth0.com/",
-		Audience:                 "https://tigris-db-dev-preview-test",
+		Audience:                 "https://tigris-db-api",
 		JWKSCacheTimeout:         5 * time.Minute,
 		LogOnly:                  true,
 		EnableNamespaceIsolation: false,
+		AdminNamespaces:          set.New("tigris-data"),
```
ovaistariq, line 111: We should have a generic name for admin namespace?

server/config/options.go
```diff
@@ -102,10 +104,11 @@ var DefaultConfig = Config{
 	},
 	Auth: AuthConfig{
 		IssuerURL:                "https://tigrisdata-dev.us.auth0.com/",
-		Audience:                 "https://tigris-db-dev-preview-test",
+		Audience:                 "https://tigris-db-api",
```
ovaistariq, line 107: Can we change it to tigris-api

server/midddleware/auth.go
```diff
@@ -100,7 +101,16 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
 			}
 		}
 	}()
-
+	// disable health check authn/z
+	var fullMethodName string
```
himank, line 105: move this in a helper method so that you can simply call it here whether to run auth check or not.

server/midddleware/auth.go
```diff
@@ -100,7 +101,16 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
...
+	fullMethodNameFound := false
+	clientCtx := inprocgrpc.ClientContext(ctx)
+	if clientCtx != nil {
+		fullMethodName, fullMethodNameFound = runtime.RPCMethod(clientCtx)
+	}
+	if fullMethodNameFound && fullMethodName == "/.HealthAPI/Health" {
```
himank, line 111: Should we have a list for this? Or const string defined above if it is just one.

server/midddleware/namespace.go
```diff
@@ -20,15 +20,15 @@ type NamespaceSetter struct {
...
+		if !r.config.Auth.EnableNamespaceIsolation {
 			return handler(setNamespace(ctx, metadata.DefaultNamespaceName), req)
 		} else {
 			namespace, err := r.namespaceExtractor.Extract(ctx)
 			if err != nil {
 				return nil, err
 			}
 			if namespace == "" {
-				return nil, api.Errorf(api.Code_INVALID_ARGUMENT, "Could not find namespace")
+				return handler(setNamespace(ctx, "unknown"), req)
```
himank, line 31: define `unknown` const above

server/midddleware/auth.go
```diff
@@ -113,10 +123,26 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
...
+			}
 			log.Debug().Msg("Valid token received")
-			token := &AccessToken{
-				Namespace: customClaims.Namespace.Code,
-				Sub:       validatedClaims.RegisteredClaims.Subject,
+			var token *AccessToken
+			if isAdmin {
+				token = &AccessToken{
+					// set the unknown namespace for administration
+					Namespace: "unknown",
```
himank, line 138: Why do we need to do this? There is already a check above that means there is an admin namespace passed.

server/midddleware/auth.go
```diff
@@ -100,7 +101,16 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
 			}
 		}
 	}()
-
+	// disable health check authn/z
+	var fullMethodName string
+	fullMethodNameFound := false
+	clientCtx := inprocgrpc.ClientContext(ctx)
+	if clientCtx != nil {
+		fullMethodName, fullMethodNameFound = runtime.RPCMethod(clientCtx)
```
himank, line 109: also shouldn't this live in request_metadata? `GetMethodFullName(ctx) (string, bool)`

server/config/options.go
JigarJoshi: renamed.

server/config/options.go
JigarJoshi: updated.

server/midddleware/auth.go
JigarJoshi: refactored. thanks

server/midddleware/auth.go
JigarJoshi: sounds good. fixed.

server/midddleware/namespace.go
JigarJoshi: done

server/midddleware/auth.go
JigarJoshi: discussed and updated to only set `unknown` in case if incoming namespace is `""` with legit token.

server/midddleware/auth.go
```diff
@@ -124,3 +145,11 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
 	// this should never happen.
 	return ctx, api.Errorf(api.Code_UNAUTHENTICATED, "You are not authorized to perform this action")
 }
+
+func getFullMethodName(ctx context.Context) (string, bool) {
```
himank, line 149: not for this diff, but move this method in request_metadata.

himank (approved):

server/midddleware/auth.go
JigarJoshi: yes you are right this is more suitable there. moved it to request_metadata.

server/midddleware/auth.go
JigarJoshi: yes you are right this is more suitable there. moved it to request_metadata.

himank (approved):

PR #374, opened 2022-07-20
feat: first iteration on datadog tracing with spans

pboros:


server/transaction/manager.go
```diff
@@ -130,10 +145,45 @@ func newTxSession(kv kv.KeyValueStore) (*TxSession, error) {
...
+	}
+	return &TxSessionWithMetrics{&TxSession{
+		context: &SessionCtx{},
+		kvStore: kv,
+		state:   sessionCreated,
+		txCtx:   generateTransactionCtx(),
+	}}, nil
+}
+
+func (m *TxSessionWithMetrics) measure(ctx context.Context, name string, f func(ctx context.Context) error) {
```
himank, line 160: this will be overkill here, once you move it to sessions.go then let's remove it from here.

store/kv/kv.go
```diff
@@ -102,6 +103,22 @@ func measureLow(ctx context.Context, name string, f func() error) {
...
 		defer metrics.FdbRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets).Start().Stop()
 	}
+	if config.DefaultConfig.DatadogTrace.Enabled {
+		spanOpts := []tracer.StartSpanOption{
+			tracer.ServiceName("tigris.fdb.kv"),
+			tracer.ResourceName(name),
+			tracer.SpanType("fdb_kv"),
+			tracer.Measured(),
+		}
+		parentSpan, exists := tracer.SpanFromContext(ctx)
```
himank, line 113: move the spanOpts initialization inside the `if exists` check.

store/kv/kv.go
```diff
@@ -102,6 +103,22 @@ func measureLow(ctx context.Context, name string, f func() error) {
 		metrics.FdbRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets)
 		defer metrics.FdbRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets).Start().Stop()
 	}
+	if config.DefaultConfig.DatadogTrace.Enabled {
+		spanOpts := []tracer.StartSpanOption{
```
himank, line 107: Should we move this in a helper method? `GetSpanOptions() []tracer.StartSpanOption`

store/search/ts.go
```diff
@@ -38,72 +40,88 @@ type storeImplWithMetrics struct {
...
 }
 
+func (m *storeImplWithMetrics) measure(ctx context.Context, name string, f func(ctx context.Context) error) {
+	// Low level measurement wrapper that is called by the measure functions on the appropriate receiver
+	tags := metrics.GetSearchTags(ctx, name)
+	if config.DefaultConfig.Metrics.Search.ResponseTime {
+		metrics.SearchRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets)
+		defer metrics.SearchRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets).Start().Stop()
+	}
+	if config.DefaultConfig.DatadogTrace.Enabled {
```
himank, line 50: is it possible to move this tracing logic in a single method? We can have a `tracing` package that can have a function startTracing?

```
func startTracing(ctx context.Context, serviceName string, resourceName string, spanType string) (context.Context, func()){
	if config.DefaultConfig.DatadogTrace.Enabled {
		parentSpan, exists := tracer.SpanFromContext(ctx)
		if exists {
			spanOpts := []tracer.StartSpanOption{
				tracer.ServiceName(serviceName),
				tracer.ResourceName(resourceName),
				tracer.SpanType(spanType),
				tracer.Measured(),
			}
			
			// There is a parent span, keep on tracing in the kv layer
			spanOpts = append(spanOpts, tracer.ChildOf(parentSpan.Context()))
			span := tracer.StartSpan(resourceName, spanOpts...)
			defer span.Finish()
			ctx = tracer.ContextWithSpan(ctx, span)
			return ctx, func() {
				span.Finish()
			}
		}
	}

	return ctx, func(){}
}
```

and then the caller can simply do,
```
ctx, cb := StartTracing(...)
defer cb()
```

server/main.go
```diff
@@ -62,14 +62,22 @@ func main() {
...
+	}
 	if err != nil {
 		log.Fatal().Err(err).Msg("error initializing search store")
 	}
 
 	tenantMgr := metadata.NewTenantManager(kvStore)
 	log.Info().Msg("initialized tenant manager")
+
 	txMgr := transaction.NewManager(kvStore)
+
```
rbarabas, line 80: Should here be a space?

server/main.go
```diff
@@ -62,14 +62,22 @@ func main() {
...
+	} else {
+		searchStore, err = search.NewStore(&config.DefaultConfig.Search)
+	}
 	if err != nil {
 		log.Fatal().Err(err).Msg("error initializing search store")
 	}
 
 	tenantMgr := metadata.NewTenantManager(kvStore)
 	log.Info().Msg("initialized tenant manager")
+
```
rbarabas, line 78: If we have a space here we likely will want it after L81 also.

server/config/options.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 	Tracing      TracingConfig   `yaml:"tracing" json:"tracing"`
 	Profiling    ProfilingConfig `yaml:"profiling" json:"profiling"`
 	Metrics      MetricsConfig
+	DatadogTrace DatadogTraceConfig
```
rbarabas, line 38: Should we just add this under Tracing?

server/config/options.go
```diff
@@ -82,6 +83,11 @@ type MetricsConfig struct {
 	Search SearchMetricsConfig
 }
 
+type DatadogTraceConfig struct {
+	// Global switch
+	Enabled bool `mapstructure:"enabled" yaml:"enabled" json:"enabled"`
```
rbarabas, line 88: Do we expect other variables here? If not, consider adding it as an item under Tracing.

server/midddleware/middleware.go
```diff
@@ -71,9 +70,10 @@ func Get(config *config.Config, tenantMgr *metadata.TenantManager, txMgr *transa
 		grpc_ratelimit.StreamServerInterceptor(&RateLimiter{}),
 		grpc_auth.StreamServerInterceptor(authFunction),
 		namespaceInitializer.NamespaceSetterStreamServerInterceptor(),
-		grpctrace.StreamServerInterceptor(grpctrace.WithServiceName(util.Service)),
-		grpc_logging.StreamServerInterceptor(grpc_zerolog.InterceptorLogger(sampledTaggedLogger), []grpc_logging.Option{}...),
-		validatorStreamServerInterceptor(),
+	}
+
+	if config.DatadogTrace.Enabled {
```
rbarabas, line 75: Should we put this logic into the Interceptor and just pass out required components (Logger etc. Config is global so no need to pass that down)? Am I missing the benefit of having these interceptors exposed at this level?

server/midddleware/middleware.go
```diff
@@ -71,9 +70,10 @@ func Get(config *config.Config, tenantMgr *metadata.TenantManager, txMgr *transa
...
+		streamInterceptors = append(streamInterceptors, ddTraceStream())
 	}
 
 	if config.Metrics.Grpc.Enabled && config.Metrics.Grpc.ResponseTime {
```
rbarabas, line 79: What's our goal with the "basic" GRPC interceptor? Do we still need it? Or is this in place to make the transition smoother?

server/midddleware/middleware.go
```diff
@@ -103,10 +105,10 @@ func Get(config *config.Config, tenantMgr *metadata.TenantManager, txMgr *transa
...
-		grpc_logging.UnaryServerInterceptor(grpc_zerolog.InterceptorLogger(sampledTaggedLogger)),
-		validatorUnaryServerInterceptor(),
-		timeoutUnaryServerInterceptor(DefaultTimeout),
+	}
+
+	if config.DatadogTrace.Enabled {
+		unaryInterceptors = append(unaryInterceptors, ddTraceUnary())
 	}
 
 	if config.Metrics.Grpc.Enabled && config.Metrics.Grpc.ResponseTime {
```
rbarabas, line 114: I wonder if we should put these Enabled flags under a single config structure and just case switch on them. WDYT?

server/midddleware/middleware.go
```diff
@@ -118,6 +120,10 @@ func Get(config *config.Config, tenantMgr *metadata.TenantManager, txMgr *transa
 	}
 
 	unaryInterceptors = append(unaryInterceptors, []grpc.UnaryServerInterceptor{
+		//grpctrace.UnaryServerInterceptor(grpctrace.WithServiceName(util.Service)),
+		grpc_logging.UnaryServerInterceptor(grpc_zerolog.InterceptorLogger(sampledTaggedLogger)),
```
rbarabas, line 124: Shall we add a TODO to hook this up to the span IDs? Perhaps perform the logging in the code that manages the high level request?

server/config/options.go
himank: +1 better to just use Tracing if we already have otherwise it may be confusing.

server/midddleware/ddtrace.go
```diff
@@ -0,0 +1,97 @@
...
+	middleware "github.com/grpc-ecosystem/go-grpc-middleware"
+	"github.com/tigrisdata/tigris/server/metrics"
+	"github.com/tigrisdata/tigris/server/request"
+	ulog "github.com/tigrisdata/tigris/util/log"
+	"google.golang.org/grpc"
+	"gopkg.in/DataDog/dd-trace-go.v1/ddtrace/tracer"
+)
+
+func getMethodName(fullMethod string) string {
+	return strings.Split(fullMethod, "/")[2]
```
rbarabas, line 30: This might be prone to formatting changes and likely more Tigris than tracing specific at the moment. Should we handle this a level above in "Tigris context"?

rbarabas: I guess the whole file is very tigris specific. Based on the name I expected datadog-specific tracing here. Should we just rename the file to tracing.go?

server/midddleware/ddtrace.go
```diff
@@ -0,0 +1,97 @@
...
+		ctx = tracer.ContextWithSpan(ctx, span)
+		for k, v := range getTags(ctx, info.FullMethod, "unary") {
+			span.SetTag(k, v)
+		}
+		resp, err := handler(ctx, req)
+		return resp, err
+	}
+}
+
+func ddTraceStream() grpc.StreamServerInterceptor {
```
rbarabas, line 77: Both this method and the GRPC are technically datadog tracers. Consider simply calling these `traceStream()`, `traceUnary` etc.

server/midddleware/metrics.go
```diff
@@ -19,13 +19,12 @@ import (
...
 	"strconv"
 
-	"github.com/tigrisdata/tigris/server/request"
-	ulog "github.com/tigrisdata/tigris/util/log"
-	"github.com/uber-go/tally"
-
 	"github.com/apple/foundationdb/bindings/go/src/fdb"
 	api "github.com/tigrisdata/tigris/api/server/v1"
 	"github.com/tigrisdata/tigris/server/metrics"
+	"github.com/tigrisdata/tigris/server/request"
```
rbarabas, line 25: Nice, whitespace commando at its best!!! :- )

store/kv/kv.go
rbarabas: +1! (store/kv/tracing.go?) 

Also, I'd at least define const for these presets in said file. Configurability probably not required.

store/kv/kv.go
```diff
@@ -102,6 +103,22 @@ func measureLow(ctx context.Context, name string, f func() error) {
...
+			tracer.ServiceName("tigris.fdb.kv"),
+			tracer.ResourceName(name),
```
rbarabas, line 109: I would actually move the resource name mangling logic into that helper as well, if applicable, from `server/"middddddddddddleware"/ddtrace.go`. :- )

store/search/ts.go
rbarabas: +1 to this as well. It could be a bit distracting to go through this when we aim to debug the behavior of measure. If we encapsulate the tracing logic it will be still visible but the focus in tracing will be... on tracing so it would feel like a more natural scope.

store/search/ts.go
```diff
@@ -38,72 +40,88 @@ type storeImplWithMetrics struct {
...
+		parentSpan, exists := tracer.SpanFromContext(ctx)
+		if exists {
+			spanOpts := []tracer.StartSpanOption{
+				tracer.ServiceName("tigris.search"),
```
rbarabas, line 54: Consider using const for these and encapsulating the defintions in a package.

store/search/ts.go
```diff
@@ -38,72 +40,88 @@ type storeImplWithMetrics struct {
...
+			}
+			// There is a parent span, keep on tracing in the kv layer
+			spanOpts = append(spanOpts, tracer.ChildOf(parentSpan.Context()))
+			span := tracer.StartSpan(name, spanOpts...)
+			defer span.Finish()
+			ctx = tracer.ContextWithSpan(ctx, span)
+		}
+	}
+	err := f(ctx)
+	if err == nil {
```
rbarabas, line 67: We have a nice way to do the log tracing here instead of higher above as the Span ID is exposed. I'd just add a // TODO for now to remind me later and follow up with a separate PR to address that.

server/config/options.go
pboros: I can just use tracing enabled for now.

server/config/options.go
pboros: For now I will just use tracing enabled, when we will have multiple tracers, we will think about how to configure them.

server/midddleware/ddtrace.go
pboros: I refactored this so it uses the same things as the counters, so all the tagging logic is in one place now.

server/midddleware/ddtrace.go
pboros: Makes sense, I renamed these.

server/midddleware/metrics.go
pboros: This was goland I think :).

server/midddleware/middleware.go
pboros: Initializing the interceptor has a very visible cost according to the profiler. I think it's better only to initialize them when they are needed. What do you think?

server/midddleware/middleware.go
pboros: This is not something I changed, I just moved it in the initialization order. This and logs attached to spans are different.

server/transaction/manager.go
pboros: Will do.

store/kv/kv.go
pboros: I did that, not in kv package, but in metrics (same place as the tagging logic).

store/kv/kv.go
pboros: Done.

store/kv/kv.go
pboros: Good catch, moved.

store/kv/kv.go
```diff
@@ -102,6 +103,17 @@ func measureLow(ctx context.Context, name string, f func() error) {
 		metrics.FdbRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets)
 		defer metrics.FdbRequests.Tagged(tags).Histogram("histogram", tally.DefaultBuckets).Start().Stop()
 	}
+	if config.DefaultConfig.Tracing.Enabled {
```
rbarabas, line 106: Would the following work?

```
err := startTracing(ctx, name)
```

Above assumes no defer needs to be passed up but if you think you need the span later after L117 we can bring it back for Finish() or other things.

The config will be available as it's global so no need to pass that. 

```
func startTracing(ctx context.Context, name string) error {
    if !config.DefaultConfig.Tracing.Enabled {
        return nil
    }
    
    parentSpan, exists := tracer.SpenFromContext(ctx)
    if exists {
        ...
    }
    
    return nil
}
```

store/search/ts.go
pboros: Thanks, this is great. I refactored it this way.

store/search/ts.go
pboros: Done.

store/search/ts.go
pboros: I will add the log related things in the next PR.

himank (approved):

PR #375, opened 2022-07-20
fix: Hits and Facets cannot be null

adilansari:
Includes fixes for:
- null `Hits` and `Facets` to be serialized as empty JSON objects instead of null
- totalPages were having an extra page being computed in some cases (when found elements was either `0` or same as `pageSize`)

### DIFF
--- a/server/services/v1/query_runner.go
+++ b/server/services/v1/query_runner.go
@@ -17,6 +17,7 @@ package v1
 import (
 	"bytes"
 	"context"
+	"math"
 
 	jsoniter "github.com/json-iterator/go"
 	api "github.com/tigrisdata/tigris/api/server/v1"
@@ -539,6 +540,7 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 	if pageSize == 0 {
 		pageSize = defaultPerPage
 	}
+	var totalPages *int32
 
 	searchQ := qsearch.NewBuilder().
 		Query(runner.req.Q).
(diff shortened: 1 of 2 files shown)


server/services/v1/query_runner.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/tigrisdata/tigris/store/kv"
 	"github.com/tigrisdata/tigris/store/search"
 	ulog "github.com/tigrisdata/tigris/util/log"
+	"math"
```
himank, line 36: goimports

server/services/v1/query_runner.go
```diff
@@ -580,11 +583,15 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 		}
 
 		resp.Facets = rowReader.getFacets()
-		ps := int64(pageSize)
-		totalPages := int32(rowReader.getTotalFound()/ps) + 1
+		if totalPages == nil {
+			ps := float64(pageSize)
+			tp := int32(math.Ceil(float64(rowReader.getTotalFound()) / ps))
```
himank, line 588: just combine ps with this line `tp := int32(math.Ceil(float64(rowReader.getTotalFound()) / float64(pageSize)))`

server/services/v1/query_runner.go
```diff
@@ -18,6 +18,8 @@ import (
 	"bytes"
 	"context"
 
+	"math"
+
```
himank, line 22: remove extra line.

PR #377, opened 2022-07-21
feat: field selection for search

adilansari:
- Ability to request certain document fields from search results, by inclusion or exclusion:
```
{
    "q": "running",
    "fields": {"title": false, "price": false}
}
```
- Fixed a minor bug where facet sizes was not being respected in response. Only the max size was being honored

### DIFF
--- a/server/services/v1/search_reader.go
+++ b/server/services/v1/search_reader.go
@@ -183,6 +183,11 @@ func (p *pageReader) buildFacets(facets *[]tsApi.FacetCounts) {
 	if facets == nil {
 		return
 	}
+	// count of values to include in response
+	facetSizeRequested := map[string]int{}
+	for _, f := range p.query.Facets.Fields {
+		facetSizeRequested[f.Name] = f.Size
+	}
 
 	for _, f := range *facets {
 		var facet = &api.SearchFacet{
@@ -190,7 +195,10 @@ func (p *pageReader) buildFacets(facets *[]tsApi.FacetCounts) {
 		}
 
 		if f.Counts != nil {
-			for _, c := range *f.Counts {
+			for i, c := range *f.Counts {
+				if i >= facetSizeRequested[*f.FieldName] {
+					break
+				}
 				facet.Counts = append(facet.Counts, &api.FacetCount{
 					Count: int64(*c.Count),
 					Value: *c.Value,


JigarJoshi:
do we need true/false explicitly? how about this

```
    "fields": ["title", "price"]
```

and for advanced support

```
    "excluded_fields": ["author", "language"]
```

what do you think?

JigarJoshi:
please add test exercising this code path.

query/search/search.go
```diff
@@ -16,18 +16,20 @@ package search
...
-	Fields   []string
-	Facets   Facets
-	PageSize int
-	WrappedF *filter.WrappedFilter
+	Q              string
+	SearchFields   []string
+	Facets         Facets
+	PageSize       int
+	WrappedF       *filter.WrappedFilter
+	FieldSelection *read.FieldFactory
```
himank, line 32: just have a simple list, no need to have it as a map because I don't see in the future if we need to add any functionality inside the fields object

server/services/v1/search_reader.go
```diff
@@ -89,11 +93,22 @@ func (p *page) readRow(row *Row) bool {
...
+		// unmarshal the row data
+		var rawData []byte
+		if rawData, p.err = jsoniter.Marshal(doc); p.err != nil {
 			return false
 		}
-
+		// apply the field selection
+		if p.fieldSelection != nil {
+			newValue, err := p.fieldSelection.Apply(rawData)
+			if ulog.E(err) {
```
himank, line 104: What about user input is a list but in code, we can build the fields using `read.FieldFactory` then we can still use this method but user input is simple.

query/read/fields.go
```diff
@@ -37,6 +38,37 @@ func BuildFields(reqFields jsoniter.RawMessage) (*FieldFactory, error) {
 	factory.Include = make(map[string]Field)
 	factory.Exclude = make(map[string]Field)
 
+	var v interface{}
```
adilansari, line 41: allowing read fields input to be an array, will update user facing documentation for both APIs separately

adilansari:
> do we need true/false explicitly? how about this
> 
> ```
>     "fields": ["title", "price"]
> ```
> 
> and for the other scenario
> 
> ```
>     "excluded_fields": ["author", "language"]
> ```
> 
> what do you think?

So should we make an API change now for both read and search to accept, `include`/`exclude` or just implicitly assume `fields` array to be inclusion list

server/services/v1/search_reader.go
adilansari: I did make the change but doesn't look too intuitive to implicitly assume `fields` array to inclusion list. It may not appear intuitive if a user is building `[a, b]` for inclusion list and `[a: false, b: false]` for exclusion. Even if this array is easy to use, we should rather have `include` and `exclude` explicitly

himank:
> > do we need true/false explicitly? how about this
> > ```
> >     "fields": ["title", "price"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and for the other scenario
> > ```
> >     "excluded_fields": ["author", "language"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > what do you think?
> 
> So should we make an API change now for both read and search to accept, `include`/`exclude` or just implicitly assume `fields` array to be inclusion list

So we don't need to make any changes to read because read supports `expressions` as a value which the current implementation only supports `true/false/1/0` but we will be extending it later to support arithmetic operations etc. 

For search, the value is not an expression. The only value for it would be `true/false/1/0` or separating it out as an `include/exclude` list. 

Thinking more about it, I think either approach should be fine. to support it as an array with explicit exclude/include or to have a map.

server/services/v1/search_reader.go
himank: No, I was not saying it to have `[a: false, b: false]`. I was saying to just have an array `[a, b]`, similarly for exclude it will be an array.

query/read/fields.go
```diff
@@ -37,6 +38,37 @@ func BuildFields(reqFields jsoniter.RawMessage) (*FieldFactory, error) {
...
+			})
+		}
+	case map[string]interface{}:
+		// it's an object like {"a": 1, "b": 0}
+		err = unmarshalObject(reqFields, factory)
+	default:
+		// it's something else
+		err = api.Errorf(api.Code_INVALID_ARGUMENT, "fields selection can only be array or object")
+	}
+
```
himank, line 63: nit, remove the extra line.

query/read/fields.go
```diff
@@ -37,6 +38,37 @@ func BuildFields(reqFields jsoniter.RawMessage) (*FieldFactory, error) {
 	factory.Include = make(map[string]Field)
 	factory.Exclude = make(map[string]Field)
 
+	var v interface{}
+	if err := json.Unmarshal(reqFields, &v); err != nil {
+		return nil, err
+	}
+
+	var err error
+	switch v := v.(type) {
```
himank, line 47: What will we have in the open API spec? Object or array?

himank: if it is confusing, I am fine to stick with having it as an object to avoid any confusion and if needed in clients we can have an array.

adilansari:
This PR now only has fix for facet sizes, I'll send a separate one for API change

himank (approved):

PR #379, opened 2022-07-22
fix: allow enabling fdb delete using config, fix a metadata version bug

himank:
By default, drop collection will be only metadata change. Setting `fdb_delete` to true will be performing fdb delete(clearrange call).

### DIFF
--- a/server/config/options.go
+++ b/server/config/options.go
@@ -21,8 +21,9 @@ import (
 )
 
 type ServerConfig struct {
-	Host string
-	Port int16
+	Host      string
+	Port      int16
+	FDBDelete bool `mapstructure:"fdb_delete" yaml:"fdb_delete" json:"fdb_delete"`
 }
 
 type Config struct {
@@ -105,8 +106,9 @@ var DefaultConfig = Config{
 		SampleRate: 0.01,
 	},
 	Server: ServerConfig{
-		Host: "0.0.0.0",
-		Port: 8081,
+		Host:      "0.0.0.0",
+		Port:      8081,
+		FDBDelete: false,
 	},
 	Auth: AuthConfig{
 		IssuerURL:                "https://tigrisdata-dev.us.auth0.com/",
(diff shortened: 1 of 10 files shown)


server/config/options.go
```diff
@@ -105,8 +106,9 @@ var DefaultConfig = Config{
 		SampleRate: 0.01,
 	},
 	Server: ServerConfig{
-		Host: "0.0.0.0",
-		Port: 8081,
+		Host:      "0.0.0.0",
+		Port:      8081,
+		FDBDelete: false,
```
efirs, line 111: This is not needed. It's false by default.

server/midddleware/validate.go
```diff
@@ -39,7 +38,7 @@ func validatorUnaryServerInterceptor() grpc.UnaryServerInterceptor {
 // validatorStreamServerInterceptor returns a new streaming server interceptor that validates incoming messages.
 func validatorStreamServerInterceptor() grpc.StreamServerInterceptor {
 	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
-		if tx := api.GetTransaction(stream.Context(), nil); api.IsTxSupported(stream.Context()) && tx != nil {
+		if tx := api.GetTransaction(stream.Context()); api.IsTxSupported(stream.Context()) && tx != nil {
```
efirs, line 41: Same as above `!api.IsTxSupported`

server/config/options.go
himank: I intentionally added it to make sure behavior is clear.

server/midddleware/validate.go
himank: sure

efirs (approved):

server/config/options.go
```diff
@@ -21,8 +21,9 @@ import (
 )
 
 type ServerConfig struct {
-	Host string
-	Port int16
+	Host      string
+	Port      int16
+	FDBDelete bool `mapstructure:"fdb_delete" yaml:"fdb_delete" json:"fdb_delete"`
```
ovaistariq, line 26: This config param name can be confusing because delete also means deletes done in response to the DELETE query. How about something like `FDBHardDrop`, or something else which makes it clear that this is applicable to Drops only.

himank: Sure will update.

PR #381, opened 2022-07-25
feat: read fields support for search

adilansari:
- Implementation for `include_fields` and `exclude_fields` on search api

### DIFF
--- a/api/server/v1/validator.go
+++ b/api/server/v1/validator.go
@@ -103,6 +103,10 @@ func (x *SearchRequest) Validate() error {
 		return err
 	}
 
+	if len(x.IncludeFields) > 0 && len(x.ExcludeFields) > 0 {
+		return Errorf(Code_INVALID_ARGUMENT, "Cannot use both `include_fields` and `exclude_fields` together")
+	}
+
 	if err := isValidPaginationParam("page", int(x.Page)); err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


server/services/v1/query_runner.go
```diff
@@ -675,6 +680,46 @@ func (runner *SearchQueryRunner) getFacetFields(collFields []*schema.Field) (qse
...
+				found = true
+			}
+		}
+		if !found {
+			return nil, api.Errorf(api.Code_INVALID_ARGUMENT, "`%s` is not a schema field", sf)
+		}
+
+		factory.AddField(&read.SimpleField{
+			Name: sf,
+			Incl: include,
```
efirs, line 716: We can just set `Incl: len(runner.req.IncludeFields) > 0` and remove `include` variable.

api/server/v1/validator.go
```diff
@@ -103,6 +103,10 @@ func (x *SearchRequest) Validate() error {
 		return err
 	}
 
+	if len(x.IncludeFields) > 0 && len(x.ExcludeFields) > 0 {
+		return Errorf(Code_INVALID_ARGUMENT, "Cannot use both `include_fields` and `exclude_fields` together")
+	}
```
efirs, line 108: What is the behavior when none of them set?

adilansari: No field projection to apply, this particular code path will be a no-op and all document fields will be returned

efirs: got it.

server/services/v1/query_runner.go
adilansari: done

efirs (approved):

PR #382, opened 2022-07-27
Added methods to calculate namespace,database,collection data size

efirs:
* Data sizes exposed in describe database and describe collection APIs.
* Fixed flaky admin tests
* Upgrade dependencies including Tigris client
* Rename DictionaryEncoder to MetadataDictionary and it's usage
  from `encoder` to `metaStore`.
* Remove dependency of KeyEncoder on TenantManager.
* Remove dependency of SessionManager on cdc.Manager and search.Store


store/kv/fdb.go
```diff
@@ -229,6 +229,24 @@ func (d *fdbkv) DropTable(ctx context.Context, name []byte) error {
...
+	s := subspace.FromBytes(name)
+
+	var sz int64
+	_, err := d.txWithRetry(ctx, func(tr fdb.Transaction) (interface{}, error) {
+		var err error
+		sz, err = tr.GetEstimatedRangeSizeBytes(s).Get()
+		return nil, err
+	})
+
+	log.Err(err).Str("name", string(name)).Int64("size", sz).Msg("table size")
```
himank, line 245: if err is nil then we don't want to log it? Or move it to debug log if err is nil.

store/kv/kv.go
```diff
@@ -60,6 +60,7 @@ type KeyValueStore interface {
 	CreateTable(ctx context.Context, name []byte) error
 	DropTable(ctx context.Context, name []byte) error
 	GetInternalDatabase() (interface{}, error) // TODO: CDC remove workaround
+	TableSize(ctx context.Context, name []byte) (int64, error)
```
himank, line 63: So Calling Table for anything may be confusing, we are using Table terminology in the code for encoded `namespace, database, collection`, should we just call it `KeyRangeSize` here?

store/kv/fdb.go
efirs: ok. fixing.
there is no functionality in zerolog to only log on error.

server/metadata/key_encoder.go
```diff
@@ -40,29 +41,20 @@ type Encoder interface {
...
+	return &DictKeyEncoder{}
 }
 
 type DictKeyEncoder struct {
-	mgr *TenantManager
 }
 
 func (d *DictKeyEncoder) EncodeTableName(ns Namespace, db *Database, coll *schema.DefaultCollection) ([]byte, error) {
-	if db == nil {
-		return nil, api.Errorf(api.Code_INVALID_ARGUMENT, "database is missing")
```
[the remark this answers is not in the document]
efirs: There are already checks at the callers of this function, for example both runner.GetDatbase, runner.GetCollection have this check.
So something will fail before getting to this function.

store/kv/kv.go
efirs: Currently there is a table concept used in every KV method.
I think it's easier to understand that keys are scoped by table and it'll be easier to move to other KV backend just in case.
But we can think and change it in later diffs, it would be a big change.

server/metadata/key_encoder.go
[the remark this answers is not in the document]
himank: Then let's document the API accordingly because till now the contract was strict so TableName maps to ns, db, and collection and when we decode we strictly know that we will get these three back but with your change, it can mean anything and any API may choose to just call this and Encode it by passing simply a single value(just namespace or just namespace and database).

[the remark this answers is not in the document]
efirs: documented

himank (approved):

PR #383, opened 2022-07-27
Optional rate, throughput and size quota checking

efirs:


server/midddleware/quota.go
```diff
@@ -0,0 +1,30 @@
...
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		ns, _ := request.GetNamespace(ctx)
+		if err := quota.Check(ctx, ns, proto.Size(req.(proto.Message))); err != nil {
+			return nil, err
+		}
+		return handler(ctx, req)
+	}
+}
+
+func quotaStreamServerInterceptor() grpc.StreamServerInterceptor {
```
ovaistariq, line 22: If I am streaming 1000 documents and I hit the quota on 100th document, what will be the behavior here?

ovaistariq:
@rbarabas now this is where envoyproxy would have been helpful https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting

server/midddleware/quota.go
efirs: This is per request quota. There is no read throughput quota. We can add it later.

server/metadata/tenant.go
```diff
@@ -135,6 +137,13 @@ func newTenantManager(kvStore kv.KeyValueStore, searchStore search.Store, mdName
 	}
 }
 
+func (m *TenantManager) EnsureDefaultNamespace(txMgr *transaction.Manager) error {
```
himank, line 140: What type of issues are you seeing?

server/quota/quota.go
```diff
@@ -0,0 +1,117 @@
...
+	txMgr     *transaction.Manager
+}
+
+var mgr Manager
+
+func Init(t *metadata.TenantManager, tx *transaction.Manager, c *config.QuotaConfig) {
+	mgr = *newManager(t, tx, c)
+}
+
+func Check(ctx context.Context, namespace string, size int) error {
```
himank, line 45: Check --> Allowed? Btw, who will be passing size? and what is this size?

server/config/options.go
```diff
@@ -159,6 +160,12 @@ var DefaultConfig = Config{
 			ResponseTime: true,
 		},
 	},
+	Quota: QuotaConfig{
+		Enabled:         true,
```
himank, line 164: Make sure to bump up the limit for the dev cluster after landing it otherwise it may impact the benchmarking.

server/config/options.go
```diff
@@ -159,6 +160,12 @@ var DefaultConfig = Config{
...
+		RateLimit:       1000,        // requests per second
```
himank, line 165: in prod, 1000 is very high per account per worker, it should be less than 100 for a single account.

server/quota/quota.go
```diff
@@ -0,0 +1,117 @@
...
+
+	return &Manager{cfg: c, tenantMgr: t, txMgr: tx}
+}
+
+func (m *Manager) check(ctx context.Context, namespace string, size int) error {
+	is, ok := m.quota.Load(namespace)
+	if !ok {
+		// Create new state if didn't exist before
+		is = &State{
+			Rate:       rate.NewLimiter(rate.Limit(m.cfg.RateLimit), m.cfg.RateLimit),
```
himank, line 62: the burst is very high for a single worker, should we restrict it to ~50 or so? Or some percentage derived from the RateLimit.

server/quota/quota.go
```diff
@@ -0,0 +1,117 @@
...
+	mgr.cfg = c
+	mgr.tenantMgr = t
+	mgr.txMgr = tx
+
+	return &Manager{cfg: c, tenantMgr: t, txMgr: tx}
+}
+
+func (m *Manager) check(ctx context.Context, namespace string, size int) error {
+	is, ok := m.quota.Load(namespace)
+	if !ok {
```
himank, line 59: btw, this is racy as all the threads will try to create the State object but the end state will be the same so I think it will be fine.

server/quota/quota.go
```diff
@@ -0,0 +1,117 @@
...
+type State struct {
+	Rate         *rate.Limiter
+	Throughput   *rate.Limiter
+	Size         atomic.Int64
+	SizeUpdateAt atomic.Int64
+	SizeLock     sync.Mutex
+}
+
+type Manager struct {
+	quota     sync.Map
```
himank, line 33: should we call it `tenantQuota` to make it clear that the tenant state is managed on this map.

server/metadata/tenant.go
```diff
@@ -248,11 +256,11 @@ func (m *TenantManager) GetTenant(ctx context.Context, namespaceName string, txM
 	}
 	id, ok := namespaces[namespaceName]
 	if !ok {
-		return nil, nil
+		return nil, fmt.Errorf("namespace not found: %s", namespaceName)
```
himank, line 259: do you need to return this from here? Or can the caller handle it? Also, returning like this means not an easy comparison for the caller to detect what type of error it is.

server/quota/quota.go
```diff
@@ -0,0 +1,117 @@
...
+	return &Manager{cfg: c, tenantMgr: t, txMgr: tx}
+}
+
+func (m *Manager) check(ctx context.Context, namespace string, size int) error {
+	is, ok := m.quota.Load(namespace)
+	if !ok {
+		// Create new state if didn't exist before
+		is = &State{
+			Rate:       rate.NewLimiter(rate.Limit(m.cfg.RateLimit), m.cfg.RateLimit),
+			Throughput: rate.NewLimiter(rate.Limit(m.cfg.ThroughputLimit), m.cfg.ThroughputLimit),
```
himank, line 63: burst is high.

server/quota/quota.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		m.quota.Store(namespace, is)
+	}
+
+	s := is.(*State)
+
+	if !s.Rate.Allow() {
```
himank, line 70: just FYI, this means high contention as the rate-limiter will take a lock which may impact throughput so better to also take a note on QPS before/after enabling it.

server/services/v1/sessions.go
```diff
@@ -62,15 +62,7 @@ func (sessMgr *SessionManager) Create(ctx context.Context, reloadVerOutside bool
...
-	var tenant *metadata.Tenant
-	if namespaceForThisSession == metadata.DefaultNamespaceName {
-		tenant, err = sessMgr.tenantMgr.CreateOrGetTenant(ctx, sessMgr.txMgr, metadata.NewDefaultNamespace())
-	} else {
-		tenant, err = sessMgr.tenantMgr.GetTenant(ctx, namespaceForThisSession, sessMgr.txMgr)
-	}
-	if tenant == nil {
-		return nil, api.Errorf(api.Code_NOT_FOUND, "Tenant %s not found", namespaceForThisSession)
-	}
+	tenant, err := sessMgr.tenantMgr.GetTenant(ctx, namespaceForThisSession, sessMgr.txMgr)
```
himank, line 65: how will we create a default namespace if it doesn't exist?

himank:
> @rbarabas now this is where envoyproxy would have been helpful https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_features/global_rate_limiting

But then we need to integrate it with Redis.

server/metadata/tenant.go
efirs: Tenant is required in quota middleware earlier than session manager is called.
Also creating it in session manager looks a bit hacky.

server/quota/quota.go
efirs: Will rename to "Allow".
Size is request payload size. We are limiting write throughput using it.

server/config/options.go
efirs: Right, I think it's better to disable by default then, and put in dev and prod cluster config specific values.

server/quota/quota.go
efirs: quota.tenantQuota doesn't sound right to me.
quota.Manager is better.

server/services/v1/sessions.go
efirs: It's created in the main.

server/quota/quota.go
efirs: This is per tenant and if we are talking about 100 requests per second, contention would be negligible.

server/quota/quota.go
efirs: yeah, maybe put it in the config too

server/metadata/tenant.go
himank: But why do you think it is hacky? It is the default namespace provided by the cluster if there is no namespace. It is only required in the middleware in the first request? But default namespace, we can create during startup if it is missing in the cluster?

server/quota/quota.go
himank: should we call it payload then? or reqSize?

server/config/options.go
himank: sg

server/quota/quota.go
himank: no, I mean for the map,`quota sync.Map`. The quota.Manager is fine.

server/quota/quota.go
himank: So in shared, yes. But in dedicated there may be just one tenant.

server/quota/quota.go
efirs: Yeah, that should be fine.

server/quota/quota.go
efirs: In the dedicated cluster there will be finer granular quotas.

server/quota/quota.go
efirs: So I think we should put max request size in the future here.

himank (approved):

PR #386, opened 2022-07-28
fix: Eventing framework shouldn't listen internal events

himank:


server/metadata/key_encoder_test.go
```diff
@@ -15,6 +15,7 @@
 package metadata
 
 import (
+	"github.com/tigrisdata/tigris/internal"
```
efirs, line 18: goimports

store/kv/listener.go
```diff
@@ -58,7 +61,15 @@ type DefaultListener struct {
 	Events []*Event
 }
 
+func (l *DefaultListener) shouldListen(table []byte) bool {
```
efirs, line 64: `skip`? `ignore`?

internal/data.go
```diff
@@ -24,6 +24,10 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+var (
+	UserTableKeyPrefix = []byte("data")
```
efirs, line 28: I think we should log DDLs too, but we can add it later also.

himank: DDLs will automatically get skipped with this approach because this prefix is added for any operation performed on the collection.

store/kv/listener.go
himank: sure

efirs (approved):

PR #387, opened 2022-07-28
fix: handle drop collection, database in streams

himank:


server/services/v1/api.go
```diff
@@ -411,19 +411,39 @@ func (s *apiService) Events(r *api.EventsRequest, stream api.Tigris_EventsServer
...
+			}
+
+			nsId, dbId, cId, ok := s.tenantMgr.GetEncoder().DecodeTableName(op.Table)
+			if !ok {
+				log.Err(err).Str("table", string(op.Table)).Msg("unexpected key in event streams")
+				return api.Errorf(api.Code_INTERNAL, "unexpected key in event streams")
+			}
+
+			if dbId != reqDatabaseId || cId != reqCollectionId {
+				// this probably means database/collection is dropped and this entry is some old entry in the log, ignore it
```
efirs, line 439: This just means that the event is not for the collection we are listening to

server/services/v1/api.go
```diff
@@ -411,19 +411,39 @@ func (s *apiService) Events(r *api.EventsRequest, stream api.Tigris_EventsServer
...
+				continue
+			}
+
+			_, _, collection, ok := s.tenantMgr.GetTableNameFromIds(nsId, dbId, cId)
```
efirs, line 443: We don't need collection name check here and below anymore.
ids of the event is equal to cached ids, so it must an event for the collection we are listening to

server/services/v1/api.go
himank: database as well, because the user may have dropped the existing database.

efirs (approved):

adilansari (approved):

PR #390, opened 2022-07-29
fix: Allow Searching and Faceting over nested fields

adilansari:
Validating search request against `queryable` fields instead of schema fields.


himank:
Great!

himank (approved):

PR #391, opened 2022-08-01
fix: get service names consistent

pboros:
Make service names consistent.

### DIFF
--- a/server/metrics/tracing.go
+++ b/server/metrics/tracing.go
@@ -22,9 +22,8 @@ import (
 )
 
 const (
-	GrpcTracingServiceName      = "tigris.grpc"
-	KvTracingServiceName        = "tigris.fdb.kv"
-	TxManagerTracingServiceName = "tigris.tx.manager"
+	KvTracingServiceName        = "kv"
+	TxManagerTracingServiceName = "txmanager"
 )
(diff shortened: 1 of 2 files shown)


server/metrics/tracing.go
```diff
@@ -22,9 +22,8 @@ import (
 )
 
 const (
-	GrpcTracingServiceName      = "tigris.grpc"
-	KvTracingServiceName        = "tigris.fdb.kv"
-	TxManagerTracingServiceName = "tigris.tx.manager"
+	KvTracingServiceName        = "tigris-server.fdb.kv"
```
rbarabas, line 25: Nice job adding these to `const`! Should the service here simply be called `fdb`?

server/metrics/tracing.go
```diff
@@ -22,9 +22,8 @@ import (
...
+	TxManagerTracingServiceName = "tigris-server.fdb.txmanager"
```
rbarabas, line 26: Would we lose a lot of context if we called this simply `txmanager`?

rbarabas (approved):

rbarabas (dismissed):

rbarabas (approved):

PR #393, opened 2022-08-01
fix: on drop remove entry from key generator table as well

himank:
- On dropping a collection, we need to drop the entry from the corresponding counter table that we use to allocate the int32 key.
- Changing search action to create for inserts (as we know it is strictly inserts so no need to use upserts)
- Renaming FDB hard drop config name.


server/config/options.go
```diff
@@ -21,9 +21,9 @@ import (
 )
 
 type ServerConfig struct {
-	Host      string
-	Port      int16
-	FDBDelete bool `mapstructure:"fdb_delete" yaml:"fdb_delete" json:"fdb_delete"`
+	Host        string
+	Port        int16
+	FDBHardDrop bool `mapstructure:"fdb_hard_drop" yaml:"fdb_hard_drop" json:"fdb_hard_drop"`
```
efirs, line 26: Should we call it just "HardDrop"?

efirs (approved):

PR #394, opened 2022-08-02
fix: fix datadog traces

pboros:
Fix datadog traces, all methods are visible in datadog traces under the same operation name.


rbarabas (approved):
LGTM based on our testing. Let's create a debug image and see how it fares!

PR #400, opened 2022-08-03
feat: added minimal pubsub implementation

mattayres:


schema/schema.go
```diff
@@ -180,6 +190,29 @@ func addPrimaryKeyIfMissing(reqSchema jsoniter.RawMessage) (jsoniter.RawMessage,
 	return jsoniter.Marshal(schema)
 }
 
+func setPrimaryKeyForMessages(reqSchema jsoniter.RawMessage) (jsoniter.RawMessage, error) {
```
himank, line 193: Possible to consolidate this with `addPrimaryKeyIfMissing` method? or extract out the common functionality? Looks like the idea of both the methods is the same except for the type of the id field.

server/services/v1/query_runner.go
```diff
@@ -720,6 +730,100 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
+		return nil, ctx, err
+	}
+
+	table, err := runner.encoder.EncodeTableName(tenant.GetNamespace(), db, collection)
+	if ulog.E(err) {
+		return nil, ctx, err
+	}
+
+	startTime := time.Now().UTC().Format(time.RFC3339Nano)
+	endTime := time.Now().AddDate(34, 0, 0).Format(time.RFC3339Nano)
```
himank, line 757: If endTime is very high then the read-range will lock this whole range and any future write will conflict, should we move this to snapshot read?

server/services/v1/query_runner.go
```diff
@@ -720,6 +730,100 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
+			return nil, ctx, err
+		}
+
+		kvIterator, err := tickerTx.ReadRange(ctx, startKey, endKey)
+		if ulog.E(err) {
+			return nil, ctx, err
+		}
+
+		var last []byte
+		for {
```
himank, line 783: remove if and move the condition in `for`?
```
		var keyValue kv.KeyValue
		for kvIterator.Next(&keyValue) {
			err = runner.streaming.Send(&api.SubscribeResponse{
				Message: keyValue.Data.RawData,
			})
			if ulog.E(err) {
				return nil, ctx, err
			}
		}
		if ulog.E(kvIterator.Err()) {
			return nil, ctx, kvIterator.Err()
		}
```

server/services/v1/query_runner.go
```diff
@@ -720,6 +730,100 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
+				break
+			}
+
+			if ulog.E(kvIterator.Err()) {
+				return nil, ctx, kvIterator.Err()
+			}
+		}
+
+		if last != nil {
+			var data map[string]any
```
himank, line 804: the timestamp is part of the key, do we need to unmarshal the body? Can we just extract it out from the key?

server/services/v1/query_runner.go
```diff
@@ -720,6 +730,100 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
+			if ulog.E(err) {
+				return nil, ctx, err
+			}
+
+			lastTime, err := time.Parse(time.RFC3339, data[schema.AutoPrimaryKeyF].(string))
+			if ulog.E(err) {
+				return nil, ctx, err
+			}
+
+			startTime = lastTime.Add(1 * time.Nanosecond).Format(time.RFC3339)
```
himank, line 815: Or read from the last time but throw away the first row that is already sent? To avoid any corner case missing entry issue. That will also ensure the correctness of the stream that all rows are sent since the first start time of the request.

server/services/v1/query_runner.go
```diff
@@ -720,6 +730,100 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
+		}
+
+		err = tickerTx.Commit(ctx)
```
himank, line 818: add a ToDo that as this is a read-only transaction we should ignore any error like transaction timed-out etc.

server/services/v1/query_runner.go
```diff
@@ -720,6 +730,100 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
+	}
+
+	startTime := time.Now().UTC().Format(time.RFC3339Nano)
+	endTime := time.Now().AddDate(34, 0, 0).Format(time.RFC3339Nano)
+	endKey, err := runner.encoder.EncodeKey(table, collection.Indexes.PrimaryKey, []interface{}{endTime})
+	if ulog.E(err) {
+		return nil, ctx, err
+	}
+
+	ticker := time.NewTicker(500 * time.Millisecond)
```
himank, line 763: let's start with 100 msec? not for this diff, but this needs to be moved to config.

server/services/v1/query_runner.go
```diff
@@ -720,6 +730,96 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
+		if ulog.E(err) {
+			return nil, ctx, err
+		}
+
+		end, err := time.Parse(time.RFC3339, startTime)
+		if ulog.E(err) {
+			return nil, ctx, err
+		}
+
+		endTime := end.Add(34 * time.Second).Format(time.RFC3339)
```
himank, line 778: We still need to do a snapshot read otherwise the writes may see conflict in this window but that can be done in next diff.

himank (approved):

server/services/v1/query_runner.go
mattayres: Yeah, will do next iteration.

himank (dismissed):

himank (approved):

PR #401, opened 2022-08-03
feat: Added minimal support for oauth

JigarJoshi:


server/config/options.go
```diff
@@ -118,6 +121,9 @@ var DefaultConfig = Config{
 		LogOnly:                  true,
 		EnableNamespaceIsolation: false,
 		AdminNamespaces:          []string{"tigris-admin"},
+		ClientId:                 "",
```
efirs, line 124: Zero values of the type do not need to be assigned explicitly.
Just remove this three lines.
This may be removed as well `EnableNamespaceIsolation: false`

server/services/v1/auth.go
```diff
@@ -0,0 +1,115 @@
...
+type authService struct {
+	api.UnimplementedAuthServer
+	OAuthProvider
+}
+
+type OAuthProvider interface {
+	GetAccessToken(request *api.GetAccessTokenRequest) (*api.GetAccessTokenResponse, error)
+}
+
+type Auth0Oauth struct {
```
efirs, line 48: The name looks funny :)

server/services/v1/service.go
```diff
@@ -35,11 +36,12 @@ type Service interface {
...
 }
 
-func GetRegisteredServices(kvStore kv.KeyValueStore, searchStore search.Store, tenantMgr *metadata.TenantManager, txMgr *transaction.Manager) []Service {
+func GetRegisteredServices(kvStore kv.KeyValueStore, searchStore search.Store, tenantMgr *metadata.TenantManager, txMgr *transaction.Manager, defaultConfig config.Config) []Service {
 	var v1Services []Service
 
 	v1Services = append(v1Services, newApiService(kvStore, searchStore, tenantMgr, txMgr))
 	v1Services = append(v1Services, newHealthService())
 	v1Services = append(v1Services, newAdminService(tenantMgr, txMgr))
+	v1Services = append(v1Services, newAuthService(defaultConfig.Auth))
```
efirs, line 45: can we just access config.DefaultConfig here and not pass it around?

server/services/v1/auth.go
```diff
@@ -0,0 +1,115 @@
...
+
+func newAuthService(authConfig config.AuthConfig) *authService {
+	if authConfig.OAuthProviderType == "auth0" {
+		return &authService{
+			OAuthProvider: Auth0Oauth{
+				authConfig,
+			},
+		}
+	}
+	log.Error().Msg("Unable to configure external oauth provider")
```
efirs, line 90: add `.Str("type", authConfig.OAuthProviderType).` to the log.

efirs: Actually, I am thinking if we should crash here?

server/services/v1/auth.go
efirs: May be just Auth0 would work.

server/config/options.go
```diff
@@ -46,6 +46,9 @@ type AuthConfig struct {
 	LogOnly                  bool          `mapstructure:"log_only" yaml:"log_only" json:"log_only"`
 	EnableNamespaceIsolation bool          `mapstructure:"enable_namespace_isolation" yaml:"enable_namespace_isolation" json:"enable_namespace_isolation"`
 	AdminNamespaces          []string      `mapstructure:"admin_namespaces" yaml:"admin_namespaces" json:"admin_namespaces"`
+	OAuthProviderType        string        `mapstructure:"oauth_provider" yaml:"oauth_provider" json:"oauth_provider"`
```
efirs, line 49: maybe just `OAuthProvider`

server/config/options.go
JigarJoshi: removed.

server/services/v1/auth.go
JigarJoshi: renamed. I agree auth0 name is much simpler here.

server/services/v1/service.go
JigarJoshi: updated.

server/services/v1/auth.go
JigarJoshi: yes I think we should crash here too. updated log and updated it to crash for such exceptional configuration.

server/config/options.go
JigarJoshi: renamed.

server/services/v1/auth.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (a Auth0) GetAccessToken(req *api.GetAccessTokenRequest) (*api.GetAccessTokenResponse, error) {
+	data := url.Values{
+		"refresh_token": {req.RefreshToken},
+		"client_id":     {a.AuthConfig.ClientId},
+		"grant_type":    {refreshToken},
+		"scope":         {scope},
+	}
+	resp, err := http.PostForm(a.AuthConfig.ExternalTokenURL, data)
```
himank, line 59: the response must be having status code in case it is not null, let's log it or return it with the error. Would help in debugging.

server/services/v1/auth.go
```diff
@@ -0,0 +1,118 @@
...
+	refreshToken = "refresh_token"
+	scope        = "offline_access openid"
+)
+
+type authService struct {
+	api.UnimplementedAuthServer
+	OAuthProvider
+}
+
+type OAuthProvider interface {
```
himank, line 44: Ideally, the provider should be outside the auth services because based on the provider there may be custom rules/custom validation/etc that we need to honor in the middleware where we are validating the request. The current way is embedding the provider inside the auth service, it should be passed from the outside. But we don't need to make that change now and it is fine to make this change later.

server/services/v1/auth.go
```diff
@@ -0,0 +1,118 @@
...
+	bodyStr := string(body)
+	if resp.StatusCode == 200 {
+		getAccessTokenResponse := api.GetAccessTokenResponse{}
+
+		err = json.Unmarshal([]byte(bodyStr), &getAccessTokenResponse)
+		if err != nil {
+			return nil, api.Errorf(api.Code_INTERNAL, "Failed to parse external response: reason = "+err.Error())
+		}
+		return &getAccessTokenResponse, nil
+	} else {
```
efirs, line 77: else is not needed here, just return the error

server/services/v1/service.go
```diff
@@ -35,11 +36,12 @@ type Service interface {
 	RegisterGRPC(grpc *grpc.Server) error
 }
 
-func GetRegisteredServices(kvStore kv.KeyValueStore, searchStore search.Store, tenantMgr *metadata.TenantManager, txMgr *transaction.Manager) []Service {
+func GetRegisteredServices(kvStore kv.KeyValueStore, searchStore search.Store, tenantMgr *metadata.TenantManager, txMgr *transaction.Manager, defaultConfig config.Config) []Service {
```
efirs, line 39: this is not neede now.

server/services/v1/auth.go
JigarJoshi: added logging.

server/services/v1/auth.go
JigarJoshi: sounds good.

server/services/v1/auth.go
JigarJoshi: removed else.

server/services/v1/service.go
JigarJoshi: yes removed it.

efirs (dismissed):

efirs (dismissed):

efirs (approved):

PR #403, opened 2022-08-04
Improve integrations tests

efirs:
* Allow to run/debug individual tests in IDE
* Fix local_run target. Now FDB on the host should not be needed.
* Fix exit code of `make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,7 @@ docker_compose_build:
 # current user owner instead of root
 docker_test: generate
 	$(DOCKER_COMPOSE) up --build tigris_test tigris_test
+	@[ $$(docker inspect tigris_test --format='{{.State.ExitCode}}') = "0" ]
 
 docker_test_no_build:
 	$(DOCKER_COMPOSE) up --exit-code-from tigris_test --no-build tigris_test tigris_test
@@ -64,9 +65,9 @@ run: generate
 	$(DOCKER_COMPOSE) up --build --detach tigris_server2
 
 local_run: server
-	$(DOCKER_COMPOSE) up --no-build --detach tigris_search tigris_db
+	$(DOCKER_COMPOSE) up --no-build --detach tigris_search tigris_db2
 	fdbcli -C ./test/config/fdb.cluster --exec "configure new single memory" || true
-	TIGRIS_ENVIRONMENT=dev ./server/service
+	./server/service -c config/server.dev.yaml
 
 # Runs tigris server and foundationdb, plus additional tools for it like:
 # - prometheus and grafana for monitoring
--- a/test/v1/server/integration_test.go
+++ b/test/v1/server/integration_test.go
@@ -19,45 +19,168 @@ package server
 import (
 	"fmt"
 	"math/rand"
+	"net/http"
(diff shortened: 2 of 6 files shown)


himank (approved):

test/v1/server/integration_test.go
```diff
@@ -19,45 +19,169 @@ package server
...
+					},
+				},
+			},
+		},
+		"primary_key": []interface{}{"pkey_int"},
+	},
+}
+
+func setupTests(t *testing.T) (string, string) {
+	db := fmt.Sprintf("integration_%s_%x", t.Name(), rand.Uint64())
```
himank, line 119: Should we not do this? Let's see what all is breaking if we reuse the same name?

ovaistariq (approved):

Makefile
```diff
@@ -64,9 +65,9 @@ run: generate
 	$(DOCKER_COMPOSE) up --build --detach tigris_server2
 
 local_run: server
-	$(DOCKER_COMPOSE) up --no-build --detach tigris_search tigris_db
+	$(DOCKER_COMPOSE) up --no-build --detach tigris_search tigris_db2
 	fdbcli -C ./test/config/fdb.cluster --exec "configure new single memory" || true
-	TIGRIS_ENVIRONMENT=dev ./server/service
+	./server/service -c config/server.dev.yaml
```
rbarabas, line 70: Nice!

rbarabas (approved):

test/v1/server/integration_test.go
```diff
@@ -19,45 +19,169 @@ package server
...
+	//	coll := fmt.Sprintf("integration_coll1_%x", rand.Uint64())
+	dropDatabase(t, db)
+	createDatabase(t, db).Status(http.StatusOK)
+	createCollection(t, db, testCollection, testCreateSchema).Status(http.StatusOK)
+
+	return db, testCollection
+}
+
+func cleanupTests(t *testing.T, db string) {
```
himank, line 128: nit: instead of this you can return the cleanup func from the setupTests so that whatever we are setting up the cleanup is in sync with that setup.

test/v1/server/integration_test.go
```diff
@@ -19,45 +19,169 @@ package server
...
+				},
+			},
+		},
+		"primary_key": []interface{}{"pkey_int"},
+	},
+}
+
+func setupTests(t *testing.T) (string, string) {
+	db := fmt.Sprintf("integration_%s_%x", t.Name(), rand.Uint64())
+	//	coll := fmt.Sprintf("integration_coll1_%x", rand.Uint64())
```
himank, line 120: remove this line.

test/v1/server/integration_test.go
efirs: There is not much difference. setup and cleanup tests defined next to each other.
But yes, we may change it in the future.

himank: yes, but this way of setup and cleanup is very broad and maybe not be applicable for all use-cases. Attaching it to the setup reduces it.

himank (approved):

PR #404, opened 2022-08-08
Fix search response marshalling

efirs:
Skip custom marshalling wherever it's not needed.

### DIFF
--- a/api/server/v1/marshaler.go
+++ b/api/server/v1/marshaler.go
@@ -85,7 +85,7 @@ func (x *ReadRequest) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// UnmarshalJSON for SearchRequest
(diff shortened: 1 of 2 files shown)


api/server/v1/marshaler.go
```diff
@@ -438,26 +438,6 @@ func (x *ReadResponse) MarshalJSON() ([]byte, error) {
 	return json.Marshal(resp)
 }
 
-func (x *SearchResponse) MarshalJSON() ([]byte, error) {
-	resp := struct {
-		Hits   []*SearchHit            `json:"hits"`
```
adilansari, line 443: We explicitly want to have a response like:

```
{
    "hits": [],
    "facets":  {},
    "meta": {"found": 0, "total_pages": 0, "page": 0}
}
```

adilansari (approved):

PR #406, opened 2022-08-09
feat: Added basic observability service implementation

JigarJoshi:
TODO:

- [ ] Add namespace tag to scope metrics query.
- [ ] Configure allowed queries.


server/services/v1/observability.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, api.Errorf(api.Code_INTERNAL, "Failed to query metrics: reason = "+err.Error())
+	}
+
+	bodyStr := string(body)
+	if resp.StatusCode == 200 {
+		var ddResp MetricsQueryResp
+		err = json.Unmarshal(body, &ddResp)
```
efirs, line 97: add error check.

server/services/v1/observability.go
```diff
@@ -0,0 +1,155 @@
...
+		var dataPoints = make([]*api.DataPoint, len(ddResp.Series[0].PointList))
```
efirs, line 98: check that there is atleast one series in the response before accessing it.

server/services/v1/observability.go
```diff
@@ -0,0 +1,155 @@
...
+		for i := range ddResp.Series[0].PointList {
+			dataPoints[i] = &api.DataPoint{}
+			dataPoints[i].Timestamp = int64(ddResp.Series[0].PointList[i][0])
```
efirs, line 101: Do we need to set value too?

server/services/v1/observability.go
JigarJoshi: fixed.

server/services/v1/observability.go
JigarJoshi: added check.

server/services/v1/observability.go
JigarJoshi: yes it was totally missed. thanks.

server/services/v1/observability.go
```diff
@@ -0,0 +1,163 @@
...
+type Datadog struct {
+	config.ObservabilityConfig
+}
+
+type MetricsQueryResp struct {
+	FromDate int64 `json:"from_date"`
+	ToDate   int64 `json:"to_date"`
+	Query    string
+	Series   []struct {
+		PointList [][]float64 `json:"pointlist"`
```
efirs, line 60: Can we declare it like below? So we can remove size checks after unmarshaling:
```
Series [1]struct {
		PointList [][2]float64 `json:"pointlist"`
```

JigarJoshi: yes definitely.

JigarJoshi:
abandoning this PR. I will create new one.

PR #412, opened 2022-08-11
Support read throughput limiting

efirs:


server/config/options.go
```diff
@@ -137,8 +137,9 @@ var DefaultConfig = Config{
 	},
 	Quota: QuotaConfig{
 		Enabled:              false,
-		RateLimit:            1000,        // requests per second
+		RateLimit:            1000,        // requests per second both reads and writes
```
himank, line 140: Should we reduce the defaults? So that we can start testing it on the preview cluster?

himank (dismissed):

server/config/options.go
efirs: Yes, it's disable for now. Will adjust it later.

himank (dismissed):

himank (approved):

JigarJoshi (approved):

PR #415, opened 2022-08-11
feat: instrumentation for session

pboros:
Instrumentation for session manager (both tracing and metrics)

### DIFF
--- a/server/services/v1/sessions.go
+++ b/server/services/v1/sessions.go
@@ -23,17 +23,28 @@ import (
 	"github.com/rs/zerolog/log"
 	api "github.com/tigrisdata/tigris/api/server/v1"
 	"github.com/tigrisdata/tigris/server/metadata"
+	"github.com/tigrisdata/tigris/server/metrics"
 	"github.com/tigrisdata/tigris/server/middleware"
 	"github.com/tigrisdata/tigris/server/request"
 	"github.com/tigrisdata/tigris/server/transaction"
 	"github.com/tigrisdata/tigris/store/kv"
 	ulog "github.com/tigrisdata/tigris/util/log"
+	"github.com/uber-go/tally"
 )
 
 // SessionManager is used to manage all the explicit query sessions. The execute method is executing the query.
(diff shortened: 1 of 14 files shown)


server/services/v1/sessions.go
```diff
@@ -16,6 +16,8 @@ package v1
 
 import (
 	"context"
+	"github.com/tigrisdata/tigris/server/metrics"
```
himank, line 19: goimports.

server/services/v1/sessions.go
```diff
@@ -45,6 +56,64 @@ type SessionManager struct {
...
+
+func (m SessionManagerWithMetrics) Create(ctx context.Context, trackVerInOwnTxn bool, instantVerTracking bool, track bool) (qs *QuerySession, err error) {
+	m.measure(ctx, "Create", func(ctx context.Context) error {
+		qs, err = m.s.Create(ctx, trackVerInOwnTxn, instantVerTracking, track)
+		return err
+	})
+	return
+}
+
+func (m SessionManagerWithMetrics) Get(ctx context.Context) (qs *QuerySession, err error) {
```
himank, line 85: We can skip measuring Get/Remove because this is an in-memory operation only so overhead here will be mostly for reporting metrics.

server/services/v1/sessions.go
```diff
@@ -45,6 +56,64 @@ type SessionManager struct {
...
+
+func (m SessionManagerWithMetrics) Execute(ctx context.Context, req *ReqOptions) (resp *Response, err error) {
+	m.measure(ctx, "Execute", func(ctx context.Context) error {
+		resp, err = m.s.Execute(ctx, req)
+		return err
+	})
+	return
+}
+
+func (m SessionManagerWithMetrics) executeWithRetry(ctx context.Context, req *ReqOptions) (resp *Response, err error) {
```
himank, line 109: as you're measuring `Execute` do we need to measure `executeWithRetry`? Here probably again we may not get much insight from measuring this function and the Execute should be enough.

himank (approved):

himank (approved):

PR #416, opened 2022-08-11
feat: Support for ordering search results

adilansari:
- Ability to order search results by specifying an order and an additional tie breaker
- Only available to `numeric` fields at the moment

#### Sample request
```json
{
    "q": "",
    "sort": [
            {"price":"$desc"},
            {"review.votes.upvotes":"$asc"}
        ]
}
```

### DIFF
--- a/query/search/sort.go
+++ b/query/search/sort.go
@@ -1,0 +1,89 @@
+package search
+
+import (
+	"github.com/buger/jsonparser"
+	jsoniter "github.com/json-iterator/go"
+	api "github.com/tigrisdata/tigris/api/server/v1"
+)
+
+const maxSortOrders = 3
+
+const (
+	ASC  = "$asc"
+	DESC = "$desc"
+)
+
+type Ordering = []SortField
+
+type SortField struct {
+	// Required; Name of field to enable sorting for
+	Name string
+	// Required; True if ascending order is requested, False for descending
+	Ascending bool
+	// Optional; True if missing/empty/null values to be presented at the top of sort order,
+	// else they are sorted to the end by default
+	MissingValuesFirst bool
+}
+
+func newSortField(order jsoniter.RawMessage) (SortField, error) {
+	var s SortField
+	err := jsonparser.ObjectEach(order, func(k []byte, v []byte, vt jsonparser.ValueType, offset int) error {
+		switch string(v) {
+		case ASC:
+			s.Ascending = true
+		case DESC:
+			s.Ascending = false
+		default:
+			return api.Errorf(api.Code_INVALID_ARGUMENT, "Sort order can only be `%s` or `%s`", ASC, DESC)
+		}
+		s.Name = string(k)
+		s.MissingValuesFirst = false
+		return nil
+	})
+	if err != nil {
+		return s, err
+	}
+	return s, nil
+}
+
+func UnmarshalSort(input jsoniter.RawMessage) (*Ordering, error) {
+	if len(input) == 0 {
+		return nil, nil
+	}
+
+	var orders = Ordering{}
+	var err error
+	_, err2 := jsonparser.ArrayEach(input, func(item []byte, vt jsonparser.ValueType, offset int, err1 error) {
+		if err1 != nil {
+			err = err1
+			return
+		}
+
+		if vt != jsonparser.Object {
+			err = api.Errorf(api.Code_INVALID_ARGUMENT, "Invalid value for `%s`", "sort")
+			return
+		}
+
+		if len(orders) >= maxSortOrders {
+			err = api.Errorf(api.Code_INVALID_ARGUMENT, "Sorting can support up to `%d` fields only", maxSortOrders)
+			return
+		}
+
+		var f SortField
+		f, err = newSortField(item)
+		if err != nil {
+			return
+		}
+		orders = append(orders, f)
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	if err2 != nil {
+		return nil, err2
+	}
+
+	return &orders, nil
+}
--- a/query/search/sort.go
+++ b/query/search/sort.go
@@ -38,7 +38,7 @@ func newSortField(order jsoniter.RawMessage) (SortField, error) {
 			return api.Errorf(api.Code_INVALID_ARGUMENT, "Sort order can only be `%s` or `%s`", ASC, DESC)
 		}
 		s.Name = string(k)
-		s.MissingValuesFirst = false
+		s.MissingValuesFirst = false // Forcing empty/null/missing values to the end
 		return nil
 	})
 	if err != nil {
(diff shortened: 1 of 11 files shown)


query/search/sort.go
```diff
@@ -0,0 +1,90 @@
+package search
+
+import (
+	"github.com/buger/jsonparser"
+	jsoniter "github.com/json-iterator/go"
+	api "github.com/tigrisdata/tigris/api/server/v1"
+)
+
+// TODO: Update this to 3 once https://github.com/typesense/typesense/issues/690 is resolved
+const maxSortOrders = 2
```
adilansari, line 10: I have raised an issue with TS, it appears that they may have fixed this in `v24`. For now limiting our implementation to a primary sort order and 1 tie-breaker only. Once, the issue is resolved, we can accept two tie-breaker ranking params

query/search/sort.go
```diff
@@ -0,0 +1,90 @@
...
+type Ordering = []SortField
+
+type SortField struct {
+	// Required; Name of field to enable sorting for
+	Name string
+	// Required; True if ascending order is requested, False for descending
+	Ascending bool
+	// Optional; True if missing/empty/null values to be presented at the top of sort order,
+	// else they are sorted to the end by default
+	MissingValuesFirst bool
```
adilansari, line 26: We are not accepting this as user input for now, just pushing all empty/missing/null values down the order.

schema/collection.go
```diff
@@ -173,6 +173,7 @@ func buildSearchSchema(name string, queryableFields []*QueryableField) *tsApi.Co
 			Type:     s.SearchType,
 			Facet:    &s.Faceted,
 			Index:    &s.Indexed,
+			Sort:     &s.Sortable,
```
adilansari, line 176: Write the sort parameter downstream when creating collection schema in TS

server/services/v1/query_runner.go
```diff
@@ -560,6 +560,11 @@ func (runner *SearchQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 		return nil, ctx, err
 	}
 
+	sortOrder, err := runner.getSortOrdering(collection.GetQueryableFields())
```
adilansari, line 563: parse `sort` from search request input

store/search/ts.go
```diff
@@ -185,6 +185,9 @@ func (s *storeImpl) getBaseSearchParam(query *qsearch.Query, pageNo int) tsApi.M
 			baseParam.MaxFacetValues = &size
 		}
 	}
+	if sortBy := query.ToSortFields(); len(sortBy) > 0 {
```
adilansari, line 188: send sort params to TS

query/search/sort.go
```diff
@@ -0,0 +1,90 @@
+package search
```
himank, line 1: as sorting would be applicable for read queries as well, let's move sorting to its own package.

query/search/sort.go
```diff
@@ -0,0 +1,90 @@
...
+	DESC = "$desc"
+)
+
+type Ordering = []SortField
+
+type SortField struct {
+	// Required; Name of field to enable sorting for
+	Name string
+	// Required; True if ascending order is requested, False for descending
+	Ascending bool
```
himank, line 23: Do you know if the cost of both ascending and descending is same? Or do you maintain some order in memory?

schema/collection.go
himank: So as per their docs looks like for numeric fields it is enabled by default, so do we need to enable it in the schema?

> While sorting is enabled by default on numerical and boolean fields, sorting on a string field is only allowed if that field has sort property enabled in the collection schema

https://typesense.org/docs/0.23.1/api/search.html#sort-results

server/services/v1/query_runner.go
```diff
@@ -736,6 +742,31 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
 
+func (runner *SearchQueryRunner) getSortOrdering(collFields []*schema.QueryableField) (*qsearch.Ordering, error) {
+	ordering, err := qsearch.UnmarshalSort(runner.req.GetSort())
+	if err != nil || ordering == nil {
+		return nil, err
+	}
+
+	for _, sf := range *ordering {
+		found := false
+		for _, cf := range collFields {
```
himank, line 753: not for this diff, but probably we can have a `GetField(name string)` method on the collection. Then the few checks that we are doing in the case of facet search, sorting, etc can be unified by just getting this field from the collection and if not present returning an error otherwise check if the field can be used to serve the request.

query/search/sort.go
```diff
@@ -0,0 +1,90 @@
...
+		s.MissingValuesFirst = false
+		return nil
+	})
+	if err != nil {
+		return s, err
+	}
+	return s, nil
+}
+
+func UnmarshalSort(input jsoniter.RawMessage) (*Ordering, error) {
```
himank, line 50: nit: add a comment on the expected JSON.

schema/collection.go
adilansari: Relying on their defaults can cause unexpected changes in future, if they change this behavior, it can go un-notice once we upgrade i.e. collections created post upgrade may have a different behavior. We should be explicit in our definition.

query/search/sort.go
adilansari: I think order is generated in real time based on `text match` score for the query or `sort_by` parameter. At least, documentation does not say anything about ranking performance. 

If `default_sorting_field` is specified in collection definition, there could be some performance improvement from keeping dataset ordered in memory.

schema/collection.go
himank: I don't think we should worry about that because that will be a breaking change and it is okay to rely on these functionalities if anything changes then we can make it work for our forked version. But my main concern is when it is specified in the schema then that means they need to do additional work with each write(additional in-memory indexing structure) which means even though users may not be using this feature or don't even care about it there is an extra cost for maintaining it. 

Rather if we move the cost during the query time which means if a user has an intended use-case then they are aware that the query will be expensive because of sorting.

server/services/v1/query_runner.go
```diff
@@ -717,25 +710,39 @@ func (runner *SearchQueryRunner) getFieldSelection(collFields []*schema.Queryabl
...
 }
 
+func (runner *SearchQueryRunner) getSortOrdering(coll *schema.DefaultCollection) (*sort.Ordering, error) {
+	ordering, err := sort.UnmarshalSort(runner.req.GetSort())
+	if err != nil || ordering == nil {
+		return nil, err
+	}
+
+	for _, sf := range *ordering {
+		cf, err := coll.GetQueryableField(sf.Name)
```
himank, line 734: Neat!

himank (approved):

PR #417, opened 2022-08-11
fix: Fixed oauth/token endpoint to accept x-www-form-urlencoded

JigarJoshi:
Fixed oauth/token endpoint to accept application/x-www-form-urlencoded payload

### DIFF
--- a/api/server/v1/marshaler.go
+++ b/api/server/v1/marshaler.go
@@ -16,18 +16,61 @@ package api
 
 import (
 	"encoding/json"
+	"io"
+	"net/url"
 	"time"
 
+	"github.com/ajg/form"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	jsoniter "github.com/json-iterator/go"
 	spb "google.golang.org/genproto/googleapis/rpc/status"
 	"google.golang.org/protobuf/proto"
 )
 
+const (
+	refreshToken = "refresh_token"
+)
+
+type CustomDecoder struct {
+	formDecoder *form.Decoder
+	jsonDecoder *json.Decoder
+	reader      io.Reader
+}
+
+func (f CustomDecoder) Decode(dst interface{}) error {
+	switch dst.(type) {
+	case *GetAccessTokenRequest:
+		{
+			byteArr, err := io.ReadAll(f.reader)
+			if err != nil {
+				return err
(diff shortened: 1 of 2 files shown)


api/server/v1/marshaler.go
```diff
@@ -507,10 +575,10 @@ func (x *FacetStats) MarshalJSON() ([]byte, error) {
...
 		Count int64    `json:"count"`
 	}{
-		Avg:   x.Avg,
-		Max:   x.Max,
-		Min:   x.Min,
-		Sum:   x.Sum,
+		Avg:   &x.Avg,
+		Max:   &x.Max,
+		Min:   &x.Min,
+		Sum:   &x.Sum,
```
JigarJoshi, line 581: this is not related change.

api/server/v1/marshaler.go
```diff
@@ -16,18 +16,61 @@ package api
...
+		}
+	}
+	return f.jsonDecoder.Decode(dst)
+}
+
 // CustomMarshaler is a marshaler to customize the response. Currently, it is only used to marshal custom error message
 // otherwise it just uses the inbuilt mux marshaller.
 type CustomMarshaler struct {
-	*runtime.JSONBuiltin
+	JsonBuiltin *runtime.JSONBuiltin
```
efirs, line 57: We can keep embedded version or we should rename the member to JSONBuiltin otherwise linters would complain. It expectes JSON to be capitalized.

JigarJoshi: updated the variable name and keeping the variable here to be clear explicitly on how the methods are implemented.

efirs (approved):

api/server/v1/marshaler.go
```diff
@@ -53,6 +96,31 @@ func (c *CustomMarshaler) Marshal(v interface{}) ([]byte, error) {
 	return c.JSONBuiltin.Marshal(v)
 }
 
+func (c *CustomMarshaler) Unmarshal(data []byte, v interface{}) error {
+	switch v.(type) {
+	case *GetAccessTokenRequest:
+		{
```
himank, line 102: inside the case, you don't need any braces i.e. 
```
switch v.(type) {
	case *GetAccessTokenRequest:
	    return unmarshalInternal(data, v)
}
```

PR #419, opened 2022-08-12
feat: Size metrics for namespace, database and collection

pboros:
* Namespace level metrics are updated on quota calculation refresh, they are not appearing if quotas are disabled.
* Database level metrics are updated on DescribeDatabase requests.
* Collections level metrics are updated on DescribeCollection requests (metrics are updated for that single collection) and on DescribeDatabase requests (metrics are updated for all collections in the database).

### DIFF
--- a/server/metrics/requests.go
+++ b/server/metrics/requests.go
@@ -21,12 +21,19 @@ import (
 
 	"github.com/tigrisdata/tigris/server/config"
 	"github.com/tigrisdata/tigris/server/request"
+	"github.com/uber-go/tally"
 	"google.golang.org/grpc"
 )
 
 const (
 	SystemTigrisTenantName = "system"
 	UnknownValue           = "unknown"
+)
+
+var (
+	OkRequests       tally.Scope
+	ErrorRequests    tally.Scope
+	RequestsRespTime tally.Scope
 )
 
 type RequestEndpointMetadata struct {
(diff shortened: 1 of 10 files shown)


server/metrics/requests.go
```diff
@@ -17,6 +17,7 @@ package metrics
 import (
 	"context"
 	"fmt"
+	"github.com/uber-go/tally"
```
efirs, line 20: goimports

server/quota/quota.go
```diff
@@ -2,6 +2,7 @@ package quota
 
 import (
 	"context"
+	"github.com/tigrisdata/tigris/server/metrics"
```
efirs, line 5: goimports

server/services/v1/query_runner.go
```diff
@@ -935,6 +937,11 @@ func (runner *CollectionQueryRunner) Run(ctx context.Context, tx transaction.Tx,
 			return nil, ctx, err
 		}
 
+		namespace, err := request.GetNamespace(ctx)
```
efirs, line 940: Can we have a wrapper for it?
so we don't need to error check every time.

pboros: Makes sense, moved it with other namespace related tagging helper.

efirs (approved):

PR #421, opened 2022-08-15
fixed multiple issues with subscribe query runner

mattayres:


server/services/v1/query_runner.go
```diff
@@ -824,11 +826,19 @@ func (runner *SubscribeQueryRunner) Run(ctx context.Context, tx transaction.Tx,
 			startTime = keyValue.Key[1].(string)
 		}
 
-		// TODO: read-only transaction should ignore any error like transaction timed-out
 		err = tickerTx.Commit(ctx)
-		if ulog.E(err) {
+		var ep fdb.Error
```
himank, line 830: Usually, we are avoiding accessing the fdb package directly in the services layer. Can we move this in a helper method in the kv package?

mattayres: Done.

himank (dismissed):

himank (approved):

PR #422, opened 2022-08-15
feat: Fallback to full table scan if indexing store is not available/disabled

himank:
- Supporting read by any fields even if indexing store is disabled
- Implicit transactions now allow long-running reads
- Introduce ReadOnly Session so that implicit transactional reads do not create any transaction if reads need to happen from the indexing store. Also, moving out the search API to ReadOnlySession so that it happens without creating any transaction.

### DIFF
--- a/keys/key.go
+++ b/keys/key.go
@@ -14,7 +14,15 @@
 
 package keys
 
-import "fmt"
+import (
+	"bytes"
+	"fmt"
+	"unsafe"
+
+	"github.com/apple/foundationdb/bindings/go/src/fdb"
+	"github.com/apple/foundationdb/bindings/go/src/fdb/subspace"
+	"github.com/apple/foundationdb/bindings/go/src/fdb/tuple"
+)
 
 // Key is an interface that provides an encoded key which will be used for storing Key, Value in FDB. The Key has two
 // elements, the first set of bytes is the encoded table name and the remaining is the actual index values.
@@ -30,6 +38,12 @@ type Key interface {
 	// key index or some other user defined index. Essentially to encode key for a given row. Some other internal packages
 	// may use this as simply to form a key without any significance of index identifier.
 	IndexParts() []interface{}
+	// SerializeToBytes follows the ordering of how the Key is persisted in database so to compare a Key call this method
+	// get bytes and compare it with raw bytes stored in database.
+	SerializeToBytes() []byte
+	// CompareBytes compares the serialized form of keys. It returns 0 if p == input, -1 if p < input, and +1 if p > input.
+	// A nil argument is equivalent to an empty slice.
+	CompareBytes(input []byte) int
 }
 
 type tableKey struct {
@@ -55,4 +69,31 @@ func (p *tableKey) IndexParts() []interface{} {
 
 func (p *tableKey) String() string {
 	return fmt.Sprintf("table:%v, indexKeyAndValues:%v", string(p.table), p.indexParts)
+}
+
+// SerializeToBytes follows the ordering of how the Key is persisted in database so to compare a Key call this method
+// get bytes and compare it with raw bytes stored in database
+func (p *tableKey) SerializeToBytes() []byte {
+	if len(p.indexParts) == 0 {
+		return p.table
+	}
+
+	sb := subspace.FromBytes(p.table)
+	return sb.Pack(*(*tuple.Tuple)(unsafe.Pointer(&p.indexParts)))
+}
+
+// CompareBytes compares the serialized form of keys. It returns 0 if p == input, -1 if p < input, and +1 if p > input.
+// A nil argument is equivalent to an empty slice.
+func (p *tableKey) CompareBytes(input []byte) int {
+	return bytes.Compare(p.SerializeToBytes(), input)
+}
+
+func FromBinary(table []byte, fdbKey []byte) (Key, error) {
+	sb := subspace.FromBytes(table)
+	tp, err := sb.Unpack(fdb.Key(fdbKey))
+	if err != nil {
+		return nil, err
+	}
+
+	return NewKey(table, *(*[]interface{})(unsafe.Pointer(&tp))...), nil
 }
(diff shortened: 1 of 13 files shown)


server/services/v1/query_runner.go
```diff
@@ -20,6 +20,8 @@ import (
 	"math"
 	"time"
 
+	"github.com/tigrisdata/tigris/server/config"
```
efirs, line 23: goimports

server/services/v1/query_runner.go
```diff
@@ -430,78 +459,189 @@ type StreamingQueryRunner struct {
...
+		return options, err
+	}
+	if options.table, err = runner.encoder.EncodeTableName(tenant.GetNamespace(), db, collection); err != nil {
+		return options, err
+	}
+	if options.fieldFactory, err = read.BuildFields(runner.req.GetFields()); err != nil {
+		return options, err
+	}
+	if runner.req.Options != nil && len(runner.req.Options.Offset) > 0 {
+		if options.from, err = keys.NewKeyFromFDB(options.table, runner.req.Options.Offset); err != nil {
```
efirs, line 485: We should make this generic and move FDB specifics inside KV package.
Maybe call this `keys.Deserialize`? or `keys.FromBinary`?

efirs: Also I think it's better to serialize our key and return it as an offset to the user, instead of binary KV backend key.

server/services/v1/query_runner.go
```diff
@@ -430,78 +459,189 @@ type StreamingQueryRunner struct {
...
+			return options, err
+		}
+	}
+	if filter.None(runner.req.Filter) {
+		options.noFilter = true
+	} else {
```
efirs, line 491: Combine `} else if options.ikeys, err = runner.buildKeysUsingFilter(tenant, db, collection, runner.req.Filter); err != nil {`

Also, noFilter looks redundant. `noFilter = options.iKeys == nil`

server/config/options.go
```diff
@@ -180,3 +180,7 @@ type QuotaConfig struct {
 	LimitUpdateInterval       int64 `mapstructure:"limit_update_interval" yaml:"limit_update_interval" json:"limit_update_interval"`
 	TenantSizeRefreshInterval int64 `mapstructure:"tenant_size_refresh_interval" yaml:"tenant_size_refresh_interval" json:"tenant_size_refresh_interval"`
 }
+
+func IsIndexingStoreReadEnabled() bool {
+	return DefaultConfig.Search.WriteEnabled && DefaultConfig.Search.ReadEnabled
```
efirs, line 185: I think `Enabled` and `ReadOnly` options would be easier to understand.

keys/key.go
```diff
@@ -14,7 +14,14 @@
 
 package keys
```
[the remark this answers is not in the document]
himank: then it will mix lib-key and kv-key which we need to unify first otherwise it would be confusing.

server/config/options.go
himank: So Enabled if false doesn't mean that read is disabled as well it will only disable writes?

server/transaction/manager.go
```diff
@@ -228,7 +227,13 @@ func (s *TxSession) ReadRange(ctx context.Context, lKey keys.Key, rKey keys.Key,
 		return nil, err
 	}
 
-	return s.kTx.ReadRange(ctx, lKey.Table(), kv.BuildKey(lKey.IndexParts()...), kv.BuildKey(rKey.IndexParts()...), isSnapshot)
+	if rKey != nil && lKey != nil {
+		return s.kTx.ReadRange(ctx, lKey.Table(), kv.BuildKey(lKey.IndexParts()...), kv.BuildKey(rKey.IndexParts()...), isSnapshot)
+	} else if lKey != nil {
+		return s.kTx.ReadRange(ctx, lKey.Table(), kv.BuildKey(lKey.IndexParts()...), nil, isSnapshot)
+	} else {
+		return s.kTx.ReadRange(ctx, lKey.Table(), nil, kv.BuildKey(rKey.IndexParts()...), isSnapshot)
```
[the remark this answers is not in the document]
himank: no, it's not possible.

server/services/v1/sessions.go
```diff
@@ -292,6 +349,7 @@ func (s *QuerySession) Commit(versionMgr *metadata.VersionHandler, incVersion bo
 		return err
 	}
 
+	fmt.Println("Committing")
```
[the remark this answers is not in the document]
himank: ah!

server/services/v1/row_reader.go
```diff
@@ -30,69 +31,179 @@ type Row struct {
...
 
-type RowReader interface {
-	Next(context.Context, *Row) bool
-	Err() error
+// Iterator is to iterate over a single collection.
+type Iterator interface {
+	// Next fills the next element in the iteration. Returns true if the Iterator has more element.
+	Next(*Row) bool
+	// Interrupted returns an error if iterator encounters any error.
+	Interrupted() error
```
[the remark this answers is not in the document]
himank: Either is fine, the intention is more apparent with interrupted that an error caused streaming/iteration to stop.

server/services/v1/query_runner.go
himank: Regarding first, then it means the building of the key is at two places in the keys package and in the kv package. 

For (2), how will you be serializing our key? It is an array of interfaces. The binary backend key is the serialized form of our key. To serialize our key, we again need to iterate our primary keys defined in the schema otherwise there is no easy way to do it.

server/services/v1/query_runner.go
himank: sure

server/services/v1/search_reader.go
```diff
@@ -254,78 +212,111 @@ func (p *pageReader) buildStats(stats tsApi.FacetCounts) *api.FacetStats {
...
+func NewSearchReader(ctx context.Context, store search.Store, coll *schema.DefaultCollection, query *qsearch.Query) *SearchReader {
+	return &SearchReader{
+		ctx:        ctx,
+		store:      store,
+		query:      query,
+		collection: coll,
+	}
+}
+
+func (reader *SearchReader) SinglePageIterator(collection *schema.DefaultCollection, filter *filter.WrappedFilter, pageNo int32) *FilterableSearchIterator {
```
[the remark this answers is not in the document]
himank: I want to decouple iterators with readers as I just want to have iterators with iterating logic and reader can have more functionality.

efirs (approved):

PR #429, opened 2022-08-17
feat: Support for range queries on time fields

adilansari:


lib/date/converter.go
```diff
@@ -0,0 +1,17 @@
...
+import "time"
+
+const (
+	// RFC3339 time layout
+	RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"
+)
+
+// ToUnixNano converts a RFC3339
+func ToUnixNano(dateStr string) (int64, error) {
+	t, err := time.Parse(RFC3339Nano, dateStr)
```
adilansari, line 12: What's a good way to detect date format from user input? I would prefer to add additional layouts here.

efirs: Maybe use a library? https://github.com/araddon/dateparse
Not ideal though. New dependency.

adilansari: Yeah, I guess better to stick to standard date time, else it can cause confusions and different queries can return different results.

himank: So the date time format is defined by us, which we are using for validation purposes. Let's stick to that format.

Also, this format needs to be in the schema package close to the DateTime data type so that we know that this format is what we are proposing to the end user for this data type.  This function can still stay here and you can accept the format which will be defined in the schema package.

lib/date/converter.go
```diff
@@ -0,0 +1,21 @@
...
+const (
+	// RFC3339Nano time layout
+	RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"
+)
+
+// ToUnixNano converts a RFC3339 to Unix nano seconds
+func ToUnixNano(dateStr string) (int64, error) {
+	t, err := time.Parse(RFC3339Nano, dateStr)
+	if err != nil {
+		return 0, api.Errorf(api.Code_INVALID_ARGUMENT, "Validation failed, %s is not a valid date-time", dateStr)
```
himank, line 18: the library should avoid returning API errors because the caller can choose to return some other error. So just return the error that you are seeing from here.

schema/reserved.go
```diff
@@ -39,3 +41,7 @@ func IsReservedField(name string) bool {
 
 	return false
 }
+
+func GetShadowedDateKey(key string) string {
```
himank, line 45: `ToSearchDateKey` and also add a comment for this usage.

server/services/v1/search_indexer.go
```diff
@@ -179,12 +181,29 @@ func PackSearchFields(data *internal.TableData, collection *schema.DefaultCollec
...
+			continue
+		}
+		if f.ShouldPack() {
+			switch f.DataType {
+			case schema.ArrayType:
+				if decData[key], err = jsoniter.MarshalToString(value); err != nil {
+					return nil, err
+				}
+			case schema.DateTimeType:
+				dateStr := fmt.Sprint(value)
```
himank, line 197: just do `dateStr, ok := value.(string)` otherwise, Sprint will do allocation which we don't need here.

himank (approved):

PR #439, opened 2022-08-18
fix: Added db and collection name restriction

JigarJoshi:
db and collection name

- it cannot contain any programming language keyword (from our curated list).
- it must consist of alphanumeric and `_`.
- it must begin with alphabet.
- It must have length atleast 1.

### DIFF
--- a/api/server/v1/validator.go
+++ b/api/server/v1/validator.go
@@ -14,6 +14,14 @@
 
 package api
 
+import (
+	"regexp"
+
+	"github.com/tigrisdata/tigris/util"
+)
+
+var validNamePattern = regexp.MustCompile("^[a-zA-Z]+[a-zA-Z0-9_]+$")
+
 type Validator interface {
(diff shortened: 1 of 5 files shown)


api/server/v1/validator.go
```diff
@@ -182,15 +191,20 @@ func isValidCollection(name string) error {
...
 	}
-
+	if !validNamePattern.MatchString(name) || util.LanguageKeywords.Contains(name) {
+		return Errorf(Code_INVALID_ARGUMENT, "invalid collection name")
+	}
 	return nil
 }
 
 func isValidDatabase(name string) error {
+	fmt.Println("db:" + name)
```
efirs, line 201: remove

JigarJoshi: removed

efirs (approved):

himank (approved):

PR #441, opened 2022-08-20
feat: Added APIs for app based authentication/authorization

JigarJoshi:
- Tigris users will be able to create their applications. 
- Each application will have credentials generated.
- Users then can supply these application credentials to their software application for authentication and authorization while calling Tigris API.

### DIFF
--- a/server/services/v1/auth.go
+++ b/server/services/v1/auth.go
@@ -15,25 +15,35 @@
 package v1
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"io"
 	"net/http"
 	"net/url"
+	"time"
 
+	"github.com/auth0/go-auth0/management"
 	"github.com/fullstorydev/grpchan/inprocgrpc"
 	"github.com/go-chi/chi/v5"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
 	"github.com/rs/zerolog/log"
 	api "github.com/tigrisdata/tigris/api/server/v1"
+	"github.com/tigrisdata/tigris/lib/date"
 	"github.com/tigrisdata/tigris/server/config"
+	"github.com/tigrisdata/tigris/server/request"
 	"google.golang.org/grpc"
 )
 
 const (
-	oauthPattern = "/oauth/*"
-	refreshToken = "refresh_token"
-	scope        = "offline_access openid"
+	authPattern       = "/auth/*"
+	refreshToken      = "refresh_token"
+	scope             = "offline_access openid"
+	createdBy         = "created_by"
+	createdAt         = "created_at"
+	tigrisNamespace   = "tigris_namespace"
+	clientCredentials = "client_credentials"
+	auth0             = "auth0"
 )
 
 type authService struct {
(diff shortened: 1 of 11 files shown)


ovaistariq (approved):

server/services/v1/auth.go
```diff
@@ -116,3 +283,91 @@ func (h *authService) RegisterGRPC(grpc *grpc.Server) error {
...
+	log.Error().Msgf("Auth0 response status code=%d", resp.StatusCode)
+	return nil, api.Errorf(api.Code_INTERNAL, "Failed to get access token: reason = %s", bodyStr)
+}
+
+func deleteApplication(a *Auth0, applicationId string) error {
+	return a.management.Client.Delete(applicationId)
+}
+
+// helper to read created_at metadata from client application
+func readCreatedAt(c *management.Client) int64 {
```
himank, line 365: Use a standard date-time format(similar to what we have in schema package) and then you can use "ToUnixNano" in the lib package.

server/services/v1/auth.go
```diff
@@ -42,53 +52,194 @@ type authService struct {
...
+	metadata[tigrisNamespace] = currentNamespace
+
+	c := &management.Client{
+		Name:           &req.Name,
+		Description:    &req.Description,
+		AppType:        &nonInteractiveApp,
+		ClientMetadata: metadata,
+	}
+
+	err = a.management.Client.Create(c)
```
himank, line 100: What happens if it is already created? i.e. leftover from the last time. The case when the requests didn't reach the cleanup flow. Do they return some error that you can use to identify that case and then we don't need to return from here but try to grant again.

server/services/v1/auth.go
```diff
@@ -42,53 +52,194 @@ type authService struct {
...
+		CreatedAt:   readCreatedAt(c),
+	}
+	return &api.CreateApplicationResponse{
+		CreatedApplication: createdApp,
+	}, nil
+}
+
+func (a *Auth0) DeleteApplication(ctx context.Context, req *api.DeleteApplicationsRequest) (*api.DeleteApplicationResponse, error) {
+	// validate
+	client, err := a.management.Client.Read(req.GetId())
```
himank, line 138: Is there a way to know from the error that the application doesn't exist? In that case, we don't need to return an internal error.

server/services/v1/auth.go
```diff
@@ -116,3 +283,91 @@ func (h *authService) RegisterGRPC(grpc *grpc.Server) error {
 	api.RegisterAuthServer(grpc, h)
 	return nil
 }
+
+func getCurrentSub(ctx context.Context) (string, error) {
+	// further filter for this particular user
+	token, err := request.GetAccessToken(ctx)
+	if err != nil {
+		return "", api.Errorf(api.Code_INTERNAL, "Failed to list applications: reason = %s", err.Error())
```
himank, line 291: Error string needs to be updated here.

server/services/v1/auth.go
```diff
@@ -116,3 +283,91 @@ func (h *authService) RegisterGRPC(grpc *grpc.Server) error {
...
+		getAccessTokenResponse := api.GetAccessTokenResponse{}
+
+		err = json.Unmarshal([]byte(bodyStr), &getAccessTokenResponse)
+		if err != nil {
+			return nil, api.Errorf(api.Code_INTERNAL, "Failed to parse external response: reason = %s", err.Error())
+		}
+		return &getAccessTokenResponse, nil
+	}
+	log.Error().Msgf("Auth0 response status code=%d", resp.StatusCode)
+	return nil, api.Errorf(api.Code_INTERNAL, "Failed to get access token: reason = %s", bodyStr)
```
himank, line 323: I think we need to make sure we are returning the correct error code that we have received in the response otherwise everything will be an internal error. But you can fix it iteratively.

server/services/v1/auth.go
```diff
@@ -116,3 +283,91 @@ func (h *authService) RegisterGRPC(grpc *grpc.Server) error {
...
```
himank, line 357: same as above.

server/services/v1/auth.go
```diff
@@ -116,3 +283,91 @@ func (h *authService) RegisterGRPC(grpc *grpc.Server) error {
...
+}
+
+func deleteApplication(a *Auth0, applicationId string) error {
+	return a.management.Client.Delete(applicationId)
```
himank, line 361: not seeing much benefit of having a helper here, just call `Delete` directly.

server/services/v1/auth.go
```diff
@@ -42,53 +52,194 @@ type authService struct {
...
+	}
+
+	currentSub, err := getCurrentSub(ctx)
+	if err != nil {
+		return nil, api.Errorf(api.Code_INTERNAL, "Failed to list applications: reason = %s", err.Error())
+	}
+
+	var apps []*api.Application
+	for _, client := range appList.Clients {
+		// filter for this user's apps
```
himank, line 208: Can a user have more than one app?

himank: Also, if it is for a single client then is it possible to issue a `list` on that client?

server/services/v1/auth.go
JigarJoshi: there is no way to detect if this is the same application being created again because id is auto-generated. We can improve this later and limit the unique name per account/org. wdyt?

server/services/v1/auth.go
JigarJoshi: I have reached out to auth0 on how to filter in this query for the metadata attribute. I am waiting for them.

server/services/v1/auth.go
JigarJoshi: right I had this more than one statement before and then trimmed and it was left to cleanup. I will refactor it.

server/services/v1/auth.go
himank: sg

server/services/v1/auth.go
himank: Are there any other metadata attached to it? We can use that to read first.

server/services/v1/auth.go
JigarJoshi: updated to use standard string date format and ToUnixNano.

server/services/v1/auth.go
JigarJoshi: fixed.

server/services/v1/auth.go
JigarJoshi: updated the error handling to convey the closest status code back.

server/services/v1/auth.go
JigarJoshi: updated the error handling to convey the closest status code back.

server/services/v1/auth.go
JigarJoshi: updated the error handling to convey the closest status code back.

server/services/v1/auth.go
JigarJoshi: cleanedup.

himank (approved):

PR #448, opened 2022-08-23
feat: size metrics updated periodically

pboros:
All size metrics for all tenants updates every 5 minutes regardless of the namespace of the request it hits (the amount of time is configurable).

### DIFF
--- a/server/quota/quota.go
+++ b/server/quota/quota.go
@@ -57,6 +57,7 @@ type Manager struct {
 }
 
 var mgr Manager
+var lastGlobalRefresh int64
 
 func Init(t *metadata.TenantManager, tx *transaction.Manager, c *config.QuotaConfig) {
 	mgr = *newManager(t, tx, c)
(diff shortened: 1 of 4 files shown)


server/quota/quota.go
```diff
@@ -168,7 +227,19 @@ func (m *Manager) updateTenantMetrics(ctx context.Context, namespace string) {
 
 		s.TenantSizeUpdateAt.Store(currentTimeStamp)
 		metrics.UpdateNameSpaceSizeMetrics(namespace, sz)
-		m.updateTenantSize(ctx, namespace)
+		err := m.updateTenantSize(ctx, namespace)
```
efirs, line 230: Do not return error from updateTenantSize, so as there is no action on it and the error has already been logged in getTenantSize

server/quota/quota.go
```diff
@@ -68,12 +69,28 @@ func Allow(ctx context.Context, namespace string, reqSize int) error {
 	// Emit size metrics regardless of enabled quota
 	mgr.updateTenantMetrics(ctx, namespace)
 
+	// Update all tenant quota and size metrics periodically
+	mgr.updateAll(ctx)
```
efirs, line 73: Is it called just once? I don't see where it's called periodically.

efirs: ah, sorry. I thought it's in the Init.

server/quota/quota.go
```diff
@@ -128,14 +145,31 @@ func getCollSize(ctx context.Context, tenant *metadata.Tenant, db *metadata.Data
 	return collSize
 }
 
-func (m *Manager) updateTenantSize(ctx context.Context, namespace string) {
-	if m.txMgr == nil {
-		return
-	}
+func (m *Manager) getTenant(ctx context.Context, namespace string) *metadata.Tenant {
 	tenant, err := m.tenantMgr.GetTenant(ctx, namespace, m.txMgr)
```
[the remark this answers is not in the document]
pboros: Fixed.

server/quota/quota.go
```diff
@@ -128,14 +145,31 @@ func getCollSize(ctx context.Context, tenant *metadata.Tenant, db *metadata.Data
...
 	if err != nil {
 		ulog.E(err)
-		// Could not determine tenant, just exit
+		return nil
+	}
+	return tenant
+}
+
+func (m *Manager) getTenantSize(ctx context.Context, namespace string) (int64, error) {
+	tenant := m.getTenant(ctx, namespace)
```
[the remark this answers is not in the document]
pboros: I added the check, the size will return 0 in that case, so there is no need to return a separate error. At the time of emitting the metrics, I already check if the tenant is nil, so there will be no 0 size metrics for nil tenants.

server/quota/quota.go
```diff
@@ -128,14 +145,31 @@ func getCollSize(ctx context.Context, tenant *metadata.Tenant, db *metadata.Data
...
+	size, err := tenant.Size(ctx)
```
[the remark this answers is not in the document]
pboros: Fixed.

server/quota/quota.go
pboros: Makes sense, I changed this.

PR #452, opened 2022-08-23
feat: To support update and delete queries by filtering on any field

himank:


store/kv/fdb.go
```diff
@@ -401,12 +401,16 @@ func (t *ftx) Insert(ctx context.Context, table []byte, key Key, data []byte) er
...
 }
 
-func (t *ftx) Replace(ctx context.Context, table []byte, key Key, data []byte) error {
+func (t *ftx) Replace(ctx context.Context, table []byte, key Key, data []byte, isUpdate bool) error {
 	listener := GetEventListener(ctx)
 	k := getFDBKey(table, key)
 
 	t.tx.Set(k, data)
-	listener.OnSet(ReplaceEvent, table, k, data)
+	if isUpdate {
```
efirs, line 409: My concern is that we are changing the interface just to distinguish the type in CDC.
Can we just pass something in the context instead?

himank: Context+Metrics, let me pass in the context but then I need to extract it out at a few places.

himank: Context may be overkill because setting it means recreation of context and then attaching the associated metadata again. Better to expose options or other things for these APIs.

efirs (approved):

efirs (approved):

PR #453, opened 2022-08-24
feat: Skip commit/rollback for read-only transactions

himank:


store/kv/fdb.go
```diff
@@ -584,14 +588,26 @@ func (t *ftx) Commit(ctx context.Context) error {
...
 
 	return nil
 }
 
+func (t *ftx) End(_ context.Context) error {
+	if t.err != nil {
+		return t.err
+	}
+
+	log.Err(t.err).Msg("tx End")
```
ovaistariq, line 604: Should we log this as error?

efirs: Actually it should be already logged at the time we get in previous calls.

server/services/v1/query_runner.go
```diff
@@ -539,19 +539,19 @@ func (runner *StreamingQueryRunner) ReadOnly(ctx context.Context, tenant *metada
 		}
 
 		var last []byte
-		if last, err = runner.iterateOnKvStore(ctx, tx, options); err == nil {
-			_ = tx.Commit(ctx)
-			return &Response{}, ctx, nil
-		}
-		_ = tx.Rollback(ctx)
+		last, err = runner.iterateOnKvStore(ctx, tx, options)
+		_ = tx.End(ctx)
```
efirs, line 543: Why not just tx.Rollback() ?

efirs: It's not obvious what End does from the name.

himank: So I want to avoid calling Rollback, it has readability concerns. I am open to naming though.

store/kv/fdb.go
himank: this is some cleanup left, removing.

efirs (approved):

PR #454, opened 2022-08-24
fix: fix panic on inconsistent metric tags

pboros:
Fixes panic on inconsistent metrics tags by forcing tags on each tally scope to be always the same. 
* If a tag key is missing, it's filled up with "unknown"
* If there is an extra tag key, it's removed

Added stronger testing to ensure this no longer happens.

### DIFF
--- a/server/metrics/fdb.go
+++ b/server/metrics/fdb.go
@@ -26,17 +26,59 @@ var (
 	FdbRespTime      tally.Scope
 )
 
+func getFdbOkTagKeys() []string {
+	return []string{
+		"grpc_method",
+		"grpc_service",
+		"tigris_tenant",
+		"grpc_service_type",
+		"env",
+		"db",
+		"collection",
+		"fdb_method",
+	}
+}
+
+func getFdbTimerTagKeys() []string {
+	return []string{
+		"grpc_method",
+		"grpc_service",
+		"tigris_tenant",
+		"grpc_service_type",
+		"env",
+		"db",
+		"collection",
+		"fdb_method",
+	}
+}
+
+func getFdbErrorTagKeys() []string {
+	return []string{
+		"grpc_method",
+		"grpc_service",
+		"tigris_tenant",
+		"grpc_service_type",
+		"env",
+		"db",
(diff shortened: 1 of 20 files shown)


server/metrics/fdb.go
```diff
@@ -26,17 +26,59 @@ var (
 	FdbRespTime      tally.Scope
 )
 
+func getFdbOkTagKeys() []string {
```
rbarabas, line 29: What do you think about moving all the tag helper functions to a separate file in the same package, like `tags.go`?

server/metrics/network.go
```diff
@@ -21,17 +21,29 @@ var (
 	BytesSent     tally.Scope
 )
 
+func getNetworkTagKeys() []string {
```
rbarabas, line 24: Tags.go?

server/metrics/requests.go
```diff
@@ -36,6 +35,44 @@ var (
 	RequestsRespTime tally.Scope
 )
 
+func getRequestOkTagKeys() []string {
```
rbarabas, line 38: Tags.go?

server/metrics/search.go
```diff
@@ -26,6 +26,47 @@ var (
 	SearchRespTime      tally.Scope
 )
 
+func getSearchOkTagKeys() []string {
```
rbarabas, line 29: Tags.go! :- )

server/metrics/session.go
```diff
@@ -26,10 +26,51 @@ var (
 	SessionRespTime      tally.Scope
 )
 
+func getSessionOkTagKeys() []string {
```
rbarabas, line 29: ...

rbarabas (approved):
LGTM. Left a preference related note!

server/metrics/fdb.go
pboros: I think they belong to the metric group. When you create a new group of metric, you create one of these files, which has the initialisation methods, the scope, and the tags the metrics will have. What do you think?

PR #469, opened 2022-08-29
fix: Cache access token validation

JigarJoshi:
- Added one switch that will control if the auth middleware should be invoked or not.
- cache access token for better performance.
- validate expiration after reading from cache.
- TODO: handle key rotation

### DIFF
--- a/server/config/options.go
+++ b/server/config/options.go
@@ -40,19 +40,22 @@ type Config struct {
 }
 
 type AuthConfig struct {
-	IssuerURL                string        `mapstructure:"issuer_url" yaml:"issuer_url" json:"issuer_url"`
-	Audience                 string        `mapstructure:"audience" yaml:"audience" json:"audience"`
-	JWKSCacheTimeout         time.Duration `mapstructure:"jwks_cache_timeout" yaml:"jwks_cache_timeout" json:"jwks_cache_timeout"`
-	LogOnly                  bool          `mapstructure:"log_only" yaml:"log_only" json:"log_only"`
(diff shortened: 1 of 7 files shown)


server/config/options.go
```diff
@@ -118,6 +121,7 @@ var DefaultConfig = Config{
 		FDBHardDrop: false,
 	},
 	Auth: AuthConfig{
+		EnableAuthCheck:  false,
```
efirs, line 124: Can this be just Enabled?

efirs: to be consistent with other packages configs.

server/middleware/auth.go
```diff
@@ -138,12 +139,26 @@ func AuthFunction(ctx context.Context, jwtValidator *validator.Validator, config
...
+		validatedToken, err = jwtValidator.ValidateToken(ctx, tkn)
+		if err != nil {
+			log.Warn().Err(err).Msg("Failed to validate access token")
+			return ctx, api.Errorf(api.Code_UNAUTHENTICATED, "Failed to validate access token")
+		}
+		cache.Add(tkn, validatedToken)
+	} else {
+		// validate expiration
+		validatedClaims, ok := validatedToken.(*validator.ValidatedClaims)
+		if ok {
```
efirs, line 153: This check can be moved inside below `if`.

efirs: so the else branch is not needed here.

server/config/options.go
JigarJoshi: updated. thanks

server/middleware/auth.go
JigarJoshi: updated. thanks

efirs (approved):

PR #470, opened 2022-08-29
feat: added first pass implementation of topic partitioning

mattayres:


server/metadata/key_encoder.go
```diff
@@ -63,6 +65,18 @@ func (d *DictKeyEncoder) EncodeIndexName(idx *schema.Index) []byte {
...
 }
 
+func (d *DictKeyEncoder) EncodeKeyWithPartition(encodedTable []byte, idx *schema.Index, idxValue interface{}, partition uint16) (keys.Key, error) {
+	var allParts []interface{}
+	allParts = append(allParts, encoding.UInt16ToByte(partition))
+	allParts = append(allParts, idxValue)
+	return d.EncodeKey(encodedTable, idx, allParts)
+}
+
+func (d *DictKeyEncoder) DecodeKeyWithPartition(key keys.Key) (interface{}, uint16) {
```
himank, line 75: don't we need validation if the key is the topic key?

server/services/v1/query_runner.go
```diff
@@ -962,13 +978,78 @@ func (runner *SearchQueryRunner) getSortOrdering(coll *schema.DefaultCollection)
...
+	var err error
+	var ts = internal.NewTimestamp()
+	var allKeys [][]byte
+	for _, doc := range messages {
+		err = runner.validate(coll, doc)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		table, err := runner.encoder.EncodeTableName(tenant.GetNamespace(), db, coll)
```
himank, line 1022: We need to encode `topic` as well in the table name key and then similarly in the decode we need to check if the first 5bytes is `topic` so that we don't decode other keys.

schema/schema.go
```diff
@@ -105,7 +105,8 @@ func Build(collection string, reqSchema jsoniter.RawMessage) (*Factory, error) {
 func BuildWithType(collection string, reqSchema jsoniter.RawMessage, cType api.CollectionType) (*Factory, error) {
 	var err error
 	if cType == api.CollectionType_MESSAGES {
-		if reqSchema, err = setPrimaryKey(reqSchema, jsonSpecFormatDateTime, false); err != nil {
+		// TODO: need to specify better properties for primary key
+		if reqSchema, err = setPrimaryKey(reqSchema, "", false); err != nil {
```
himank, line 109: do we need to call this?

server/metadata/key_encoder.go
himank, line 75: Can you also document this API that this is returning index value and partition?

server/services/v1/query_runner.go
```diff
@@ -962,13 +978,78 @@ func (runner *SearchQueryRunner) getSortOrdering(coll *schema.DefaultCollection)
...
+		if err != nil {
+			return nil, nil, err
+		}
+
+		partition := rand.Intn(partitions)
```
himank, line 1027: Not for now, but this can be moved in some common sharding package/class and have the following:

- GetAnyPartition(totalPartition int)
- GetPartition(key string, totalPartition int)

server/metadata/key_encoder.go
```diff
@@ -56,13 +59,34 @@ type DictKeyEncoder struct {
...
+	}
+
+	var allParts []interface{}
+	allParts = append(allParts, encoding.UInt16ToByte(partition))
+	allParts = append(allParts, idxParts...)
+	return d.EncodeKey(encodedTable, idx, allParts)
+}
+
+// DecodePartitionKey returns index parts and partition number
+func (d *DictKeyEncoder) DecodePartitionKey(key keys.Key) ([]interface{}, uint16) {
```
himank, line 85: Decode needs to validate the first 4 bytes as well

himank (approved):

himank (approved):

himank (approved):

himank (approved):

PR #484, opened 2022-09-05
feat: added partition options for pubsub

mattayres:


server/services/v1/query_runner.go
```diff
@@ -1071,14 +1080,26 @@ func (runner *SubscribeQueryRunner) Run(ctx context.Context, tx transaction.Tx,
...
 		skipFirst bool
 		startTime time.Time
 	}
 
-	var parts [partitions]Part
+	parts := make([]Part, len(partNums))
 
-	for i := 0; i < len(parts); i++ {
+	for i := 0; i < len(partNums); i++ {
+		parts[i].num = uint16(partNums[i])
```
himank, line 1102: Do we need to add a check here if part is valid?

himank (dismissed):

server/services/v1/query_runner.go
mattayres: Yes, good catch.

himank (approved):

PR #485, opened 2022-09-06
refactor: tracing for the whole request, including auth

pboros:


api/openapi.yaml
```diff
@@ -0,0 +1,1519 @@
+# Generated with protoc-gen-openapi
```
efirs, line 1: Is it checked in by mistake?

pboros: yes

server/middleware/middleware.go
```diff
@@ -78,21 +78,20 @@ func Get(config *config.Config, tenantMgr *metadata.TenantManager, txMgr *transa
...
+	if config.Tracing.Enabled {
+		streamInterceptors = append(streamInterceptors, traceStream())
 	}
 
-	streamInterceptors = append(streamInterceptors, namespaceInitializer.NamespaceSetterStreamServerInterceptor())
+	streamInterceptors = append(streamInterceptors, forwarderStreamServerInterceptor())
 
-	if config.Tracing.Enabled {
-		streamInterceptors = append(streamInterceptors, traceStream())
+	if config.Auth.Enabled {
```
rbarabas, line 89: Should we make tracing granularly configurable? Such as, make it feasible to turn ot on/off for other components in case there's measurable or significant overhead?

pboros: We eventually should, in a separate pull request. This is just refactoring and making auth part of it.

pboros: But even if it was more granuarly configurable, loading the top level interceptor would be needed if there is any kind of tracing enabled.

rbarabas (approved):

PR #491, opened 2022-09-08
fix: Add checks based on collection type

himank:


server/services/v1/query_runner.go
```diff
@@ -307,6 +307,9 @@ func (runner *InsertQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 	if err != nil {
 		return nil, ctx, err
 	}
+	if coll.Type() != api.CollectionType_DOCUMENTS {
```
efirs, line 310: Let' create a function for it?
if err := checkCollType(api.CollectionType_DOCUMENTS); err != nil {
   return err
}
or at least define error variable so we don't duplicate it message.

himank: I'll do first, easy to return the custom message with the method name.

ovaistariq (dismissed):

server/services/v1/query_runner.go
```diff
@@ -341,6 +360,9 @@ func (runner *ReplaceQueryRunner) Run(ctx context.Context, tx transaction.Tx, te
 	if err != nil {
 		return nil, ctx, err
 	}
+	if err = runner.mustBeDocumentsCollection(coll, "replace"); err != nil {
```
efirs, line 363: Can we actually do this in validator?

himank: No, collection object is created in the service layer. Validator is only meant for request validation.

server/services/v1/api.go
```diff
@@ -16,6 +16,9 @@ package v1
 
 import (
 	"context"
+	"fmt"
```
efirs, line 19: goimports

efirs (approved):

PR #507, opened 2022-09-12
refactor: Merge multi-search results in a sorted order

adilansari:
- Merging search hits from multiple search backends in a sorted order before assembling search results
- This only impacts when search requests contain `OR` queries and search backend is queried multiple times for results


server/search/hits.go
```diff
@@ -0,0 +1,168 @@
...
+
+package search
+
+import (
+	"github.com/tigrisdata/tigris/query/sort"
+	tsApi "github.com/typesense/typesense-go/typesense/api"
+	"github.com/tigrisdata/tigris/lib/container"
+)
+
+type Hit struct {
```
adilansari, line 23: internal structure for a search hit, used only for immediate ordered queue

lib/container/priority_queue.go
```diff
@@ -0,0 +1,96 @@
...
+	return pq.comparator(pq.data[i], pq.data[j])
+}
+
+func (pq queue[T]) Swap(i, j int) {
+	pq.data[i], pq.data[j] = pq.data[j], pq.data[i]
+}
+
+func (pq *queue[T]) Push(x any) {
+	item := x.(*T)
+	(*pq).data = append((*pq).data, item)
```
efirs, line 86: Why `pq` dereference is needed?
Can it be just `pq.data = append(pq.data, item)`?

lib/container/priority_queue.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// queue is the internal data structure used to satisfy heap.Interface and not
+// supposed to be used directly. Use PriorityQueue instead.
+type queue[T any] struct {
+	data       []*T
+	comparator func(this, that *T) bool
+}
+
+func (pq queue[T]) Len() int {
```
efirs, line 72: Maybe `q` would be better receiver here, to distinquish from PriorityQueue's one.

lib/container/priority_queue.go
adilansari: ah right

lib/container/priority_queue.go
```diff
@@ -0,0 +1,96 @@
...
+	q.data[i], q.data[j] = q.data[j], q.data[i]
+}
+
+func (q *queue[T]) Push(x any) {
+	item := x.(*T)
+	q.data = append(q.data, item)
+}
+
+func (q *queue[T]) Pop() any {
+	old := (*q).data
```
efirs, line 90: here too, no need to dereference explicitly?

server/search/hits.go
```diff
@@ -0,0 +1,168 @@
...
+const (
+	This  = 1
+	That  = -1
+	Equal = 0
+)
+
+// hitsComparator returns 0 if priority of this and that are equal
+// return positive int if priority of "this" is greater than "that"
+// return negative int if priority of "that" is greater than "this"
+func hitsComparator(this, that *Hit, sortingOrder *sort.Ordering) int {
```
efirs, line 96: Slice are pretty cheap to pass by value. It will improve the readability too, so just pass `sortingOrder sort.Ordering`

adilansari: Sorting order may never have been initialized though

efirs (approved):

PR #510, opened 2022-09-12
feat: Allowing object data types to not have a strict schema

himank:


adilansari:
This will still index data on search backend for nested fields, but users won't be able to query against nested fields as of now. Should we rather persist this schemaless object as a string on disk and save on memory?

PR #511, opened 2022-09-12
fix: unauthenticated namespace available regardless of tracing

pboros:
* On panic, the full stack trace can to to the error tag, cutting that after the first newline
* The SendMsg and ReceiveMsg are present on the streams regardless if the interceptors are present, so they should not expect values populated by the interceptor
* More logging in quota in case of unknown tenant 
* Move namespace name extraction before authentication to it's own interceptor, so it's not dependent on tracing

### DIFF
--- a/server/metrics/tags.go
+++ b/server/metrics/tags.go
@@ -88,9 +88,16 @@ func getTagsForError(err error, source string) map[string]string {
 			"error_value":  "none",
 		}
 	}
+
+	genericErrorValue := err.Error()
+	// Don't capture the full stack trace as a tag for a panic, just the error message
+	if strings.HasPrefix(genericErrorValue, "panic") {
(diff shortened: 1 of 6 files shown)


server/metrics/tags.go
```diff
@@ -88,9 +88,16 @@ func getTagsForError(err error, source string) map[string]string {
 			"error_value":  "none",
 		}
 	}
+
+	genericErrorValue := err.Error()
+	// Don't capture the full stack trace as a tag for a panic, just the error message
+	if strings.HasPrefix(genericErrorValue, "panic") {
+		genericErrorValue = strings.Split(genericErrorValue, "\n")[0]
```
JigarJoshi, line 95: nice

JigarJoshi (approved):

rbarabas (approved):

PR #519, opened 2022-09-13
Allow getAccessToken API without namespace set

efirs:
* Allow getAccessToken to proceed with default_namespace, when
  called with app_id/app_secret credentials instead of token.
* Return and error if namespace is empty in the token
* Set namespace to "unknown" if token is not present in the request
* Set namespace to "default_namespace" if we know that token should not
  be present or request is across namespaces
* Use default namespace if namespace isolation is disabled by config

### DIFF
--- a/server/middleware/namespace.go
+++ b/server/middleware/namespace.go
@@ -18,55 +18,54 @@ import (
 	"context"
 
 	middleware "github.com/grpc-ecosystem/go-grpc-middleware"
-	api "github.com/tigrisdata/tigris/api/server/v1"
 	"github.com/tigrisdata/tigris/lib/container"
-	"github.com/tigrisdata/tigris/server/config"
-	"github.com/tigrisdata/tigris/server/metadata"
 	"github.com/tigrisdata/tigris/server/request"
 	"google.golang.org/grpc"
 )
 
-type NamespaceSetter struct {
-	tenantManager      *metadata.TenantManager
-	namespaceExtractor request.NamespaceExtractor
-	excludedMethods    container.HashSet
-	config             *config.Config
-}
+var (
+	excludedMethods = container.NewHashSet(
+		"/HealthAPI/Health",
+		"/tigrisdata.admin.v1.Admin/createNamespace",
+		"/tigrisdata.admin.v1.Admin/listNamespaces",
+	)
 
-func (r *NamespaceSetter) NamespaceSetterUnaryServerInterceptor() func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	namespaceExtractor = &request.AccessTokenNamespaceExtractor{}
+)
+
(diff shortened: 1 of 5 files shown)


server/middleware/namespace.go
```diff
@@ -18,55 +18,54 @@ import (
...
+				if info.FullMethod != "/tigrisdata.auth.v1.Auth/getAccessToken" {
+					// We return and error when the token is set, but namespace is empty
+					return nil, err
+				}
+				namespace = request.DefaultNamespaceName
 			}
-			return handler(request.SetNamespace(ctx, namespace), req)
 		}
+
+		return handler(request.SetNamespace(ctx, namespace), req)
```
himank, line 52: I guess this is just for metric reporting?

server/middleware/namespace.go
```diff
@@ -18,55 +18,54 @@ import (
...
+func namespaceSetterStreamServerInterceptor(enabled bool) grpc.StreamServerInterceptor {
 	return func(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
-		if !r.config.Auth.EnableNamespaceIsolation {
-			wrapped := middleware.WrapServerStream(stream)
-			wrapped.WrappedContext = request.SetNamespace(stream.Context(), request.DefaultNamespaceName)
-			return handler(srv, wrapped)
-		} else {
-			namespace, err := r.namespaceExtractor.Extract(stream.Context())
-			if err != nil {
+		namespace := request.DefaultNamespaceName
```
himank, line 58: does this mean if no namespace is set it would default to `default_namespace`?

server/middleware/namespace.go
```diff
@@ -18,55 +18,54 @@ import (
...
-type NamespaceSetter struct {
-	tenantManager      *metadata.TenantManager
-	namespaceExtractor request.NamespaceExtractor
-	excludedMethods    container.HashSet
-	config             *config.Config
-}
+var (
+	excludedMethods = container.NewHashSet(
+		"/HealthAPI/Health",
+		"/tigrisdata.admin.v1.Admin/createNamespace",
```
JigarJoshi, line 29: This list should also contain getAccessToken methdo.

server/middleware/namespace.go
efirs: This is intended behavior:
* Return and error if namespace is empty in the token
* Set namespace to "unknown" if token is not present in the request
* Set namespace to "default_namespace" if we know that token should not
be present or request is across namespaces

server/middleware/namespace.go
efirs: This is for everything where namespace is needed downstream.

JigarJoshi (approved):

server/middleware/namespace.go
efirs: First two handled by auth, it doesn't matter much at this level.

JigarJoshi (dismissed):

JigarJoshi (approved):

server/middleware/namespace.go
himank: I don't think we should use "default_namespace" for it. This will be confusing because now we are using default_namespace for internal metadata, the default for the users where they don't want isolation, and across the namespaces now. Probably OK for now but let's think about it.

PR #527, opened 2022-09-14
feat: split error and ok response times in metrics

pboros:
* Split error and not error timers, like request counters
* Timers are tied to tracing spans now (like counters)
* Simplifies tagging (timers no longer need to have separate tags)

### DIFF
--- a/server/metrics/requests.go
+++ b/server/metrics/requests.go
@@ -19,24 +19,13 @@ import (
 )
 
 var (
-	OkRequests       tally.Scope
-	ErrorRequests    tally.Scope
-	RequestsRespTime tally.Scope
+	RequestsOkCount       tally.Scope
+	RequestsErrorCount    tally.Scope
+	RequestsRespTime      tally.Scope
+	RequestsErrorRespTime tally.Scope
 )
 
 func getRequestOkTagKeys() []string {
-	return []string{
-		"grpc_method",
-		"grpc_service",
-		"tigris_tenant",
-		"grpc_service_type",
-		"env",
-		"db",
-		"collection",
-	}
-}
-
-func getRequestTimerTagKeys() []string {
 	return []string{
 		"grpc_method",
 		"grpc_service",
@@ -63,7 +52,8 @@ func getRequestErrorTagKeys() []string {
 }
 
 func initializeRequestScopes() {
-	OkRequests = Requests.SubScope("count")
-	ErrorRequests = Requests.SubScope("count")
+	RequestsOkCount = Requests.SubScope("count")
+	RequestsErrorCount = Requests.SubScope("count")
 	RequestsRespTime = Requests.SubScope("response")
+	RequestsErrorRespTime = Requests.SubScope("error_response")
 }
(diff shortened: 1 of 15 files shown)


server/metrics/tracing.go
```diff
@@ -17,6 +17,9 @@ package metrics
 import (
 	"context"
 	"fmt"
+	"time"
+
+	"github.com/rs/zerolog/log"
```
efirs, line 22: goimports

server/metrics/tracing.go
```diff
@@ -212,6 +208,9 @@ func (s *SpanMeta) StartTracing(ctx context.Context, childOnly bool) context.Con
 }
 
 func (s *SpanMeta) FinishTracing(ctx context.Context) context.Context {
+	if !s.started {
+		log.Error().Str("service_name", s.serviceName).Str("resource_name", s.resourceName).Msg("Finish tracing called before starting the trace")
```
himank, line 212: don't we need to return as well from here?

server/metrics/tracing.go
```diff
@@ -225,9 +224,19 @@ func (s *SpanMeta) FinishTracing(ctx context.Context) context.Context {
...
 	}
+	s.stopped = true
+	s.stoppedAt = time.Now()
 	return ctx
 }
 
+func (s *SpanMeta) RecordDuration(scope tally.Scope, tags map[string]string) {
+	// Should be called after tracing is finished
+	if !s.started || !s.stopped {
+		log.Error().Str("service_name", s.serviceName).Str("resource_name", s.resourceName).Msg("RecordDuration was called on a span that was not started ot stopped")
```
himank, line 235: don't we need to return as well?

server/metrics/requests.go
```diff
@@ -33,18 +34,10 @@ func getRequestOkTagKeys() []string {
...
-func getRequestTimerTagKeys() []string {
-	return []string{
-		"grpc_method",
-		"grpc_service",
-		"tigris_tenant",
-		"grpc_service_type",
-		"env",
-		"db",
-		"collection",
+		"read_type",
```
himank, line 37: how are these getting used?

server/metrics/tracing.go
pboros: I was thinking about this if we should return, panic, or just log in here. I ended up with logging to see how it goes. If this is called without a start tracing, it is a programming error.

server/metrics/requests.go
pboros: These are not used yet, will be used by the query runner instrumentation, I will put these in that pr.

pboros: I removed them for now.

server/metrics/tracing.go
pboros: Added the return.

server/metrics/tracing.go
pboros: Added the return here as well.

himank (dismissed):

himank (approved):

PR #532, opened 2022-09-16
fix: Adding schema version in the payload, and maxLength check

himank:


JigarJoshi (approved):

schema/collection.go
```diff
@@ -39,7 +39,7 @@ type DefaultCollection struct {
 	// Id is the dictionary encoded value for this collection.
 	Id uint32
 	// SchVer returns the schema version
-	SchVer int
+	SchVer int32
```
JigarJoshi, line 42: nit: since the version is always unsigned - use uint32 instead?

himank: avoiding the translation from uint32 --> int32 because in the payload we are storing it as int32.

PR #566, opened 2022-09-30
Coordinated quota

efirs:
Quota package made composable and broken down into:                              
  * Node quota (total limits one node can handle)                                
  * Storage quota (maximum namespace size quota)                                 
  * Per cluster namespace quota, with Datadog backend                            
                                                                                 
Storage quota, size metrics and namespace rate updates                           
moved to background threads.                                                     
                                                                                 
Observability code is generalized and Datadog backend                            
is used by both O11y and Quota.                                                  
                                                                                 
Unrelated tests fixes


server/config/options.go
```diff
@@ -157,17 +162,42 @@ var DefaultConfig = Config{
...
-		ReadThroughputLimit:       10000000,    // bytes per second
-		DataSizeLimit:             10000000000, // bytes
-		LimitUpdateInterval:       5,           // seconds
-		TenantSizeRefreshInterval: 60,          // seconds
-		AllTenantsRefreshInternal: 300,         // seconds
+		Node: LimitsConfig{
+			Enabled: false,
+
+			ReadRateLimit:        1000,     // read requests per second per namespace
+			ReadThroughputLimit:  10000000, // read bytes per second
```
efirs, line 169: If Throughput is not set then it deduces from rate and unit size.

api/server/v1/tx.go
```diff
@@ -21,24 +21,70 @@ import (
...
+		return true
+	case describeCollection, describeDatabase:
+		return true
+	default:
+		return false
+	}
+}
+
+func isWrite(name string) bool {
+	return !isRead(name)
```
JigarJoshi, line 70: we classified requests in three categories for metrics query api. read, write and metadata. https://github.com/tigrisdata/tigris/blob/main/server/services/v1/observability.go#L213 from quota perspective what you did makes sense but just bringing it up here if we should restrict the usage of N number of db creation for example.

api/server/v1/tx.go
```diff
@@ -21,24 +21,70 @@ import (
...
+	case listCollections, listDatabases:
+		return true
+	case describeCollection, describeDatabase:
+		return true
+	default:
+		return false
+	}
+}
+
+func isWrite(name string) bool {
```
JigarJoshi, line 69: probably we can move these methods in `server/request/request.go` in `RequestMetadata`. wdyt?

api/server/v1/tx.go
ovaistariq: Yeah we don't count list database, list collection, etc as read queries when displaying in the web console

server/quota/node.go
```diff
@@ -0,0 +1,57 @@
...
+// This limiter protects the node from overloading,
+// It limits the total number of request and throughput
+// handled by the node, by all the namespaces.
+
+type node struct {
+	state *State
+	cfg   *config.QuotaConfig
+}
+
+func (c *node) Allow(_ context.Context, namespace string, size int, isWrite bool) (err error) {
```
himank, line 19: not seeing much benefit of named return value here, same applies to below method.

server/quota/namespace.go
```diff
@@ -0,0 +1,225 @@
...
+	"go.uber.org/atomic"
+	"golang.org/x/time/rate"
+)
+
+const (
+	rateHysteresis       = 10         // ±10 rps per instance wouldn't cause rate regulation
+	throughputHysteresis = 100 * 1024 // ±100Kb per instance wouldn't cause throughput regulation
+
+	rateIncrement       = 5     // when rate adjustment is needed increment current rate by ± this value
+	throughputIncrement = 10000 // when throughput adjustment is needed increment by ± this value
```
himank, line 20: don't we need to adjust it by (+-)10% of the rate?

server/quota/quota.go
```diff
@@ -17,35 +17,32 @@ package quota
...
-	"github.com/tigrisdata/tigris/server/transaction"
 	"go.uber.org/atomic"
-	"golang.org/x/time/rate"
 )
 
 var (
-	ErrRateExceeded        = errors.ResourceExhausted("request rate limit exceeded")
-	ErrThroughputExceeded  = errors.ResourceExhausted("request throughput limit exceeded")
-	ErrStorageSizeExceeded = errors.ResourceExhausted("data size limit exceeded")
+	ErrReadRateExceeded        = errors.ResourceExhausted("request read rate limit exceeded")
```
himank, line 30: why not just use `api.Errorf(api.Code_ResourceExhausted...)`?

himank:
Overall very cool!

server/config/options.go
```diff
@@ -157,17 +162,42 @@ var DefaultConfig = Config{
...
+		Storage: StorageLimitsConfig{
+			Enabled:         false,
+			DataSizeLimit:   100 * 1024 * 1024,
+			RefreshInternal: 60 * time.Second,
+		},
 	},
 	Observability: ObservabilityConfig{
-		Enabled: false,
+		Provider:    "datadog",
+		ProviderUrl: "https://us3.datadoghq.com",
```
pboros, line 199: Should this be default configuration?

api/server/v1/tx.go
efirs: Right, there is discrepancy, but we need to enforce quota even for metadata requests, so we classify them as reads and writes here.

server/config/options.go
```diff
@@ -157,17 +162,42 @@ var DefaultConfig = Config{
...
-		WriteThroughputLimit:      10000000,    // bytes per second
-		ReadThroughputLimit:       10000000,    // bytes per second
-		DataSizeLimit:             10000000000, // bytes
-		LimitUpdateInterval:       5,           // seconds
-		TenantSizeRefreshInterval: 60,          // seconds
-		AllTenantsRefreshInternal: 300,         // seconds
+		Node: LimitsConfig{
+			Enabled: false,
+
+			ReadRateLimit:        1000,     // read requests per second per namespace
```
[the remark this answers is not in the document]
efirs: This is per node config. I'll adjust it after testing.

server/config/options.go
```diff
@@ -187,17 +217,61 @@ type SearchConfig struct {
...
+func (l *LimitsConfig) ReadThroughput() int64 {
+	if l.ReadThroughputLimit != 0 {
+		return l.ReadThroughputLimit
+	}
+	return int64(ReadUnitSize) * int64(l.ReadRateLimit)
+}
+
+type NamespaceLimitsConfig struct {
+	Enabled         bool
+	Default         LimitsConfig            // default per namespace limit
```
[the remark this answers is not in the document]
efirs: Default is default per namespace per cluster config.
Node is maximum per namespace per node config, single namespace cannot use more than this on one node.

server/quota/quota.go
efirs: `errors.ResourceExhausted(` is wrapper of `api.Errorf(api.Code_ResourceExhausted...`

server/config/options.go
efirs: > Should this be default configuration?

Do you mean have empty default instead?

> Maybe if there is no observability configured, only do local limits.

Good point. Changed it that way.

server/metrics/datadog.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	if hResp.StatusCode == http.StatusTooManyRequests {
+		log.Warn().Str(rateLimitLimit, hResp.Header.Get(rateLimitLimit)).
+			Str(rateLimitPeriod, hResp.Header.Get(rateLimitPeriod)).
+			Str(rateLimitRemaining, hResp.Header.Get(rateLimitRemaining)).
+			Str(rateLimitReset, hResp.Header.Get(rateLimitReset)).
+			Str(rateLimitName, hResp.Header.Get(rateLimitName)).
+			Msgf("Datadog rate-limit hit")
+		return nil, errors.ResourceExhausted("Failed to get query metrics: reason = rate-limited, reason = %s", resp.GetError())
```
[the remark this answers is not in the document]
efirs: That what `errors.ResourceExhausted(` returns.

server/config/options.go
```diff
@@ -157,17 +162,42 @@ var DefaultConfig = Config{
...
 	Quota: QuotaConfig{
-		Enabled:                   false,
-		RateLimit:                 1000,        // requests per second both reads and writes
-		WriteThroughputLimit:      10000000,    // bytes per second
-		ReadThroughputLimit:       10000000,    // bytes per second
-		DataSizeLimit:             10000000000, // bytes
-		LimitUpdateInterval:       5,           // seconds
-		TenantSizeRefreshInterval: 60,          // seconds
-		AllTenantsRefreshInternal: 300,         // seconds
+		Node: LimitsConfig{
```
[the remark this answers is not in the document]
efirs: Right. outside is across all namespaces, absolute maximum node allowed to handle.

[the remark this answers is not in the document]
efirs: Inside is maximum per namespace per node.

api/server/v1/tx.go
efirs: Fixed. There are now only two categories for the quota metrics request. Check: `FormDatadogQueryNoMeta`

server/quota/namespace.go
efirs: This is good catch. Fixed.

server/config/options.go
```diff
@@ -275,17 +280,38 @@ var DefaultConfig = Config{
...
+
+			ReadUnits:  4000, // read requests per second per namespace
+			WriteUnits: 1000, // write requests per second
+		},
+		Namespace: NamespaceLimitsConfig{
+			// Per cluster limits for single namespace
+			Default: LimitsConfig{
+				Enabled: false,
+
+				ReadUnits:  400, // read requests per second per namespace
```
himank, line 295: make it 100.

server/config/options.go
```diff
@@ -275,17 +280,38 @@ var DefaultConfig = Config{
...
+				WriteUnits: 100, // write requests per second
```
himank, line 296: and write can be 25.

server/quota/namespace.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	return setNodeLimit
+}
+
+func (i *namespace) updateLimitsLow(set int64, cur int64, maxInstRate int64, maxNodeRate int) int64 {
+	return calcLimit(set, int64(maxNodeRate), cur, maxInstRate, rateHysteresis, rateIncrement)
+}
+
+func (i *namespace) updateLimits(ns string, is *instanceState, curRead int64, curWrite int64) {
```
himank, line 151: Probably have some metrics for the new limits, basically what are the new limits for a namespace.

server/quota/quota.go
```diff
@@ -50,215 +49,102 @@ type State struct {
...
-		defer s.TenantSizeLock.Unlock()
-
-		s.TenantSizeUpdateAt.Store(currentTimeStamp)
-		metrics.UpdateNameSpaceSizeMetrics(namespace, sz)
-		m.updateTenantSize(ctx, namespace)
+// Allow checks read, write rates and storage size limits for the namespace
+// and returns error if at least one of them is exceeded.
+func Allow(ctx context.Context, namespace string, size int, isWrite bool) error {
+	for _, q := range mgr.quota {
+		if err := q.Allow(ctx, namespace, size, isWrite); err != nil {
```
himank, line 130: Here best to have metrics if we are not allowing and which rate-limiter is kicking in.

himank (dismissed):

himank (approved):

PR #567, opened 2022-10-01
fix: update messages type to topic

himank:


schema/schema.go
```diff
@@ -107,8 +107,8 @@ func GetCollectionType(reqSchema jsoniter.RawMessage) (CollectionType, error) {
 		switch string(val) {
 		case "documents":
 			return DocumentsType, nil
-		case "messages":
```
ovaistariq, line 110: Do you want to keep this check for backwards compatibility?

himank: Okay, let me keep it, and will remove it in the second round.

ovaistariq (approved):

PR #571, opened 2022-10-02
feat: added support for partition key defined in schema

mattayres:


schema/collection.go
```diff
@@ -100,6 +102,11 @@ func NewDefaultCollection(name string, id uint32, schVer int, ctype CollectionTy
 
 	queryableFields := BuildQueryableFields(fields)
 
+	partitionFields, err := BuildPartitionFields(schema, fields)
```
himank, line 105: Can this be built during schema parsing similar to how we set the primary key in the fields? Basically in `deserializeProperties` method we can pass this and set this attribute in the field

schema/fields.go
```diff
@@ -477,3 +478,34 @@ func buildQueryableField(parent string, f *Field) *QueryableField {
...
+	if dt == jsonparser.Array {
+		var keys []string
+		if err := jsoniter.Unmarshal(val, &keys); err != nil {
+			return nil, err
+		}
+
+		for _, key := range keys {
+			for _, field := range fields {
+				if key == field.FieldName {
+					partitionFields = append(partitionFields, field)
```
himank, line 503: field can have an attribute partition key, similar to the primary key that we can set.

server/services/v1/query_runner.go
```diff
@@ -1203,6 +1212,19 @@ func (runner *PublishQueryRunner) publish(ctx context.Context, tx transaction.Tx
 	return ts, allKeys, err
 }
 
+func partFromFields(fields []*schema.Field, message []byte) uint16 {
+	hash := fnv.New32()
+
+	for _, field := range fields {
+		val, _, _, _ := jsonparser.Get(message, field.FieldName)
```
himank, line 1219: Do we need it here? Also, best if the field is passed from the collection. Basically, once we create the collection we already pass the fields. If these fields have the partition key attribute set then we just need to rely on that attribute.

himank (approved):

JigarJoshi (approved):

PR #579, opened 2022-10-07
Fix docker publish workflows

efirs:


.github/workflows/push-docker-image.yaml
```diff
@@ -2,17 +2,15 @@ name: publish docker image
 
 on:
   pull_request:
+    paths:
```
ovaistariq, line 5: Does it mean push will only trigger when these paths are modified?

efirs: Only changes to this paths will trigger build check on pull request. We never push on pull request.

rbarabas (approved):

PR #580, opened 2022-10-10
feat: implement unset operator for update API

himank:


query/update/operator.go
```diff
@@ -57,30 +60,52 @@ type FieldOperatorFactory struct {
...
+		output []byte = existingDoc
 		err    error
 	)
 	err = jsonparser.ObjectEach(setDoc, func(key []byte, value []byte, dataType jsonparser.ValueType, offset int) error {
 		if dataType == jsonparser.String {
 			value = []byte(fmt.Sprintf(`"%s"`, value))
 		}
-		output, err = jsonparser.Set(output, value, string(key))
+
+		keys := strings.Split(string(key), ".")
```
efirs, line 107: Should we have a helper for this?
Also, is this how we specify nested fields? "a.b.c" ?

himank: The helper will also be a single method, not sure if it would be beneficial here. Isn't `a.b.c` the current client semantics of sending the nested fields?

efirs: ` Isn't a.b.c the current client semantics of sending the nested fields?`
I just didn't notice this nested semantic in the tests, now I see it.

efirs (approved):

PR #585, opened 2022-10-11
Allow human user go 5% beyond quota limit

efirs:
* Fix inital non-default namespace limits

### DIFF
--- a/server/request/request.go
+++ b/server/request/request.go
@@ -56,12 +56,14 @@ type Metadata struct {
 	methodInfo  grpc.MethodInfo
 	// this is authenticated namespace
 	namespace string
+	IsHuman   bool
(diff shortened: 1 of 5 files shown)


server/request/request.go
```diff
@@ -199,6 +201,15 @@ func GetNamespace(ctx context.Context) (string, error) {
 	return "", ErrNamespaceNotFound
 }
 
+func IsHumanUser(ctx context.Context) bool {
```
JigarJoshi, line 204: can this be moved as method of request metadata?

himank (approved):

server/request/request.go
efirs: We will refactor this, once you finish new signup flow.

PR #587, opened 2022-10-11
Fix crash with metrics disabled

efirs:
* Disable tracing when only metrics enabled
* Some refactoring

### DIFF
--- a/server/metrics/measurement.go
+++ b/server/metrics/measurement.go
@@ -69,27 +69,8 @@ func ClearMeasurementContext(ctx context.Context) context.Context {
 }
 
 func (m *Measurement) CountOkForScope(scope tally.Scope, tags map[string]string) {
-	switch scope {
-	case AuthOkCount:
-		if config.DefaultConfig.Metrics.Auth.Enabled {
-			m.countOk(scope, tags)
-		}
-	case RequestsOkCount:
-		if config.DefaultConfig.Metrics.Requests.Enabled && config.DefaultConfig.Metrics.Requests.Counter.OkEnabled {
-			m.countOk(scope, tags)
-		}
-	case FdbOkCount:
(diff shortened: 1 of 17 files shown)


pboros (approved):

rbarabas (approved):

rbarabas (approved):
LGTM

server/metrics/measurement.go
```diff
@@ -69,27 +69,8 @@ func ClearMeasurementContext(ctx context.Context) context.Context {
...
-		}
-	case SessionOkCount:
-		if config.DefaultConfig.Metrics.Session.Enabled && config.DefaultConfig.Metrics.Session.Counter.OkEnabled {
-			m.countOk(scope, tags)
-		}
-	case SearchOkCount:
-		if config.DefaultConfig.Metrics.Search.Enabled && config.DefaultConfig.Metrics.Session.Counter.OkEnabled {
-			m.countOk(scope, tags)
-		}
+	if scope != nil {
```
himank, line 72: why do we need a scope nil check? Can we guarantee during initialization that if it is not enabled then we pass noopScope? Otherwise, if we miss this check at some place then this may happen again.

server/metrics/metrics.go
```diff
@@ -76,20 +77,20 @@ func getTimerSummaryObjectives() map[float64]float64 {
...
-		DefaultSummaryObjectives: getTimerSummaryObjectives(),
-	})
-	root, closer = tally.NewRootScope(tally.ScopeOptions{
-		Tags:           GetGlobalTags(),
-		CachedReporter: Reporter,
-		// Panics with .
-		Separator: promreporter.DefaultSeparator,
-	}, 1*time.Second)
-
 	if cfg := config.DefaultConfig.Metrics; cfg.Enabled {
```
himank, line 82: in else can we have: `root = tally.NoopScope`

PR #598, opened 2022-10-15
feat: Upgrade TypeSense to 0.24.0rcn17 for local builds

rbarabas:
Upgrades TypeSense version in local builds and turns future upgrades into a simple variable change.

This is required for multiple reasons:
- address an issue in upstream TypeSense where deletes can cause performance stalls
- enable certain features that we aim to leverage

### DIFF
--- a/scripts/install_local_docker_deps.sh
+++ b/scripts/install_local_docker_deps.sh
@@ -61,9 +61,10 @@ echo "docker:docker@127.0.0.1:4500" >/etc/foundationdb/fdb.cluster
 dpkg --force-confold --configure foundationdb-server
 rm -f "$FDB_PACKAGE_PATH"
 
-TS_PACKAGE_NAME="typesense-server-0.23.0-${ARCH}.deb"
+TS_RELEASE_VERSION="0.24.0.rcn17"
+TS_PACKAGE_NAME="typesense-server-${TS_RELEASE_VERSION}-${ARCH}.deb"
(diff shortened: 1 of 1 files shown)


himank (approved):

rbarabas:
I think I see now why we couldn't simply upgrade TypeSense for local images. 

It appears that the rcn builds don't support arm64 : https://github.com/tigrisdata/tigris/actions/runs/3255908695/jobs/5345714369

The fetch for https://dl.typesense.org/releases/0.24.0.rcn17/typesense-server-0.24.0.rcn17-arm64.deb returns a 404.

scripts/install_local_docker_deps.sh
```diff
@@ -61,9 +61,10 @@ echo "docker:docker@127.0.0.1:4500" >/etc/foundationdb/fdb.cluster
 dpkg --force-confold --configure foundationdb-server
 rm -f "$FDB_PACKAGE_PATH"
 
-TS_PACKAGE_NAME="typesense-server-0.23.0-${ARCH}.deb"
+TS_RELEASE_VERSION="0.24.0.rcn17"
+TS_PACKAGE_NAME="typesense-server-${TS_RELEASE_VERSION}-${ARCH}.deb"
```
adilansari, line 65: `test/docker/docker-compose.yml` probably needs an upgrade as well.

himank:
Updating it as part of this diff https://github.com/tigrisdata/tigris/pull/605

PR #607, opened 2022-10-20
feat: Support filtering on array of primitive types

himank:
- Filtering on primitive array types
- Primitive arrays can be nested i.e. part of objects

Syntax to query remains the same,
- For nested Arrays: `{"filter":{"obj.arr_type_str":"foo"}}`
- For arrays on top level: `{"filter":{"arr_type_str":"foo"}}`


schema/collection.go
```diff
@@ -82,11 +84,11 @@ func disableAdditionalProperties(properties map[string]*jsonschema.Schema) {
 	}
 }
 
-func NewDefaultCollection(name string, id uint32, schVer int, ctype CollectionType, fields []*Field, indexes *Indexes, schema jsoniter.RawMessage, searchCollectionName string) *DefaultCollection {
+func NewDefaultCollection(name string, id uint32, schVer int, ctype CollectionType, factory *Factory, searchCollectionName string, fieldsInSearch []tsApi.Field) *DefaultCollection {
```
pboros, line 87: Having a prefix for factory like with schFactory below would improve readability.

himank: sure, will update in my next diff.

PR #608, opened 2022-10-20
feat: endpoint to get details for all namespaces

pboros:
Get details for all namespaces in the form of:
```
{
	"ns1": {
		"db1": {
			"coll1": {
				"schema": {},
				"size": x,  
			}
		},
		"db2": {
			...
		}
	},
	"ns2": {}
}
```

### DIFF
--- a/api/server/v1/tx.go
+++ b/api/server/v1/tx.go
@@ -51,10 +51,11 @@ const (
 	DescribeDatabaseMethodName   = apiMethodPrefix + "DescribeDatabase"
 	DescribeCollectionMethodName = apiMethodPrefix + "DescribeCollection"
 
-	ObservabilityMethodPrefix = "/tigrisdata.observability.v1.Observability/"
-	ManagementMethodPrefix    = "/tigrisdata.management.v1.Management/"
-	CreateNamespaceMethodName = ManagementMethodPrefix + "CreateNamespace"
-	ListNamespaceMethodName   = ManagementMethodPrefix + "ListNamespaces"
+	ObservabilityMethodPrefix    = "/tigrisdata.observability.v1.Observability/"
+	ManagementMethodPrefix       = "/tigrisdata.management.v1.Management/"
+	CreateNamespaceMethodName    = ManagementMethodPrefix + "CreateNamespace"
+	ListNamespaceMethodName      = ManagementMethodPrefix + "ListNamespaces"
+	DescribeNamespacesMethodName = ManagementMethodPrefix + "DescribeNamespaces"
 
 	AuthMethodPrefix         = "/tigrisdata.auth.v1.Auth/"
 	GetAccessTokenMethodName = AuthMethodPrefix + "GetAccessToken"
--- a/server/services/v1/management.go
+++ b/server/services/v1/management.go
@@ -16,8 +16,12 @@ package v1
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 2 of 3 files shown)


server/services/v1/management.go
```diff
@@ -147,6 +151,71 @@ func (m *managementService) ListNamespaces(ctx context.Context, _ *api.ListNames
...
+					"size":   strconv.FormatInt(size, 10),
+				}
+			}
+		}
+	}
+	return res, nil
+}
+
+func (m *managementService) AllNamespaceDetails(ctx context.Context, _ *api.DetailedNamespaceInfoRequest) (*api.DetailedNamespaceInfoResponse, error) {
+	resMap, err := m.getNameSpaceDetails(ctx)
```
himank, line 202: if the namespace is passed in the request then just dump that namespace otherwise iterate on all.

server/services/v1/management.go
```diff
@@ -147,6 +151,71 @@ func (m *managementService) ListNamespaces(ctx context.Context, _ *api.ListNames
 	}, nil
 }
 
+func (m *managementService) getNameSpaceDetails(ctx context.Context) (map[string]map[string]map[string]map[string]string, error) {
```
JigarJoshi, line 154: mark this as an admin api.

JigarJoshi (dismissed):

server/services/v1/management.go
efirs: I just thought we could have DescribeInstance instead.
@pboros @himank what do you think?

api/server/v1/tx.go
```diff
@@ -51,10 +51,11 @@ const (
...
 
-	ObservabilityMethodPrefix = "/tigrisdata.observability.v1.Observability/"
-	ManagementMethodPrefix    = "/tigrisdata.management.v1.Management/"
-	CreateNamespaceMethodName = ManagementMethodPrefix + "CreateNamespace"
-	ListNamespaceMethodName   = ManagementMethodPrefix + "ListNamespaces"
+	ObservabilityMethodPrefix    = "/tigrisdata.observability.v1.Observability/"
+	ManagementMethodPrefix       = "/tigrisdata.management.v1.Management/"
+	CreateNamespaceMethodName    = ManagementMethodPrefix + "CreateNamespace"
+	ListNamespaceMethodName      = ManagementMethodPrefix + "ListNamespaces"
+	DescribeNamespacesMethodName = ManagementMethodPrefix + "DescribeNamespaces"
```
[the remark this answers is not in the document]
pboros: Good point on making it consistent. Should the plural be ListSomething and the singular be DescribeSomething? That would be consistent with other databases.

server/services/v1/management.go
himank: We are not using Instance anywhere it may be confusing. What about `DescribeCluster`?

api/server/v1/tx.go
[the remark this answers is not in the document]
himank: But Database and Collection are specific to a single database and collection. Here we are saying that it will dump information for all the namespaces or a single one.

himank (dismissed):

himank (approved):

server/services/v1/management.go
himank: Actually maybe `DescribeNamespaces` is fine in this context as `DescribeCluster` or `DescribeInstance` may have a lot of other information as well.

PR #613, opened 2022-10-21
fix: remove custom unmarshalling from url params for auth get request …

himank:
…token

### DIFF
--- a/api/server/v1/marshaler.go
+++ b/api/server/v1/marshaler.go
@@ -17,7 +17,6 @@ package api
 import (
 	"encoding/json"
 	"io"
-	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


api/server/v1/marshaler.go
```diff
@@ -92,33 +80,9 @@ func (c *CustomMarshaler) Marshal(v interface{}) ([]byte, error) {
...
 func (c *CustomMarshaler) Unmarshal(data []byte, v interface{}) error {
-	if _, ok := v.(*GetAccessTokenRequest); ok {
-		return unmarshalInternal(data, v)
-	}
 	return c.JSONBuiltin.Unmarshal(data, v)
 }
 
-func unmarshalInternal(data []byte, v interface{}) error {
-	if v, ok := v.(*GetAccessTokenRequest); ok {
-		values, err := url.ParseQuery(string(data))
```
JigarJoshi, line 103: The reason why we had to do this is to support `application/x-www-form-urlencoded` request payload. With this change does the following request work?

```
curl --location --request POST 'https://api.preview.tigrisdata.cloud/v1/auth/token' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'client_id=<YOUR_CLIENT_ID>' \
--data-urlencode 'client_secret=<YOUR_CLIENT_SECRET>' \
--data-urlencode 'grant_type=client_credentials'
```

himank: Then we need to remove the other unmarshaler. Both are causing confusion.

PR #617, opened 2022-10-23
feat: metrics tagged with readable tenant name

pboros:
Add tigris_tenant_name tag, which is a readable tenant name (tigris_tenant is an uuid). The tag is in the format username_uuid. 

Example:
```
size_namespace_bytes{env="test_env",service="tigris-server",tigris_tenant="dba79777-6150-466d-a9ae-6fe07afe9804",tigris_tenant_name="username_dba79777-6150-466d-a9ae-6fe07afe9804",version="d32800b"} 0
```

The patch involves some refactoring to make tenant manager available in the requests package.


server/metrics/measurement.go
```diff
@@ -96,7 +96,7 @@ func (m *Measurement) GetResourceName() string {
 	return m.resourceName
 }
 
-func (m *Measurement) GetTags() map[string]string {
+func (m *Measurement) Gags() map[string]string {
```
himank, line 99: `GetTags` sounds better :)

server/metrics/measurement.go
```diff
@@ -192,6 +192,10 @@ func (m *Measurement) RecursiveAddTags(tags map[string]string) {
 	}
 }
 
+func (m *Measurement) GetTags() map[string]string {
+	return m.tags
```
himank, line 196: Is this the same as above?

server/request/request.go
```diff
@@ -56,11 +51,17 @@ type Metadata struct {
...
-	IsHuman   bool
+	// human readable namespace name
+	namespaceName string
+	IsHuman       bool
 	// this is parsed and set early in request processing chain - before it is authenticated for observability.
 	unauthenticatedNamespaceName string
 }
 
+func Init(tm *metadata.TenantManager) {
+	tenantMgr = tm
```
himank, line 62: Instead of directly depending on the metadata package, can we have an interface in this package to access Tenant object?
```
type TenantGetter interface {
	Get(ctx context.Context, id string) (Tenant, error)
}
```

server/metrics/measurement.go
pboros: Great catch :)

server/request/request.go
pboros: Sounds great, I changed it like this.

himank (approved):

PR #648, opened 2022-11-18
feat: Maintain database metadata

JigarJoshi:
Start to manage metadata for the database/project.

### DIFF
--- a/server/metadata/database.go
+++ b/server/metadata/database.go
@@ -12,22 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package types
+package metadata
 
-import (
-	"os"
-)
+type DatabaseMetadata struct {
+	Id        uint32
+	Creator   string
+	CreatedAt int64
+}
 
-type ContentType string
-
-const (
-	JSON      ContentType = "application/json; charset=utf-8"
-	STREAMING ContentType = "application/x-json-stream; charset=utf-8"
-)
-
-var MyOrigin, _ = os.Hostname()
-
-type AccessToken struct {
-	Namespace string
-	Sub       string
+func (dm *DatabaseMetadata) SetDatabaseId(id uint32) {
+	dm.Id = id
 }
--- a/server/metadata/namespaces.go
+++ b/server/metadata/namespaces.go
@@ -16,6 +16,7 @@ package metadata
 
 import (
 	"context"
+	"encoding/json"
 
 	"github.com/rs/zerolog/log"
 	"github.com/tigrisdata/tigris/errors"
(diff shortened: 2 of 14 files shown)


server/metadata/tenant.go
```diff
@@ -611,7 +620,24 @@ func (tenant *Tenant) CreateDatabase(ctx context.Context, tx transaction.Tx, dbN
 
 	// otherwise, proceed to create the database if there are concurrent requests on different workers then one of
 	// them will fail with duplicate entry and only one will succeed.
-	_, err := tenant.metaStore.CreateDatabase(ctx, tx, dbName, tenant.namespace.Id())
+	dbId, err := tenant.metaStore.CreateDatabase(ctx, tx, dbName, tenant.namespace.Id())
+	if config.DefaultConfig.Auth.EnableNamespaceIsolation {
```
himank, line 624: do we need this check here? if accessToken is present then we store it by default?

server/metadata/tenant.go
```diff
@@ -601,7 +610,7 @@ func (tenant *Tenant) GetNamespace() Namespace {
 // metadata version once the commit is successful so reloading happens at the next call when a transaction sees a stale
 // tenant version. This applies to the reloading mechanism on all the servers. It returns "true" If the database already
 // exists, else "false" and the error.
-func (tenant *Tenant) CreateDatabase(ctx context.Context, tx transaction.Tx, dbName string) (bool, error) {
+func (tenant *Tenant) CreateDatabase(ctx context.Context, tx transaction.Tx, dbName string, accessToken *types.AccessToken) (bool, error) {
```
himank, line 613: What about passing DatabaseMetadata as a param and then we can have a helper method to build DatabaseMetadata which is called by the runner? This way if we add anything to the database metadata then we don't need to change the function signature.

server/metadata/database.go
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metadata
+
+type DatabaseMetadata struct {
+	Version   uint32
```
himank, line 18: We don't need to have a version inside it that can go inside the TableData struct.

server/metadata/namespaces.go
```diff
@@ -27,6 +28,7 @@ import (
 
 const (
 	NamespaceSubspaceName = "namespace"
+	DatabaseKey           = "db"
```
himank, line 31: Can you also add `namespace`, `db` to the reserved/restrictions list.  Also, this key is already defined in `dictionary.go`, better to use that.

server/metadata/tenant.go
JigarJoshi: if namespace isolation is enabled, access token must be not nil here.

server/metadata/tenant.go
JigarJoshi: sounds good. made relevant updates.

server/metadata/database.go
JigarJoshi: right. deleted it.

server/metadata/namespaces.go
JigarJoshi: I made some validation. please let me know if there is any place else I missed it.

server/metadata/tenant.go
JigarJoshi: this is no longer valid with metadata being created from outside.

himank (approved):

PR #651, opened 2022-11-21
feat: Tag user sub for metrics

JigarJoshi:
metrics will contain the user `sub` tag.


server/middleware/measure.go
```diff
@@ -118,9 +121,16 @@ func (w *wrappedStream) RecvMsg(m interface{}) error {
 		err := w.ServerStream.RecvMsg(m)
 		return err
 	}
+
 	childMeasurement := metrics.NewMeasurement(TigrisStreamSpan, "RecvMsg", metrics.GrpcSpanType, parentMeasurement.GetRequestOkTags())
+	reqMetadata, err := request.GetRequestMetadataFromContext(w.Context())
+	if err != nil {
+		childMeasurement.AddTags(map[string]string{
```
pboros, line 128: This will not work. You need to add the tag https://github.com/tigrisdata/tigris/blob/main/server/metrics/requests.go#L28-L58, otherwise it will be removed at the normalisation phase.

JigarJoshi: thanks added it.

rbarabas (approved):

PR #656, opened 2022-11-23
fix: set db, collection and network tags only once for stream requests

pboros:
The db, collection and network tags were re-calculated for every document in the result of a read request unnecessarily. This PR fixes that. Yesterday's fix also broke the db and collection name reporting for reads, fixed that as well.


rbarabas:
Wonder how we could best test these

rbarabas (approved):

PR #658, opened 2022-11-24
fix: calculate db and collection tags only once for streaming requests

pboros:
Calculate db and collection tagging only once.

### DIFF
--- a/server/middleware/measure.go
+++ b/server/middleware/measure.go
@@ -17,7 +17,9 @@ package middleware
 import (
 	"context"
 
+
 	middleware "github.com/grpc-ecosystem/go-grpc-middleware/v2"
+	"github.com/rs/zerolog/log"
 	api "github.com/tigrisdata/tigris/api/server/v1"
 	"github.com/tigrisdata/tigris/errors"
 	"github.com/tigrisdata/tigris/server/metrics"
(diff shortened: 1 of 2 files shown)


server/middleware/measure.go
```diff
@@ -117,37 +117,51 @@ func measureStream() grpc.StreamServerInterceptor {
...
-	reqMetadata, err1 := request.GetRequestMetadataFromContext(w.WrappedContext)
-	if err1 != nil {
-		return errors.Internal("Could not handle stream send message")
+
+	if len(w.measurement.GetDBCollTags()) == 0 {
+		// The request is not tagged yet with db and collection, need to do it on the first message
+		db, coll := request.GetDbAndColl(m)
+		reqMetadata, err := request.GetRequestMetadataFromContext(w.WrappedContext)
+		if err != nil {
+			ulog.E(err)
```
rbarabas, line 158: Could this potentially blow up our logging if it gets triggered on `SendMsg()`?

rbarabas: Perhaps we could set the error flag on measurement to avoid logging N times?

pboros: Yes, this is for each document in the result set. One may have error, the other may not. We can also just log it at debug level or something like that.

server/middleware/measure.go
```diff
@@ -117,37 +117,51 @@ func measureStream() grpc.StreamServerInterceptor {
...
-	db, coll := request.GetDbAndColl(m)
-	reqMetadata, err := request.GetRequestMetadataFromContext(w.WrappedContext)
-	if err != nil {
-		return err
+
+	if len(w.measurement.GetDBCollTags()) == 0 {
+		// The request is not tagged yet with db and collection, need to do it on the first message
+		db, coll := request.GetDbAndColl(m)
+		reqMetadata, err := request.GetRequestMetadataFromContext(w.WrappedContext)
+		if err != nil {
```
adilansari, line 132:
```suggestion
		if ulog.E(err) {
				return err 
		}
```

server/middleware/measure.go
```diff
@@ -117,37 +117,51 @@ func measureStream() grpc.StreamServerInterceptor {
...
+			ulog.E(err)
```
himank, line 133: if err is not nil can we still safely tag it in line `135`?

server/middleware/measure.go
```diff
@@ -117,37 +117,51 @@ func measureStream() grpc.StreamServerInterceptor {
 }
 
 func (w *wrappedStream) RecvMsg(m interface{}) error {
+	err := w.ServerStream.RecvMsg(m)
+	if err != nil {
+		ulog.E(err)
```
himank, line 122: why not just return it from here?

server/middleware/measure.go
```diff
@@ -117,37 +117,51 @@ func measureStream() grpc.StreamServerInterceptor {
...
-		return errors.Internal("Could not handle stream send message")
+
+	if len(w.measurement.GetDBCollTags()) == 0 {
+		// The request is not tagged yet with db and collection, need to do it on the first message
+		db, coll := request.GetDbAndColl(m)
+		reqMetadata, err := request.GetRequestMetadataFromContext(w.WrappedContext)
+		if err != nil {
+			ulog.E(err)
+		}
+		reqMetadata.SetDb(db)
```
himank, line 160: same as above, if there is an error above is it safe to use `reqMetadata` object?

server/middleware/measure.go
pboros: Right, we can't, we need to return the error here and maybe have a log message that this request will not be measured.

server/middleware/measure.go
pboros: Thanks, good catch, but here we need to return the error or the original recv call, not the one from getting request metadata. Changing the code so it's more readable.

server/middleware/measure.go
pboros: I had trouble with that, it breaks things. I wanted to do it like the SendMsg (it returns error there), but it even breaks the tests.

server/middleware/measure.go
pboros: Right it's not, we should return the original error here.

himank (dismissed):

himank (approved):

PR #662, opened 2022-12-02
refactor! [BREAKING] Switching to Project paradigm

JigarJoshi:
This PR implements API changes https://github.com/tigrisdata/tigris-api/pull/166 & https://github.com/tigrisdata/tigris-api/pull/159.

At the high level it includes following changes.

- Project is an abstract object added. One project contains one DB.
- Application (machine credentials) are now scoped to a project.
- Event streaming has been removed for redesign and newer implementation.


api/server/v1/marshaler.go
```diff
@@ -532,12 +532,11 @@ func (x *DescribeCollectionResponse) MarshalJSON() ([]byte, error) {
 
 func (x *DescribeDatabaseResponse) MarshalJSON() ([]byte, error) {
 	resp := struct {
-		Db          string            `json:"db"`
-		Metadata    *DatabaseMetadata `json:"metadata"`
-		Collections []*collDesc       `json:"collections"`
-		Size        int64             `json:"size"`
+		Project     string           `json:"project"`
+		Metadata    *ProjectMetadata `json:"metadata"`
```
ovaistariq, line 536: This shouldn't contain the ProjectMetadata because Project is a higher level construct that would have more things than a database.

ovaistariq: approved PR 168

JigarJoshi: thanks - now https://github.com/tigrisdata/tigris-client-go/pull/193

JigarJoshi: updated the current PR.

ovaistariq (approved):

server/services/v1/api.go
```diff
@@ -40,10 +39,10 @@ import (
 )
 
 const (
-	databasePath        = "/databases"
-	databasePathPattern = databasePath + "/*"
+	projectPath        = "/projects"
+	projectPathPattern = projectPath + "/*"
 
-	collectionPath        = databasePath + "/collections"
+	collectionPath        = projectPath + "/collections"
```
himank, line 45: `collectionPath        = projectPath + "database/collections"`

JigarJoshi: that will evaluate to 

`/v1/projects/database/collections` 

I think we can just listen at `/v1/projects/*` and  `/v1/info` for api.go  - wdyt?

himank: but for collection APIs, we do want to listen to the right path, no?

JigarJoshi: synced offline and switched to

```
	collectionPath        = projectPathPattern + "/database/collections"

```

himank (approved):

himank (dismissed):

JigarJoshi:
cli-test is expected to fail. this will be taken care shortly after release.

PR #663, opened 2022-12-02
feat: database branching implementation

adilansari:
Early preview for database  branching with some pending items.

### What's included:
- Create/Drop/Describe DB can now have "branch" name string in payload
- All collection operations can have can have "branch" name string in payload
- Describe database without any "branch" option will describe "main" branch list all its branches

### Pending
- Newly created database branch needs to copy collections from original database
- Some metrics and namespaces api need to update
- Documentation update

### DIFF
--- a/server/metadata/database.go
+++ b/server/metadata/database.go
@@ -14,6 +14,8 @@
 
 package metadata
 
+import "strings"
+
 type DatabaseMetadata struct {
 	Id        uint32
 	Creator   string
@@ -22,4 +24,58 @@ type DatabaseMetadata struct {
 
 func (dm *DatabaseMetadata) SetDatabaseId(id uint32) {
 	dm.Id = id
+}
+
+const (
+	MainBranch          = "main"
+	BranchNamePrefix    = "$branch$_"
+	BranchNameSeparator = "_"
+)
+
+type DatabaseBranch struct {
+	db     string
+	branch string
+}
+
+func NewDatabaseBranch(db string, branch string) *DatabaseBranch {
+	if branch == MainBranch {
+		branch = ""
+	}
+	return &DatabaseBranch{
+		db:     db,
+		branch: branch,
+	}
+}
+
+func NewBranchFromDbName(key string) *DatabaseBranch {
+	if !strings.HasPrefix(key, BranchNamePrefix) {
+		return NewDatabaseBranch(key, "")
+	}
+	s := strings.SplitN(key, BranchNameSeparator, 3)
+	return NewDatabaseBranch(s[1], s[2])
+}
+
+func (b *DatabaseBranch) Name() string {
+	if b.IsMain() {
+		return b.Db()
+	}
+	return BranchNamePrefix + b.Db() + BranchNameSeparator + b.Branch()
+}
+
+func (b *DatabaseBranch) Db() string {
+	return b.db
+}
+
+func (b *DatabaseBranch) Branch() string {
+	if b.IsMain() {
+		return MainBranch
+	}
+	return b.branch
+}
+
+func (b *DatabaseBranch) IsMain() bool {
+	if len(b.branch) == 0 || b.branch == MainBranch {
+		return true
+	}
+	return false
 }
--- a/server/metadata/database.go
+++ b/server/metadata/database.go
@@ -28,52 +28,58 @@ func (dm *DatabaseMetadata) SetDatabaseId(id uint32) {
 
 const (
 	MainBranch          = "main"
-	BranchNamePrefix    = "$branch$_"
-	BranchNameSeparator = "_"
+	BranchNameSeparator = "_$branch$_"
 )
 
-type DatabaseBranch struct {
+// DatabaseName represents a primary database and its branch name
+type DatabaseName struct {
 	db     string
 	branch string
 }
 
-func NewDatabaseBranch(db string, branch string) *DatabaseBranch {
+func NewDatabaseNameWithBranch(db string, branch string) *DatabaseName {
 	if branch == MainBranch {
 		branch = ""
 	}
-	return &DatabaseBranch{
+	return &DatabaseName{
 		db:     db,
 		branch: branch,
 	}
 }
 
-func NewBranchFromDbName(key string) *DatabaseBranch {
-	if !strings.HasPrefix(key, BranchNamePrefix) {
-		return NewDatabaseBranch(key, "")
+func NewDatabaseName(key string) *DatabaseName {
+	s := strings.Split(key, BranchNameSeparator)
+	if len(s) < 2 {
+		return NewDatabaseNameWithBranch(s[0], "")
 	}
-	s := strings.SplitN(key, BranchNameSeparator, 3)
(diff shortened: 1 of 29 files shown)


server/metadata/database.go
```diff
@@ -23,3 +25,59 @@ type DatabaseMetadata struct {
...
 	dm.Id = id
 }
+
+const (
+	MainBranch          = "main"
+	BranchNamePrefix    = "$branch$_"
+	BranchNameSeparator = "_"
+)
+
+type DatabaseName struct {
```
adilansari, line 35: struct that will be replacing `dbName` string throughout the code.

server/metadata/database.go
```diff
@@ -23,3 +25,59 @@ type DatabaseMetadata struct {
...
+
+func NewDatabaseName(key string) *DatabaseName {
+	if !strings.HasPrefix(key, BranchNamePrefix) {
+		return NewDatabaseNameWithBranch(key, "")
+	}
+	s := strings.SplitN(key, BranchNameSeparator, 3)
+	return NewDatabaseNameWithBranch(s[1], s[2])
+}
+
+func (b *DatabaseName) Key() string {
```
adilansari, line 58: key is internal name of database.. see Unit tests for usage.. will be adding docs to it

himank, line 58: just call the method `Name`? Key would be confusing in the context of database name.

server/metadata/database.go
```diff
@@ -23,3 +25,59 @@ type DatabaseMetadata struct {
...
+		db:     db,
+		branch: branch,
+	}
+}
+
+func NewDatabaseName(key string) *DatabaseName {
+	if !strings.HasPrefix(key, BranchNamePrefix) {
+		return NewDatabaseNameWithBranch(key, "")
+	}
+	s := strings.SplitN(key, BranchNameSeparator, 3)
```
himank, line 54: If the database name starts with "_" then this will break? What if the naming convention is `$_{dbname}$branch_{branchName}` then it will be easy to extract first database name and then the branch name.

server/metadata/database.go
```diff
@@ -23,3 +25,59 @@ type DatabaseMetadata struct {
...
+
+// TODO: remove this
+func (b *DatabaseName) Branch() string {
+	if b.IsMainBranch() {
+		return MainBranch
+	}
+	return b.branch
+}
+
+func (b *DatabaseName) IsMainBranch() bool {
```
himank, line 78: Should we call it `WithBranch()` and `return len(b.branch) > 0` basically if no branch then the default case.

server/metadata/tenant.go
```diff
@@ -609,85 +609,142 @@ func (tenant *Tenant) GetNamespace() Namespace {
...
 
-	if _, ok := tenant.databases[dbName]; ok {
-		return true, nil
+	key := dbName.Key()
+
+	if dbName.IsMainBranch() {
+		if _, ok := tenant.databases[key]; ok {
+			return NewDatabaseExistsErr(key)
+		}
+	} else {
```
himank, line 622: `if dbName.WithBranch() {}`

server/metadata/tenant.go
```diff
@@ -609,85 +609,142 @@ func (tenant *Tenant) GetNamespace() Namespace {
...
-		log.Err(err).Msg("Failed to delete database metadata")
-		return false, errors.Internal("Failed to delete database metadata")
+		for _, c := range db.collections {
+			if err := tenant.dropCollection(ctx, tx, db, c.collection.Name); err != nil {
+				return true, err
+			}
+		}
+
+		// drop metadata entry
+		if err := tenant.namespaceStore.DeleteDatabaseMetadata(ctx, tx, tenant.namespace.Id(), key); err != nil {
```
himank, line 686: We only need to do it if the parent database is getting deleted?

server/metadata/tenant.go
```diff
@@ -609,85 +609,142 @@ func (tenant *Tenant) GetNamespace() Namespace {
...
 	}
 
 	// otherwise, proceed to create the database if there are concurrent requests on different workers then one of
 	// them will fail with duplicate entry and only one will succeed.
-	dbId, err := tenant.metaStore.CreateDatabase(ctx, tx, dbName, tenant.namespace.Id())
+	dbId, err := tenant.metaStore.CreateDatabase(ctx, tx, key, tenant.namespace.Id())
 	if dbMetadata != nil {
 		dbMetadata.SetDatabaseId(dbId)
-		err = tenant.namespaceStore.InsertDatabaseMetadata(ctx, tx, tenant.namespace.Id(), dbName, dbMetadata)
+		err = tenant.namespaceStore.InsertDatabaseMetadata(ctx, tx, tenant.namespace.Id(), key, dbMetadata)
```
himank, line 637: let's not store metadata for the branches? or do we have a use-case for it?

server/metadata/tenant.go
```diff
@@ -609,85 +609,142 @@ func (tenant *Tenant) GetNamespace() Namespace {
...
 		if err != nil {
 			log.Err(err).Msg("Failed to insert database metadata")
-			return false, errors.Internal("Failed to setup db metadata")
+			return errors.Internal("Failed to setup db metadata")
 		}
 	}
-	return false, err
+	// TODO: copy schema if a new branch is being created
```
himank, line 643: What if instead of adding the logic in this method, we have a CreateBranch method on the tenant that do the same what you have here but also has a logic of copying collection schema, etc? Also, if we have separate APIs for Create/Delete branches then it is fine to have different methods in the tenant to handle branches.

server/services/v1/database/errors.go
```diff
@@ -0,0 +1,38 @@
...
+
+package database
+
+import (
+	apiErrors "github.com/tigrisdata/tigris/errors"
+	"github.com/tigrisdata/tigris/server/metadata"
+)
+
+// createApiError helps construct API errors from internal errors.
+func createApiError(err error) error {
```
himank, line 23: nice!

server/services/v1/database/query_runner.go
```diff
@@ -169,36 +170,46 @@ func NewBaseQueryRunner(encoder metadata.Encoder, cdcMgr *cdc.Manager, txMgr *tr
...
+	dbBranch := metadata.NewDatabaseNameWithBranch(dbName, branch)
+	db, err := tenant.GetDatabase(ctx, dbBranch)
 	if err != nil {
 		return nil, err
 	}
+	// database not found
 	if db == nil {
-		// database not found
-		return nil, errors.NotFound("database doesn't exist '%s'", dbName)
+		errMsg := fmt.Sprintf("database doesn't exist '%s'", dbBranch.Db())
```
himank, line 181: nit: this can be moved in the above errors.go as well?

server/services/v1/database/query_runner.go
```diff
@@ -169,36 +170,46 @@ func NewBaseQueryRunner(encoder metadata.Encoder, cdcMgr *cdc.Manager, txMgr *tr
...
```
himank, line 208: ^

himank (approved):

himank (approved):

PR #666, opened 2022-12-07
Beta release

JigarJoshi:


JigarJoshi:
CLI test is expected to fail because of circular dependency between server and CLI. Once server is released CLI will get released with fixes and it will come to green ground.

himank (approved):

PR #667, opened 2022-12-08
fix: Update metrics tags from db to project

JigarJoshi:
WIth project abstraction in - this PR takes care of refactoring metric tags from `db` --> `project`. For migration purpose `db` is still kept around. It will be removed next month.


JigarJoshi:
cli-test is expected to remain broken until we release newer CLI.

server/request/request.go
```diff
@@ -73,7 +73,7 @@ func NewRequestEndpointMetadata(ctx context.Context, serviceName string, methodI
 }
 
 func GetGrpcEndPointMetadataFromFullMethod(ctx context.Context, fullMethod string, methodType string, req interface{}) Metadata {
-	db, coll := GetDbAndColl(req)
+	db, coll := GetProjectAndColl(req)
```
pboros, line 76: The db remained here.

pboros:
You need to add project here as well: https://github.com/tigrisdata/tigris/blob/main/server/metrics/requests.go and all the functions that returns the tag list in that directory and db is there.

pboros (approved):

server/request/request.go
JigarJoshi: good catch. fixed it.

PR #672, opened 2022-12-10
Import API added

efirs:
API for importing user documents with schema inference and evolution. User can provide a list of autogenerated and primary key fields. If primary key is not provided and document has field 'id' ot type 'uuid' it's used as collection's autogenerated primary key.


ovaistariq (approved):

schema/inference.go
```diff
@@ -0,0 +1,334 @@
...
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"github.com/rs/zerolog/log"
+	schema "github.com/tigrisdata/tigris/schema/lang"
+)
+
+const (
+	typeInteger = "integer"
```
himank, line 32: Probably we can merge it with existing fieldType defined in this package.

PR #673, opened 2022-12-13
feat: Support for default value attribute for fields

himank:
Allowing default values for fields. The values will only be set if it is missing in the user payload. 

The following would be supported:

Functions:
- now()
- uuid()
- cuid()

Default Values:
- Int32/Int64/Double/String/Date-Time/Arrays/Fields inside Nested Object: These can have default values

CreatedAt/UpdatedAt
- Tagging a field with created at or update at. 

Therefore,
- `@default(0/true/“hello”)` maps to `"default": 0 / "default": true / "default": "hello"` on the schema field
- `@default(now()/cuid()/uuid())` maps to `"default": "now()" / "default": "cuid()" / "default": "uuid()"`
- `@default()` maps to setting default with an empty string  `"default": ""` this means Tigris defaults for primitive fields.

Then the schema will look like this,
```
"field": {
	"type": "integer",
	"default": 1
}
```

- `@createdAt` maps to "createdAt": true similar to above, on the field attributes in the schema.
- `@updatedAt` maps to "updatedAt": true
```
"field": {
	"type": "string",
        "format": "date-time",
	"createdAt": "true"
}
```


schema/default.go
```diff
@@ -0,0 +1,150 @@
...
+		if len(updatedAt) == 0 {
+			return nil
+		}
+
+		return updatedAt
+	}
+
+	switch defaulter.value {
+	case funcNowName:
+		return createdAt
```
garrensmith, line 142: There is a case on update where the `createdAt` timestamp is set to `""` which means that a `now()` would also be an empty string. I'm guessing this shouldn't happen as the `now()` should only be done on create but is there a case where we could set the incorrect time?

schema/default.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	defaulter.value = v
+
+	return defaulter, nil
+}
+
+func (defaulter *FieldDefaulter) GetValue(createdAt string, updatedAt string) interface{} {
```
garrensmith, line 125: Do we need two different timestamps here?
I kind of think this should be:

```
func (defaulter *FieldDefaulter) GetValue(timestamp string)

```

And then use that timestamp to set the default value

server/services/v1/mutator_test.go
```diff
@@ -16,14 +16,191 @@ package v1
...
+func TestMutateSetDefaults(t *testing.T) {
+	reqSchema := []byte(`{
+		"title": "t1",
+		"properties": {
+			"id": {
+				"type": "integer"
+			},
+			"double_f": {
+				"type": "number",
+				"default": 1.5
```
garrensmith, line 38: I think it is important to add a test here to make sure that a field with `now()` and the createdAt are the same value. Users are really surprised if the timestamps used in a mutation operation are not all the same if they are set in that operation.

server/services/v1/query_runner.go
```diff
@@ -223,7 +223,7 @@ func (runner *BaseQueryRunner) insertOrReplace(ctx context.Context, tx transacti
 	allKeys := make([][]byte, 0, len(documents))
 	for _, doc := range documents {
 		// reset it back to doc
-		doc, err = runner.mutateAndValidatePayload(coll, doc)
+		doc, err = runner.mutateAndValidatePayload(coll, ts.ToRFC3339(), "", doc)
```
garrensmith, line 226: As I mentioned earlier I think this shouldn't be a separate `createdAt` and `updatedAt` timestamp just one timestamp value to be used for the mutation of this document.

garrensmith (commented):
Looks good this will be a nice DX addition. I've left a few comments.

schema/default.go
himank: But are you suggesting to not have createdAt/now at all? What will happen in this case?
```
A: @createdAt
B: @default(now())
C: @updatedAt
```

Then during an update, we shouldn't update `A` and `B` as that is set during insert. Similarly, `C` shouldn't be updated during an insert and only during an update. With just timestamp, we may not able to know and still needed this if/else somewhere.

server/services/v1/mutator_test.go
himank: Yes, it will be added.

server/services/v1/query_runner.go
himank: But are you suggesting not to support @createdAt/@updatedAt and just support something like @timestamp that is updated always either during insert/update?

schema/default.go
himank: So a field can only be tagged with either one of `createdAt`/`default(now()`/`updatedAt`. Now/CreatedAt will be set during insert/replacing a document and the caller guarantees that createdAt is not empty during insert and updatedAt is not empty during an update.

ovaistariq: The caller should be able to set it to some value in which case that would take precedence, right?

ovaistariq: How about this, we only have `default(now()`? Then if you don't specify a value for the fields that are tagged they will be automatically set to the value of `now()`. However, if the user specifies a value when inserting or updating then nothing happens.

ovaistariq: Postgres and MySQL allow doing something like this:

```
CREATE TABLE note(
    note_id serial PRIMARY KEY,
    message varchar(255) NOT NULL,
    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
    updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
);
```

In this case, the value for the fields is populated during insertion time.
In the case of MySQL, you can additionally say:
```
updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
```

In this case, the value is updated whenever the row is updated.

ovaistariq: But if the user supplies a value for these columns during insert or update time, that takes precedence.

himank: So we are not setting value if the payload has a value already set by the user. That applies to all the functions/defaults/createdAt/updatedAt supported by us. Only set if it is empty in the payload.

server/services/v1/query_runner.go
himank: We discussed about it.

server/services/v1/mutator_test.go
himank: Done, added an integration test(documents_test.go) that is validating that now() and createdAt are the same values.

schema/default.go
himank: Simplified it a bit more to handle createdAt/updatedAt differently i.e. now the caller(the runner) is explicitly using either an insertMutator or updateMutator so that we only either set createdAt or updatedAt.

garrensmith (approved):
This looks good. Nice clean up with the separate paths for update and insert.

PR #680, opened 2022-12-16
fix: Delete app keys associated with project while deleting project

JigarJoshi:
This PR will delete app-keys associated with project when project is undergoing deletion.

### DIFF
--- a/server/services/v1/auth/auth0.go
+++ b/server/services/v1/auth/auth0.go
@@ -289,6 +289,29 @@ func (a *auth0) ListAppKeys(ctx context.Context, req *api.ListAppKeysRequest) (*
 	}, nil
 }
 
+func (a *auth0) DeleteAppKeys(ctx context.Context, project string) error {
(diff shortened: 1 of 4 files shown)


server/services/v1/auth/auth0.go
```diff
@@ -289,6 +289,29 @@ func (a *auth0) ListAppKeys(ctx context.Context, req *api.ListAppKeysRequest) (*
 	}, nil
 }
 
+func (a *auth0) DeleteAppKeys(ctx context.Context, project string) error {
```
himank, line 292: so the caller will call this? Or should we just call this when we receive the DeleteProject request?

himank: Nvm, ignore my comment. I missed the above call.

himank (approved):

server/services/v1/auth/auth0.go
JigarJoshi: caller will call delete-project - the handler internally calls this function to also delete the app-keys associated with that project. this is not a public facing handler.

PR #711, opened 2023-01-06
fix: GetSet implementation and Px options for Set cache operation

JigarJoshi:
## Describe your changes

- Added support for `PX` options for Set operation. 
- Added GetSet operation.

- Integration test added.

### DIFF
--- a/server/services/v1/cache/runner.go
+++ b/server/services/v1/cache/runner.go
@@ -16,7 +16,6 @@ package cache
 
 import (
 	"context"
-	"time"
 
 	"github.com/rs/zerolog/log"
 	api "github.com/tigrisdata/tigris/api/server/v1"
@@ -67,6 +66,12 @@ type SetRunner struct {
 	*BaseRunner
 
 	req *api.SetRequest
+}
+
+type GetSetRunner struct {
+	*BaseRunner
+
+	req *api.GetSetRequest
 }
 
 type GetRunner struct {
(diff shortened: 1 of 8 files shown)


server/services/v1/cache/runner.go
```diff
@@ -221,14 +220,30 @@ func (runner *SetRunner) Run(ctx context.Context, tenant *metadata.Tenant) (*Cac
...
 		NX: runner.req.GetNx(),
 		XX: runner.req.GetXx(),
-		EX: time.Duration(runner.req.GetEx()) * time.Second,
+		EX: runner.req.GetEx(),
+		PX: runner.req.GetPx(),
 	}
+
+	var oldVal *internal.CacheData
+	if runner.req.GetGet() {
+		oldVal, _ = runner.cacheStore.Get(ctx, tableName, runner.req.GetKey(), &cache.GetOptions{
```
himank, line 229: why not use `GetSet` API in store for this?

JigarJoshi: GetSet doesn't accept expiration options.

JigarJoshi: `setArgs` returns the old value but always in form of `string`.

himank: Then can we have GetSet as a separate API that doesn't accept expiration?

JigarJoshi: sure https://github.com/tigrisdata/tigris-api/pull/188

efirs:
```
As of Redis version 6.2.0, this command is regarded as deprecated.
```

JigarJoshi: updated the PR.

himank (approved):

PR #722, opened 2023-01-10
refactor: Reduce allocation on heap and leverage stack by passing as value the Req/Resp objects

himank:
## Describe your changes
No need to allocate req/resp object on the heap as the lifetime is tied to a request therefore allocation on the stack here will be fast and cheap.

### DIFF
--- a/server/services/v1/cache/runner.go
+++ b/server/services/v1/cache/runner.go
@@ -30,6 +30,16 @@ import (
 	"github.com/tigrisdata/tigris/store/cache"
 )
 
+// Runner is responsible for executing the current query and return the response.
+type Runner interface {
+	Run(ctx context.Context, tenant *metadata.Tenant) (Response, error)
+}
+
+// TxRunner is responsible for executing the current query and return the response.
+type TxRunner interface {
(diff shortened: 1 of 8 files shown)


server/services/v1/cache/runner.go
```diff
@@ -160,30 +170,30 @@ func (f *RunnerFactory) GetKeysRunner(r *api.KeysRequest, accessToken *types.Acc
...
-		return nil, ctx, err
+		return Response{}, ctx, err
 	}
-	return &CacheResponse{
+	return Response{
 		Status: database.CreatedStatus,
 	}, ctx, nil
 }
 
-func (runner *DeleteCacheRunner) Run(ctx context.Context, tx transaction.Tx, tenant *metadata.Tenant) (*CacheResponse, context.Context, error) {
```
efirs, line 178: Should we rename everything to just Runner? e.g `DeleteRunner`, etc...

efirs: ah, nvm.

efirs (approved):

PR #743, opened 2023-01-20
Secondary indexes

garrensmith:
## Describe your changes

This is a start to the secondary indexes work to get feedback on the code layout and initial design. 
Currently, secondary indexes can index a document in a transaction the document is added. 

Still to do:

- [ ] Correctly index null, undefined and ignore byte fields
- [ ] Update the index on update/delete
- [ ] Query and in-memory filtering for reads

I'm looking for feedback on the following things:

1. I've put everything in one file. I don't think this is correct and doesn't follow the layout of other functionality. So how should I break this up more? Should I have the iterator in the row_reader file? Do I need to add a IndexStore to the collection or tenant
2. I"ve created a `coll.IdxVer` which is a string. This should be used as the version for when we handle schema migrations and upgrades. This isn't correct, but I couldn't find if we have implemented this schema index versioning yet. 
3. Any other obvious things I'm missing


server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+		return nil, err
+	}
+
+	return &QueryIndexInfo{
+		rows,
+		size,
+	}, nil
+}
+
+func (q *QueryIndex) Index(ctx context.Context, tx transaction.Tx, coll *schema.DefaultCollection, doc []byte, id []interface{}) error {
```
himank, line 147: Should we refer to `id` as `primaryKey`, probably be more readable in this context that indexes are pointing to the primary key?

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+	indexKeys, size, rowCount := q.buildKVs(coll, doc, id)
+
+	rowCountKey := genRowCountKey(coll)
+	indexSizeKey := genIndexSizeKey(coll)
+
+	tx.AtomicAdd(ctx, rowCountKey, rowCount)
+	tx.AtomicAdd(ctx, indexSizeKey, size)
+
+	for _, indexKey := range indexKeys {
+		err := tx.Insert(ctx, indexKey, nil)
```
himank, line 158: Also, if the user is using insert API, then blindly replacing an index is good enough. In the case of replace/update, we also need the old row to clean up old indexes.

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+		if err != nil {
+			return err
+		}
+
+	}
+
+	return nil
+}
+
+func (q *QueryIndex) buildKVs(coll *schema.DefaultCollection, doc []byte, id []interface{}) (indexedKeys []keys.Key, sizeIncrease int64, rowCount int64) {
```
himank, line 169: I'd avoid named return parameters in cases where they can be avoided, we have seen some tricky bugs in the past because of it :)

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	var buf collate.Buffer
+	collated := collator.Key(&buf, rawValue)
+
+	return collated
+}
+
+func newKeyWithId(id []interface{}, table []byte, indexParts ...interface{}) keys.Key {
+	indexParts = append(indexParts, id...)
```
himank, line 292: It would be better to add a comment on this method that the id is added at the end which is a primary key. Better to refer to it as a primary key.

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+
+func IndexNestedField(doc []byte, topField string, pos int) []FieldIndex {
+	var indexedFields []FieldIndex
+	processor := func(key []byte, value []byte, jsonDataType jsonparser.ValueType, offset int) error {
+		fieldType := schema.ToFieldType(jsonDataType.String(), "", "")
+
+		switch fieldType {
+		case schema.ArrayType:
+			// Do not index into nested arrays
+			return nil
```
himank, line 227: Should we add some dummy for the key so that we know this nested key exists? Then we can use the index to query?

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+		case schema.ObjectType:
+			IndexNestedField(value, topField+"."+string(key), pos)
+		default:
+			fieldName := topField + "." + string(key)
+			row, err := newFieldIndex(fieldType, fieldName, value, pos)
+
+			if err != nil {
+				log.Err(err).Msgf("Failed to index field name: %s", fieldName)
+				return err
```
himank, line 236: how we will get this err out?

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+	}, nil
+}
+
+type QueryIndexInfo struct {
+	Rows int64
+	Size int64
+}
+
+func genRowCountKey(coll *schema.DefaultCollection) keys.Key {
+	return keys.NewKey(coll.EncodedName, coll.Indexes.QueryIndex.Name, "_info", "count")
```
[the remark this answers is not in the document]
garrensmith: Each collection only has one index. So from my understanding `coll.EncodedName` would be unique for each index.

server/services/v1/database/query_index.go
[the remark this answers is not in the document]
garrensmith: Yes, I aim to work on that next. I want to make sure the insert is correct first

server/services/v1/database/query_index.go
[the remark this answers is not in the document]
himank: Each field will have its own index. So a collection will have multiple indexes. Coll.EncodedName is fine, this is the collection encoded name. I was talking about `coll.Indexes.QueryIndex.Name`

server/services/v1/database/query_index.go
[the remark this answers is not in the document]
himank: I mean you can just do `tx.Replace` instead of `tx.Insert`

server/services/v1/database/query_index.go
garrensmith: Good point. I've fixed it.

server/services/v1/database/query_index.go
garrensmith: Thanks for that. I've changed it.

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+			indexedKeys = append(indexedKeys, indexKey)
+			rowCount += 1
+			sizeIncrease += int64(len(indexKey.SerializeToBytes()))
+		}
+	}
+
+	return
+}
+
+func indexField(doc []byte, fieldName string, dataType schema.FieldType, pos int, keyPath ...string) (*FieldIndex, error) {
```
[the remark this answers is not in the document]
garrensmith: Yes correct.

query/filter/key_builder.go
```diff
@@ -162,15 +181,167 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
...
+	case LT, LTE:
+		return true
+
+	default:
+		return false
+	}
+}
+
+// I think this is in the wrong place
+func min(datatype schema.FieldType, val value.Value) any {
```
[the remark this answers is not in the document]
garrensmith: Yeah nice that makes a lot of sense.

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,423 @@
...
+}
+
+func BuildQueryIndexKeys(coll *schema.DefaultCollection, queryFilters []filter.Filter) ([]keys.Key, error) {
+
+	encoder := func(indexParts ...interface{}) (keys.Key, error) {
+		return newKeyWithPrimaryKey(indexParts, coll.EncodedName, coll.Indexes.QueryIndex.Name, "kvs"), nil
+	}
+
+	buildIndexParts := func(fieldName string, datatype schema.FieldType, value interface{}) []interface{} {
+		return []interface{}{fieldName, "v1", value}
```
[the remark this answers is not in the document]
garrensmith: This is a place holder for the field version. It is now removed.

query/filter/key_builder.go
```diff
@@ -162,15 +181,167 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
...
+	case schema.DateTimeType, schema.UUIDType, schema.StringType:
+		return []byte{0}
+	case schema.BoolType:
+		if val.String() == "true" {
+			return true
+		} else {
+			return false
+		}
+	default:
+		return 0x00
```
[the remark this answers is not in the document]
garrensmith: Yes. Fixed.

schema/schema.go
```diff
@@ -108,6 +108,8 @@ const (
 	CollectionTypeF              = "collection_type"
 	IndexingSchemaVersionKey     = "indexing_version"
 	DefaultIndexingSchemaVersion = "v1"
+
+	SecondaryKeyIndexName = "qkey"
```
himank, line 112: `skey`

query/filter/key_builder.go
```diff
@@ -141,19 +155,22 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
 	}
 
 	// keys building is dependent on the filter type
-	var allKeys []keys.Key
+	var allKeys []KeyReadType
 	for _, k := range compositeKeys {
 		switch parent {
 		case AndOP:
 			var primaryKeyParts []interface{}
```
himank, line 162: Should we rename this now can call it `KeyParts`?

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,418 @@
...
+			return nil, err
+		}
+	default:
+		return nil, errors.InvalidArgument("Incorrectly created query key range")
+	}
+
+	return reader, nil
+}
+
+func BuildQueryIndexKeys(coll *schema.DefaultCollection, queryFilters []filter.Filter) ([]keys.Key, error) {
```
himank, line 77: I am thinking if we move the entire key building logic in `key_builder.go`. As we have the logic of building primary key rows in key builder then better to have building secondary index keys logic at the same place.

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,418 @@
...
+
+	eqKeyBuilder := filter.NewKeyBuilder(filter.NewStrictEqKeyComposer(encoder, buildIndexParts))
+
+	for _, field := range coll.Fields {
+		eqKeys, err := eqKeyBuilder.Build(queryFilters, []*schema.Field{field})
+		if err != nil || len(eqKeys) == 0 {
+			continue
+		}
+
+		return eqKeys[0].Keys, nil
```
himank, line 96: let's add a comment that we are returning the first one in case of equality.

value/value.go
```diff
@@ -262,6 +262,10 @@ func (s *StringValue) CompareTo(v Value) (int, error) {
 }
 
 func (s *StringValue) AsInterface() interface{} {
+	if s.Collation.IsCollationSortKey() {
+		return s.Collation.GenerateSortKey(s.Value)
```
himank, line 266: There is a problem with this approach as `GenerateSortKey` is truncating the original string value which means if we need to perform in-memory filtering/comparison then it won't work as the original value is truncated.

query/filter/key_builder.go
```diff
@@ -162,15 +179,140 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
...
 				allKeys = append(allKeys, key)
 			}
 		}
 	}
 
 	return allKeys, nil
 }
+
+const (
+	RANGE     string = "range"
```
himank, line 199: is there a use case of setting Range/FullRange in KeyReadType?

query/filter/key_builder.go
```diff
@@ -162,15 +179,140 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
...
+	FULLRANGE string = "full_range"
+	EQUAL     string = "equal"
+)
+
+// TODO: better name for this
```
himank, line 204: +1

efirs: +1 :)

query/filter/key_builder.go
garrensmith: Yes. I haven't implemented that part. But we can use that to help with query planning. A `range` should be a smaller query than a `FULLRANGE` and a `EQUAL` would be the most narrow query.

server/services/v1/database/query_index.go
garrensmith: This section needs to be fixed to support multiple equal keys and choosing `RANGE` over `FULLRANGE`

server/services/v1/database/query_index.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+
+	return nil
+}
+
+func (q *QueryIndex) buildKVs(coll *schema.DefaultCollection, doc []byte, id []interface{}) (indexedKeys []keys.Key, sizeIncrease int64, rowCount int64) {
+
+	for _, field := range coll.QueryableFields {
+
+		if schema.IsReservedField(field.Name()) {
```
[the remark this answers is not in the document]
garrensmith: This is now fixed.

server/services/v1/database/search_indexer.go
```diff
@@ -101,6 +101,10 @@ func (i *SearchIndexer) OnPostCommit(ctx context.Context, tenant *metadata.Tenan
 				return err
 			}
 
+			if tableData == nil {
```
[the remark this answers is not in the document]
garrensmith: Yes, this happens when we add an index row. That value is nil and reaches here.

value/value.go
garrensmith: I need to write a test to check this but from what I can see, the `CompareTo` is used for in-memory and the `AsInterface` is used to store to fdb. So this shouldn't be a problem.

schema/schema.go
```diff
@@ -108,6 +108,8 @@ const (
 	CollectionTypeF              = "collection_type"
 	IndexingSchemaVersionKey     = "indexing_version"
 	DefaultIndexingSchemaVersion = "v1"
+
+	SecondaryKeyIndexName = "skey"
```
efirs, line 112: Maybe just 'key' or 'index'

query/filter/key_builder.go
```diff
@@ -162,15 +179,139 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
...
+func (s *RangeKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*schema.QueryableField, parent LogicalOP) ([]KeyReadType, error) {
+	var err error
+	var foundKeys []KeyReadType
+	for _, k := range userDefinedKeys {
+		var begin, end keys.Key
+		rangeType := FULLRANGE
+		for _, sel := range selectors {
+			if k.FieldName == sel.Field.Name() && s.isRange(sel) {
+				indexParts := s.buildIndexPartsFunc(sel.Field.Name(), sel.Field.DataType, sel.Matcher.GetValue().AsInterface())
+				if s.isGreater(sel) {
```
efirs, line 242: branches looks similar, maybe generalize, or at least move to their own functions.

himank:
Let's close this? And if anything is missing, let's add an issue for it.

PR #746, opened 2023-01-25
fix: Get the branch name where custom unmarshaler are used

adilansari:
## Describe your changes
- Some APIs were not receiving the branch name

### DIFF
--- a/api/server/v1/marshaler.go
+++ b/api/server/v1/marshaler.go
@@ -136,6 +136,10 @@ func (x *ReadRequest) UnmarshalJSON(data []byte) error {
 			if err := jsoniter.Unmarshal(value, &x.Collection); err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


api/server/v1/marshaler.go
```diff
@@ -169,6 +173,8 @@ func (x *SearchRequest) UnmarshalJSON(data []byte) error {
 			v = &x.Project
 		case "collection":
 			v = &x.Collection
+		case "branch":
```
ovaistariq, line 176: SearchRequest happens on a search index, not on a branch?

adilansari: This is for the existing search API on collections. This whole SearchRequest will eventually be deprecated or evolved. But for now, we need it.

himank (approved):

PR #748, opened 2023-01-26
Support incompatible schema changes

efirs:
## Describe your changes

In this diff we build list of deltas of incompatible schema changes, from the history of the collections schema changes.

On read and update we apply all the schema deltas, which has newer version then the version of the row.

Incompatible schema changes supported:
  * field deletions
  * field type changes
  * reducing maxLength of strings and bytes

For the type changed we are doing best effort convertion. If the type is not convertible we are returning empty value for it. For the convertion rules please refer to `TestTypeConversionRules`.

Future work:
  * Currently the feature doesn't work if search is enabled. (turned out TypeSense doesn't allow field drop/add if data present in the collection)
  * Support max_length truncation for the strings
  * We can also support adding default value to the field. Populating empty fields in the existing rows with new default value. This can work for static defaults and updated_at(), created_at().

Feature can be configured by `schema.allow_incompatible = true` setting

Unit and integration tests added

TIG-749


schema/fields.go
```diff
@@ -409,14 +410,15 @@ func (f *Field) IsSorted() bool {
...
-	if f.DataType != f1.DataType {
+	if f.DataType != f1.DataType && !config.DefaultConfig.Schema.AllowIncompatible {
 		return errors.InvalidArgument("data type mismatch for field %q", f.FieldName)
 	}
 
 	if f.IsPrimaryKey() != f1.IsPrimaryKey() {
 		return errors.InvalidArgument("primary key changes are not allowed %q", f.FieldName)
 	}
 
+	// TODO: Implement handling MaxLength reduction by schema change
```
garrensmith, line 421: Do we have a ticket or something to track this so we know we need to come back to this? 
Or are you planning to work on this soon?

schema/migration.go
```diff
@@ -37,57 +46,365 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+
+	for _, v := range schema {
+		m[v.FieldName] = v
+	}
+
+	return m
+}
+
+// buildSchemaDeltaLow is a recursive helper for building schema delta.
+func buildSchemaDeltaLow(keyPath []string, first []*Field, second []*Field) []*VersionDeltaField {
```
garrensmith, line 76: What does the `Low` part mean in the name? I've seen it in a few function names and I don't know what it means?

schema/migration.go
```diff
@@ -37,57 +46,365 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+	// Just one schema Version or no incompatible schema changes
+	return []VersionDelta{{0, nil}}, nil
 }
 
-// CompatibleSchemaSince determines if there was incompatible schema change since given version.
+// CompatibleSchemaSince determines if there was incompatible schema change since given Version.
 func (d *DefaultCollection) CompatibleSchemaSince(version int32) bool {
 	// first element in the array contains last incompatible schema change
-	return version >= d.SchemaDeltas[0].version
+	return int(version) >= d.SchemaDeltas[len(d.SchemaDeltas)-1].Version
```
garrensmith, line 154: Is the comment here still correct? It doesn't seem to match the checking of the last delta in the array?

schema/migration.go
```diff
@@ -37,57 +46,365 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+			return "true"
+		}
+		return "false"
+	case ByteType:
+		if val {
+			return []byte{0x1}
+		}
+		return []byte{0x0}
+	case UUIDType, DateTimeType, ArrayType, ObjectType:
+		// non convertible
```
garrensmith, line 177: An alternative, ee could return true if the field is non null otherwise false. We would just have to document our behaviour

schema/migration.go
```diff
@@ -37,57 +46,365 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+}
+
+func convertFromBytes(toType FieldType, val []byte) any {
+	switch toType {
+	case StringType:
+		return string(val)
+	case ByteType:
+		return val
+	case UUIDType, BoolType, Int32Type, Int64Type, DoubleType, DateTimeType, ArrayType, ObjectType:
+		// FIXME: should we try to convert int64, UUID from it's binary representation?
```
garrensmith, line 190: I would say we should try that.

schema/migration.go
```diff
@@ -37,57 +46,365 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+	return applyPrimitiveFieldDelta(doc, change)
+}
+
+func applySchemaDelta(doc map[string]any, delta *VersionDelta) {
+	for _, d := range delta.Fields {
+		applyObjectFieldDelta(doc, d, 0)
+	}
+}
+
+// UpdateRowSchema fixes the schema of  provided, unmarshalled, document from given Version to the latest
```
garrensmith, line 386: Sorry silly nit but extra space in the comment here

garrensmith (dismissed):
This looks good. There are a few FIXME and TODO comments. How do we normally handle those to make sure we come back to them?

schema/migration.go
```diff
@@ -37,57 +47,425 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+	}
+
+	return true
+}
+
+func applyPrimitiveFieldDelta(doc map[string]any, change *VersionDeltaField) bool {
+	key := change.KeyPath[len(change.KeyPath)-1]
+
+	// array is not convertible to anything
+	if _, ok := doc[key].([]any); ok {
```
himank, line 400: Is the user allowed to change the type of the array?

garrensmith (approved):
Looks good

schema/migration.go
```diff
@@ -37,57 +47,425 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+		return true
+	}
+}
+
+func applyObjectFieldDelta(doc map[string]any, change *VersionDeltaField, start int) bool {
+	// descend to the last object in the key path
+	for i := start; i < len(change.KeyPath)-1; i++ {
+		switch m := doc[change.KeyPath[i]].(type) {
+		case map[string]any:
+			doc = m
```
himank, line 426: so for objects we are not doing anything?

schema/migration.go
```diff
@@ -37,57 +47,425 @@ func (x Versions) Latest() Version { return x[len(x)-1] }
...
+
+	return true
+}
+
+func applyPrimitiveFieldDelta(doc map[string]any, change *VersionDeltaField) bool {
+	key := change.KeyPath[len(change.KeyPath)-1]
+
+	// array is not convertible to anything
+	if _, ok := doc[key].([]any); ok {
+		delete(doc, key)
```
himank, line 401: At this point are we sure that this key is from the old type? Or do we need to check the type of array?

himank:
Looks good overall. One thing I realized is that we probably need better documentation on what all changes are allowed in the case of Arrays and Objects?

schema/migration.go
himank: actually, I realized that you have a method above that is handling it. Then in which case `line 400` condition can trigger?

test/v1/server/schema_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+package server
+
+import (
+	"github.com/tigrisdata/tigris/server/config"
+	"net/http"
+	"testing"
+)
+
+func TestSchemaMigration(t *testing.T) {
```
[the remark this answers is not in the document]
efirs: Array and object is not convertible to anything.
New field will be empty.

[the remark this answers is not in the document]
himank: Then we can add some checks around it when we receive the schema if not already present.

schema/rules.go
```diff
@@ -55,6 +56,10 @@ func (v *FieldSchemaValidator) Validate(existing *DefaultCollection, current *Fa
 	for name, f := range existingFields {
 		c, ok := currentFields[name]
 		if !ok {
+			if config.DefaultConfig.Schema.AllowIncompatible {
```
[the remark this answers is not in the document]
efirs: It can't be combined. It should not go to the below check `f.IsCompatible(c)`, when incompatible is allowed and field doesn't exists.

schema/migration.go
efirs: Right, not doing anything to the objects which are in the middle of the key path. We are just going do the key path here.

schema/migration.go
efirs: Good catch. This is not needed.

test/v1/server/schema_test.go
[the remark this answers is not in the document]
efirs: ah, ok.
We allow to change object/array to primitive.
It's only that the content of the object/array is not convertible to the primitive and new field will be empty

himank (approved):

PR #754, opened 2023-01-27
docs: Add RFC for secondary indexes

garrensmith:
## Describe your changes

This is the RFC for secondary indexes in Tigris

### DIFF
--- a/docs/rfcs/001-secondary-indexes.md
+++ b/docs/rfcs/001-secondary-indexes.md
@@ -1,0 +1,263 @@
+## Introduction
+
+This document describes the data model and index management for building and querying of Tigris secondary indexes.
+
+## Background
+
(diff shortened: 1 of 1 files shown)


rfcs/001-secondary-indexes.md
```diff
@@ -0,0 +1,263 @@
+## Introduction
```
efirs, line 1: Maybe move it inside docs directory?
`docs/rfcs/...`

himank: +1 I think moving it inside docs is better.

ovaistariq:
This is great. Like @himank and @efirs commented, let's move the `rfcs` directory under `docs` directory.

rfcs/001-secondary-indexes.md
garrensmith: Good idea. I've moved it under the docs.

ovaistariq (approved):

PR #756, opened 2023-01-29
feat: Create collation wrapper for value

garrensmith:
## Describe your changes

Uses text/collate to compare strings. This will create a more consistent string comparison. Currently only supports english and we can expand it later to support more languages.

This is a small PR to assist with some of the secondary index work

All tests should pass

### DIFF
--- a/value/value.go
+++ b/value/value.go
@@ -245,10 +245,6 @@ func NewStringValue(v string, collation *Collation) *StringValue {
 		Collation: collation,
 	}
 
-	// if collation != nil && collation.IsCaseInsensitive() {
-	// 	s.Value = strings.ToLower(s.Value)
-	// }
-
 	return s
 }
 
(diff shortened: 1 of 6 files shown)


value/value.go
```diff
@@ -234,18 +232,22 @@ func (d *DoubleValue) String() string {
...
+
 	s := &StringValue{
 		Value:     v,
 		Collation: collation,
 	}
 
-	if collation != nil && collation.IsCaseInsensitive() {
-		s.Value = strings.ToLower(s.Value)
-	}
+	// if collation != nil && collation.IsCaseInsensitive() {
```
himank, line 248: Let's remove it as now it is not needed.

value/value.go
```diff
@@ -260,17 +262,11 @@ func (s *StringValue) CompareTo(v Value) (int, error) {
 		return -2, fmt.Errorf("wrong type compared ")
 	}
 
-	if s.Collation != nil && s.Collation.IsCaseInsensitive() {
-		converted.Value = strings.ToLower(converted.Value)
-	}
+	// if s.Collation != nil && s.Collation.IsCaseInsensitive() {
```
himank, line 265: ^

himank (commented):
Looks good!

himank (approved):

efirs (approved):

PR #763, opened 2023-01-31
 Added cluster and collection metadata subspaces

efirs:
* Generalized metadata subspaces CRUDs
* Generalized main and search schema CRUDs

### DIFF
--- a/server/metadata/cluster.go
+++ b/server/metadata/cluster.go
@@ -1,0 +1,127 @@
+// Copyright 2022-2023 Tigris Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metadata
+
+import (
+	"context"
+	"time"
+
+	"github.com/google/uuid"
+	jsoniter "github.com/json-iterator/go"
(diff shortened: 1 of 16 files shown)


JigarJoshi (approved):

server/metadata/collection.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// CollectionSubspace is used to store metadata about Tigris collections.
+type CollectionSubspace struct {
+	metadataSubspace
+}
+
+// CollectionMetadata contains collection wide metadata.
+type CollectionMetadata struct {
+	OldestSchemaVersion int32
```
himank, line 34: `OldestSchemaVersion` may not be that intuitive, something like `NoUpdateNeededVersion` Or `StableCollectionVersion`

JigarJoshi (approved):

server/metadata/cluster.go
```diff
@@ -0,0 +1,127 @@
...
+
+var (
+	clusterVersion     = []byte{0x01}
+	clusterID          = uuid.Nil.String()
+	clusterMetadataKey = "key"
+)
+
+// ClusterMetadata keeps cluster wide metadata.
+type ClusterMetadata struct {
+	WorkerKeepalive time.Time
```
himank, line 42: is this for async?

efirs: right. also just placeholder for now.

himank: Okay, because I think we need something on the collection level, also I was thinking if we should have async metadata subspace? But we can figure it out as we go.

ovaistariq:
Can we also return information when listing or describing collections about collections that are undergoing schema updates?

efirs:
Yes, this diff adds a mechanism for us to persist per collection metadata.
We can return Latest and Oldest schema versions rows in the collection has, if they are not equal. Otherwise just Latest.

ovaistariq:
So I talked with Himank about two things:
- API that returns all versions of the collection (schema history view in web console)
- status of schema change on collection

We can show the status here in the status column 
image

efirs:
makes sense

PR #767, opened 2023-02-02
fix: Support query for nulls and non-indexed fields

himank:
## Describe your changes
- A filter can be passed to query null values `'{"filter": {"field":null}}'` or empty quoted string `'{"filter": {"str":""}}'`
- Allowing querying for bytes field which is not indexed, this is performed using full table scan. 
- Allow scientific notation in the filter.

### DIFF
--- a/value/value.go
+++ b/value/value.go
@@ -57,6 +57,10 @@ func NewValueUsingCollation(fieldType schema.FieldType, value []byte, collation 
 
 // NewValue returns the value of the field from the raw json value. It uses schema to get the type of the field.
 func NewValue(fieldType schema.FieldType, value []byte) (Value, error) {
+	if len(value) == 0 {
+		return NewNullValue(), nil
+	}
+
 	switch fieldType {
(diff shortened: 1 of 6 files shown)


ovaistariq (approved):

efirs (approved):

value/value.go
```diff
@@ -191,9 +195,14 @@ func NewDoubleValue(raw string) (*DoubleValue, error) {
...
 
+	f, err := strconv.ParseFloat(raw, 64)
+	if err != nil {
+		return nil, errors.InvalidArgument(fmt.Errorf("unsupported value type: %w ", err).Error())
+	}
+
 	i := &DoubleValue{
 		Double:   val,
-		asString: raw,
+		asString: strconv.FormatFloat(f, 'f', -1, 64),
```
efirs, line 205: Do we have test for this?

himank: Sorry missed this, we do have integration tests for it.

PR #784, opened 2023-02-08
Change database, collection, index metadata format to JSON

efirs:
* Metadata format of collections, indexes, databases upgraded to JSON.
* Consolidate all metadata subspaces in the metadata dictionary. As soon as dictionary is initialized all subspaces intialized too and can the cruds avalable under: `metadata.Dictionary.Collection()/Index()/Database().insert()/delete()/...`
* To avoid tests interference added variablity of the version handler and fixed test subspaces isolation.
* Split collection and project handler CRUDs in query runner, to reduce the function sizes and improve readability.

Possible further steps and concerns:
* Collection/Index/Database metadata still persisted to "encoding" subspace, maybe worth moving them to their own "collection", "index", "database" subspaces.
* Namespace metadata is persisted in two places: "reserved" and "namespace" subspaces.


server/metadata/cluster.go
```diff
@@ -16,94 +16,77 @@ package metadata
...
+
 // ClusterSubspace is used to store metadata about Tigris clusters.
 type ClusterSubspace struct {
 	metadataSubspace
 }
 
 var (
-	clusterVersion     = []byte{0x01}
 	clusterID          = uuid.Nil.String()
 	clusterMetadataKey = "key"
```
himank, line 38: "key" may not that intuitive, should we follow the same convention as we have for remaining and call it "cluster" or something similar?

server/metadata/collection.go
```diff
@@ -18,107 +18,168 @@ import (
...
+) error {
 	return c.deleteMetadata(ctx, tx,
-		c.validateArgs(nsID, dbID, collID, nil),
-		c.getKey(nsID, dbID, collID),
+		c.validateArgs(nsID, dbID, name, nil),
+		c.getKey(nsID, dbID, name),
+	)
+}
+
+func (c *CollectionSubspace) softDelete(ctx context.Context, tx transaction.Tx, nsID uint32, dbID uint32, name string,
```
himank, line 120: If it is easy, just call delete and we can kill `softDelete`

server/metadata/collection.go
```diff
@@ -18,107 +18,168 @@ import (
...
+	return c.softDeleteMetadata(ctx, tx,
+		c.validateArgs(nsID, dbID, name, nil),
+		c.getKey(nsID, dbID, name),
+		newKey,
 	)
 }
 
-func (u *CollectionSubspace) validateArgs(nsID uint32, dbID uint32, collID uint32, metadata **CollectionMetadata) error {
-	if nsID == 0 || dbID == 0 || collID == 0 {
+func (_ *CollectionSubspace) validateArgs(nsID uint32, dbID uint32, name string, metadata **CollectionMetadata) error {
```
himank, line 132: why are we validating pointer to pointer here?

server/metadata/database.go
```diff
@@ -15,38 +15,31 @@
...
 	MainBranch          = "main"
 	BranchNameSeparator = "_$branch$_"
 )
 
+// DatabaseMetadata contains database wide metadata.
+type DatabaseMetadata struct {
+	ID   uint32 `json:"id,omitempty"`
+	Name string `json:"name,omitempty"`
+
+	SchemaVersion int32 `json:"schema_version"`
```
himank, line 40: remove it if we are not using it.

server/metadata/database.go
```diff
@@ -101,3 +95,148 @@ func (b *DatabaseName) IsMainBranch() bool {
...
+	metadataSubspace
+}
+
+const dbMetaValueVersion int32 = 1
+
+func newDatabaseStore(nameRegistry *NameRegistry) *DatabaseSubspace {
+	return &DatabaseSubspace{
+		metadataSubspace{
+			SubspaceName: nameRegistry.EncodingSubspaceName(),
+			KeyVersion:   []byte{encKeyVersion},
```
himank, line 110: Similar to collection, have a key version for this subspace as well.

server/metadata/errors.go
```diff
@@ -29,6 +29,9 @@ const (
 	ErrCodeSearchIndexNotFound  ErrorCode = 0x07
 	ErrCodeCacheExists          ErrorCode = 0x08
 	ErrCodeCacheNotFound        ErrorCode = 0x09
+
+	ErrDBMismatch     ErrorCode = 0x0A
+	ErrBranchMismatch ErrorCode = 0x0A
```
himank, line 34: we need to assign different codes.

server/metadata/index.go
```diff
@@ -0,0 +1,177 @@
...
+	jsoniter "github.com/json-iterator/go"
+	"github.com/rs/zerolog/log"
+	"github.com/tigrisdata/tigris/errors"
+	"github.com/tigrisdata/tigris/internal"
+	"github.com/tigrisdata/tigris/keys"
+	"github.com/tigrisdata/tigris/server/transaction"
+	ulog "github.com/tigrisdata/tigris/util/log"
+)
+
+// IndexSubspace is used to store metadata about Tigris collections.
```
himank, line 29: `about Tigris secondary indexes`

server/metadata/index.go
```diff
@@ -0,0 +1,177 @@
...
+type IndexSubspace struct {
+	metadataSubspace
+}
+
+// IndexMetadata contains collection wide metadata.
```
himank, line 34: comment need fixing.

server/metadata/metadata.go
```diff
@@ -53,29 +79,80 @@ func (m *metadataSubspace) getMetadata(ctx context.Context, tx transaction.Tx, i
...
+		return err
+	}
+
+	if payload == nil {
+		return errors.ErrNotFound
 	}
 
-	log.Debug().Str("type", string(m.SubspaceName)).Str("key", key.String()).Str("value", string(row.Data.RawData)).Msgf("reading metadata succeed")
+	if err = jsoniter.Unmarshal(payload.RawData, metadata); ulog.E(err) {
+		return errors.Internal("failed to unmarshal collection metadata")
```
himank, line 112: `failed to unmarshal metadata`

himank, line 112: error msg fix.

server/metadata/metadata.go
```diff
@@ -86,3 +163,67 @@ func (m *metadataSubspace) deleteMetadata(ctx context.Context, tx transaction.Tx
 
 	return err
 }
+
+func (m *metadataSubspace) softDeleteMetadata(ctx context.Context, tx transaction.Tx, invalidArgs error,
+	fromKey keys.Key, toKey keys.Key,
+) (err error) {
```
himank, line 169: do we need named return here?

server/services/v1/database/query_runner.go
```diff
@@ -179,7 +179,21 @@ func (runner *BaseQueryRunner) getDatabase(_ context.Context, tx transaction.Tx,
...
 		// this means that some DDL operation has modified the database object, then we need to perform all the operations
 		// on this staged database.
-		return tx.Context().GetStagedDatabase().(*metadata.Database), nil
+
+		db, ok := tx.Context().GetStagedDatabase().(*metadata.Database)
+		if !ok {
+			return nil, errors.Internal("invalid transaction staged database")
+		}
+
+		if db.Name() != projName {
```
himank, line 188: What about in case of database branch then this won't match. I think there is a method that returns db name we need to compare with that.

server/services/v1/database/query_runner.go
```diff
@@ -179,7 +179,21 @@ func (runner *BaseQueryRunner) getDatabase(_ context.Context, tx transaction.Tx,
...
+			return nil, errors.InvalidArgument("collections should belong to the same project in the transaction")
+		}
+
+		if db.BranchName() != branch && (branch != "" || db.BranchName() != metadata.MainBranch) {
```
himank, line 192: `if db.IsBranch() && db.BranchName() != branch {}`

server/metadata/metadata.go
```diff
@@ -17,31 +17,57 @@ package metadata
...
+}
+
+func (m *metadataSubspace) insertPayload(ctx context.Context, tx transaction.Tx, invalidArgs error, key keys.Key,
+	ver int32, payload []byte,
+) error {
+	if invalidArgs != nil {
+		return invalidArgs
+	}
+
+	data := internal.NewTableData(payload)
```
himank, line 58: Call `NewTableDataWithVersion(payload, ver)`

server/metadata/metadata.go
```diff
@@ -53,29 +79,80 @@ func (m *metadataSubspace) getMetadata(ctx context.Context, tx transaction.Tx, i
...
+	ver int32, payload []byte,
+) error {
 	if invalidArgs != nil {
 		return invalidArgs
 	}
 
-	_, err := tx.Update(ctx, key, func(data *internal.TableData) (*internal.TableData, error) {
-		return internal.NewTableData(payload), nil
+	_, err := tx.Update(ctx, key, func(_ *internal.TableData) (*internal.TableData, error) {
+		data := internal.NewTableData(payload)
```
himank, line 126: `NewTableDataWithVersion`

server/cdc/listener.go
```diff
@@ -46,7 +46,7 @@ func (p *Publisher) OnCommit(ctx context.Context, tx transaction.Tx, listener kv
 		return err
 	}
 
-	td := internal.NewTableDataWithEncoding(json, int32(internal.JsonEncoding))
+	td := internal.NewTableDataWithVersion(json, 1)
```
garrensmith, line 49: Lets make the magic number a const from the `internal` package. `internal.DefaultVersion`

server/metadata/cluster_test.go
```diff
@@ -22,17 +22,16 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/tigrisdata/tigris/errors"
+	"github.com/tigrisdata/tigris/lib/uuid"
 	"github.com/tigrisdata/tigris/server/transaction"
 )
 
 var testClusterPayload = &ClusterMetadata{
-	WorkerKeepalive: time.Now().UTC(),
+	ID: uuid.New(),
```
garrensmith, line 30: What do we use the cluster id for? And what do we plan to use it for in the future?

server/metadata/metadata.go
```diff
@@ -53,29 +78,77 @@ func (m *metadataSubspace) getMetadata(ctx context.Context, tx transaction.Tx, i
...
+}
+
+func (m *metadataSubspace) updatePayload(ctx context.Context, tx transaction.Tx, invalidArgs error, key keys.Key,
+	ver int32, payload []byte,
+) error {
+	if invalidArgs != nil {
+		return invalidArgs
+	}
+
+	_, err := tx.Update(ctx, key, func(_ *internal.TableData) (*internal.TableData, error) {
```
garrensmith, line 124: Would it not be better to make this a replace? Do we need to read the key before updating it?

server/metadata/key_encoder.go
```diff
@@ -93,6 +91,7 @@ func (d *DictKeyEncoder) EncodeIndexName(idx *schema.Index) []byte {
 	return d.encodedIdxName(idx)
 }
 
+/*
```
garrensmith, line 94: Should we remove this dead code?

server/services/v1/database/query_runner.go
garrensmith: Could we move this check into a function like `db.IsBranch(metaData)` I think we use something similar in a few places.

schema/schema.go
```diff
@@ -117,6 +117,7 @@ type JSONSchema struct {
 	PrimaryKeys     []string            `json:"primary_key,omitempty"`
 	CollectionType  string              `json:"collection_type,omitempty"`
 	IndexingVersion string              `json:"indexing_version,omitempty"`
+	Version         int32               `json:"version,omitempty"`
```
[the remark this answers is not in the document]
himank: the only problem would be that we need to make sure client changes don't go first otherwise the schema will have this field already filled.

server/metadata/cluster.go
himank: We need it earlier when it was a single namespace, but probably not now. But I'd suggest having it just in case? as it allows expanding the key.

server/metadata/database.go
himank: Okay, in your previous diff I think collection had its own keyVersion so I got confused.

himank (approved):

schema/schema.go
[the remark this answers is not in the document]
efirs: We'll make sure.

server/metadata/cluster.go
efirs: yeah, keeping it.

PR #795, opened 2023-02-10
feat: Add search API for search product

himank:
## Describe your changes
Adding the search API for search indexes. 

Todo: there is some redundancy in the reader/search logic that we have here and similar already exists in the database package. Need to consolidate it in further diffs.

### DIFF
--- a/server/services/v1/search/reader.go
+++ b/server/services/v1/search/reader.go
@@ -1,0 +1,304 @@
+// Copyright 2022-2023 Tigris Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


ovaistariq (approved):

ovaistariq (dismissed):

server/services/v1/search/reader.go
```diff
@@ -0,0 +1,304 @@
...
+			if doc, createdAt, updatedAt, it.err = UnpackSearchFields(it.index, doc); it.err != nil {
+				return false
+			}
+
+			// now apply the filter
+			if !it.filter.MatchesDoc(doc) {
+				continue
+			}
+
+			var rawData []byte
```
efirs, line 244: Combine with var section above.

efirs (approved):

PR #798, opened 2023-02-10
fix: Added branch tag for metrics

JigarJoshi:
## Describe your changes
Added `branch` metrics tag. (Thanks to @pboros & @adilansari)

Existing unit tests + manually verified on promethous endpoint `/metrics`. We will further test with debug image on dev environment.


server/config/options.go
```diff
@@ -67,6 +67,8 @@ type AuthConfig struct {
 	ManagementClientId        string        `mapstructure:"management_client_id" yaml:"management_client_id" json:"management_client_id"`
 	ManagementClientSecret    string        `mapstructure:"management_client_secret" yaml:"management_client_secret" json:"management_client_secret"`
 	TokenClockSkewDurationSec int           `mapstructure:"token_clock_skew_duration_sec" yaml:"token_clock_skew_duration_sec" json:"token_clock_skew_duration_sec"`
+	GotrueUrl                 string        `mapstructure:"gotrue_url" yaml:"gotrue_url" json:"gotrue_url"`
+	GotrueUsernameSuffix      string        `mapstructure:"gotrue_username_suffix" yaml:"gotrue_username_suffix" json:"gotrue_username_suffix"`
```
pboros, line 71: Maybe these are from another change?

pboros (dismissed):
One minor comment, otherwise looks good to me!

server/config/options.go
JigarJoshi: ah right this is from other changes. Fixed it.

adilansari (dismissed):

server/metrics/search.go
```diff
@@ -33,6 +33,7 @@ func getSearchOkTagKeys() []string {
 		"env",
 		"project",
 		"db",
+		"branch",
```
ovaistariq, line 36: Should this be here?

himank: In this context, this tag is not helpful. This is used by the low-level storage, there we don't need branch-specific tags.

adilansari (approved):

PR #806, opened 2023-02-13
feat: Secondary Indexer

### LINKED ISSUE #796: Add support to index a document
Index a document on insert, update and delete

garrensmith:
## Describe your changes
Adds a Secondary index indexer that reads documents in the update transaction and builds a secondary index based on the queryable fields defined in the schema.

The indexer is off by default but can be enabled via the config.

See the [RFC](https://github.com/tigrisdata/tigris/blob/main/docs/rfcs/001-secondary-indexes.md) for the full details

The code to read from the index will be in another PR. 

The current tests should all pass

This closes #796


server/services/v1/database/secondary_indexer.go
```diff
@@ -0,0 +1,489 @@
...
+	return nil
+}
+
+func (q *SecondaryIndexer) buildAddAndRemoveKVs(newTableData *internal.TableData, oldTableData *internal.TableData, primaryKey []interface{}) *IndexerUpdateSet {
+	newRows := q.buildTableRows(newTableData)
+	oldRows := q.buildTableRows(oldTableData)
+
+	// Remove any rows in the oldRow set that will be overwritten in the newRows
+	// We do not want to issue an unneeded delete when it will be overwritten with the update
+	rowsToRemove := removeDuplicateRows(newRows, oldRows)
```
himank, line 228: Your logic of removing will works fine in case of replacement. But in the case of partial updates, it will remove the rows for the fields that are not passed in the new data. For updates, we need to only remove rows that are changed in the new payload i.e. 
- New row should have the field which is also present in the old row. 
- The value in the new row for that field shouldn't match the value in the old row.

server/services/v1/database/query_runner.go
```diff
@@ -261,6 +263,22 @@ func (runner *UpdateQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
...
+				if err != nil {
+					return Response{}, nil, err
+				}
+				err = secondaryIndexer.Index(ctx, tx, newData, key.IndexParts())
+				if err != nil {
+					return Response{}, nil, err
+				}
+			}
+		} else if config.DefaultConfig.SecondaryIndex.WriteEnabled {
+			err := secondaryIndexer.Update(ctx, tx, newData, row.Data, key.IndexParts())
```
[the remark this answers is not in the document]
garrensmith: The update will do a delete and insert. It will only delete the fields that have been changed

server/services/v1/database/secondary_indexer.go
garrensmith: I've written a few tests here https://github.com/tigrisdata/tigris/blob/cb0c33395786c2b81be8d0bc063337f3a6870706/server/services/v1/database/secondary_indexer_test.go#L90

To test this logic. The basic process is:
1. Build key values for old and new doc
2. Remove keys from the old doc that are exactly the same in the new doc
3. Remove keys from the new doc that are exactly the same as the old doc
4. Remove old keys from the index
4. Add new keys into index

garrensmith (commented):
Thanks for the feedback that was helpful. I've resolved the ones I've fixed and left some comments on the other ones.

server/services/v1/database/search_indexer.go
```diff
@@ -97,6 +97,10 @@ func (i *SearchIndexer) OnPostCommit(ctx context.Context, tenant *metadata.Tenan
 				return err
 			}
 
+			if tableData == nil {
```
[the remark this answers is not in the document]
garrensmith: From what I can see, when we add an index row, on the post commit those updates are being passed here. All the index key values have `nil` for the tableData since they only adding keys. That is causing an issue here.

server/services/v1/database/secondary_indexer.go
himank: So I think you are cleaning up the old fields as well. Can we have a test for this case,
```
old := `{"a": 1, "b": 2, "c": "foo", "d": {"d1": 1, "d2": "foo"}, "e": {"e1": "1", "e2": "foo"}}`
new := `{"a: 2, "d.f1": 2, "e": {"e1": 2}}`
```

Outcome,
```
a-->2
d.f1-->2
d.d2-->foo
e.e1 --> 2
```

- `b, c` are retained from the old doc
- `d.d2` is retained from the old doc
- `e` is completely replaced by the user so only `e.e1`
- `a` is updated.

server/services/v1/database/search_indexer.go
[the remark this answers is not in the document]
himank: ah, now I see - makes sense.

server/services/v1/database/secondary_indexer.go
```diff
@@ -0,0 +1,489 @@
...
+			continue
+		}
+
+		if field.DataType == schema.ArrayType {
+			rows = append(rows, q.indexArray(tableData.RawData, field, field.KeyPath())...)
+		} else {
+			row, err := q.indexField(tableData.RawData, field.FieldName, field.DataType, 0, field.KeyPath()...)
+			if err != nil {
+				log.Err(err).Msgf("Failed to index field name: %s", field.FieldName)
+				continue
```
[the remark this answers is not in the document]
himank: Hmm, I think that's the behavior we need to make sure this code path is well-tested.

server/services/v1/database/secondary_indexer.go
```diff
@@ -0,0 +1,489 @@
...
+			indexedFields := q.indexNestedField(value, field.FieldName, pos)
+			rows = append(rows, indexedFields...)
+		case schema.ArrayType:
+			indexedField, _ := newIndexRow(toFieldType, q.collation, field.FieldName, nil, pos, true)
+			rows = append(rows, *indexedField)
+		default:
+			indexedField, err := newIndexRow(field.SubType, q.collation, field.FieldName, value, pos, false)
+			if err != nil {
+				log.Err(err).Msgf("Failed to index field name: %s", field.FieldName)
+				return
```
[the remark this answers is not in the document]
himank: But until we have any way to fix these missing indexes we need to fail the transaction. This will also help us catch all the initial bugs. Once we have some way to fix these then we can log it and skip it.

server/services/v1/database/secondary_indexer.go
himank: Just one correction, `new` I am referring to the request payload for the Update request from the user.

server/services/v1/database/secondary_indexer.go
```diff
@@ -0,0 +1,559 @@
...
+		if val, ok := rowSizes[row.name]; ok {
+			rowSizes[row.Name()] = val + sizeIncrease
+		} else {
+			rowSizes[row.Name()] = sizeIncrease
+		}
+	}
+	return indexKeys, rowSizes, rowCounts
+}
+
+func (q *SecondaryIndexer) indexField(doc []byte, fieldName string, dataType schema.FieldType, pos int, keyPath ...string) (*IndexRow, error) {
```
himank, line 384: nit: you can return `*IndexRow, bool, error` so that if a field is not indexable then you can return false instead of error but you don't need to do it in this diff. This can be combined when we have an attribute on the field to tell us whether the field needs to be indexed.

himank (approved):

server/services/v1/database/secondary_indexer.go
garrensmith: Awesome. Nice idea

PR #808, opened 2023-02-13
feat: Add a util.Last to get last item in array

garrensmith:
## Describe your changes
Instead of doing lists[len-1] have a function that does this for us. Makes the code a little more readable.

Existing tests should all pass

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -65,3 +65,11 @@ func JSONToMap(data []byte) (map[string]any, error) {
 
 	return decoded, nil
 }
+
+func Last[T any](list []T) T {
+	if len(list) == 0 {
+		var zeroVal T
+		return zeroVal
+	}
+	return list[len(list)-1]
+}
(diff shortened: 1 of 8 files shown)


util/util.go
```diff
@@ -65,3 +65,7 @@ func JSONToMap(data []byte) (map[string]any, error) {
 
 	return decoded, nil
 }
+
+func Last[T any](list []T) T {
```
himank, line 69: Return `null` if the list is empty.

efirs (approved):

util/util.go
```diff
@@ -65,3 +65,11 @@ func JSONToMap(data []byte) (map[string]any, error) {
...
+	if len(list) == 0 {
+		var zeroVal T
+		return zeroVal
```
himank, line 72: the problem with this approach is it would return `0` for example for an integer null array which is a valid value.

himank: Or we can do this,
```
func Last[T any](list []T) (T, bool) {
	if len(list) == 0 {
		var zeroVal T
		return zeroVal, false
	}
	return list[len(list)-1], true
}
```

garrensmith:
Closing this. I don't like the zeroval return value and I don't want to have to two return values. I'll see if I can come up with a better solution

PR #829, opened 2023-02-16
feat: Secondary index reader

garrensmith:
## Describe your changes
Secondary Index reader that queries from the secondary index.


server/services/v1/database/query_runner.go
```diff
@@ -413,14 +414,22 @@ func (runner *StreamingQueryRunner) buildReaderOptions(collection *schema.Defaul
...
 
+	if config.DefaultConfig.SecondaryIndex.ReadEnabled {
+		if options.iquery, err = runner.buildSecondaryIndexKeysUsingFilter(collection, runner.req.Filter); err == nil {
+			options.secondaryIndex = true
+			return options, nil
+		}
+	}
+
+	// TODO: I think this might need to be changed, we should only support search on the search endpoint
 	if options.filter.None() || !options.filter.IsIndexed() {
```
garrensmith, line 425: Do we still support search as a reader option or should it only be part of the search API?

query/filter/key_builder.go
```diff
@@ -81,38 +158,66 @@ func (k *KeyBuilder) Build(filters []Filter, userDefinedKeys []*schema.Field) ([
...
+//     are present in the filters. The following rules are applied for StrictEqKeyComposer:
+//     - The userDefinedKeys(indexes defined in the schema) passed in parameter should be present in the filter
+//     - For AND filters it is possible to build internal keys for composite indexes, for OR it is not possible.
 //
-// So for OR filter an error is returned if it is used for indexes that are composite.
-type StrictEqKeyComposer struct {
+// 2. When `matchAll=false`, it will treat all userDefined as individual and generate an `$eq` query plan for each one that is found.
+//
+// For OR filter an error is returned if it is used for indexes that are composite.
+type StrictEqKeyComposer[F fieldable] struct {
```
garrensmith, line 196: It might be better to split this out into two. `PKStrictEqKeyComposer` and `SecondaryEqComposer`

query/filter/key_builder.go
```diff
@@ -121,56 +226,176 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
...
 	}
 
-	return allKeys, nil
+	return queryPlans, nil
+}
+
+// Range Key Composer will generate a range key set on the user defined keys
+// It will set the KeyQuery to `FullRange` if the start or end key is not defined in the query
+// if there is a defined start and end key for a range then `Range` is set.
+type RangeKeyComposer[F fieldable] struct {
```
himank, line 301: For this diff, the structure is good. But if we think about it, we just need two high level abstractions here,
  - `NewQueryableKeyBuilder(composer, primaryKey)` [will be used by insert/replace only]
  - `NewQueryableKeyBuilder()`: This will internally use eq/range composer to build plans i.e. equality as well as range. The reason being once we don't rely on the search store then either the plan is formed or we need to fall back to full table scan, there will be no other option. Then `Read/Update/Delete` can rely directly on it. This means we need to add something to the QueryPlan so that caller knows whether the keys are for primary or secondary indexes.

server/services/v1/database/query_runner.go
garrensmith: Created issue https://github.com/tigrisdata/tigris/issues/869 for this

query/filter/key_builder.go
```diff
@@ -15,32 +15,109 @@
...
-	composer KeyComposer
+// KeyBuilder uses generics so that it can accept either schema.QueryableField or schema.Field
+// so that it can build a query plan for primay or secondary indexes.
+type KeyBuilder[F fieldable] struct {
+	composer   KeyComposer[F]
+	primaryKey bool
+}
+
+// NewPrimaryKeyEQBuild returns a KeyBuilder for use with schema.Field to build a primary key query plan.
+func NewPrimaryKeyEqBuilder(keyEncodingFunc KeyEncodingFunc) *KeyBuilder[*schema.Field] {
```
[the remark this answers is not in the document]
garrensmith: I think we could do that a little later once secondary indexes are deployed. At the moment we have to pass some extra params if it is a PrimaryKey versus secondary.

query/filter/key_builder.go
garrensmith: Yes that makes sense.

query/filter/key_builder.go
```diff
@@ -121,56 +226,176 @@ func (s *StrictEqKeyComposer) Compose(selectors []*Selector, userDefinedKeys []*
 		}
 
 		if len(repeatedFields) == 0 {
-			// nothing found or a gap
-			return nil, errors.InvalidArgument("filters doesn't contains primary key fields")
+			if s.matchAll {
```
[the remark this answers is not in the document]
garrensmith: Yes that will be awesome. I've created https://github.com/tigrisdata/tigris/issues/870 to remember to do this.

query/filter/key_builder.go
garrensmith: Yes I like it. We can definitely move to that direction once search and secondary are ready.

server/services/v1/database/secondary_index_reader.go
```diff
@@ -0,0 +1,177 @@
...
+	buildIndexParts := func(fieldName string, datatype schema.FieldType, value interface{}) []interface{} {
+		version := getFieldVersion(fieldName, coll)
+		return []interface{}{fieldName, version, value}
+	}
+
+	eqKeyBuilder := filter.NewSecondaryKeyEqBuilder[*schema.QueryableField](encoder, buildIndexParts)
+	eqPlan, err := eqKeyBuilder.Build(queryFilters, coll.QueryableFields)
+	if err == nil {
+		for _, plan := range eqPlan {
+			if indexedDataType(plan) {
```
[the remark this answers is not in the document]
garrensmith: This can happen if the field is a `[]byte`. Then the plan is ignored and we look for another `eq` plan or move to range plans

garrensmith (commented):
Thanks for the feedback. I've made the changes.

query/filter/filter.go
```diff
@@ -303,6 +316,12 @@ func buildValueMatcher(input jsoniter.RawMessage, field *schema.QueryableField)
 			return nil, nil, err
 		}
 		collation = value.NewCollationFrom(apiCollation)
+
+		if buildForSecondaryIndex && collation.IsCaseInsensitive() {
```
himank, line 320: So we won't be supporting case-insensitive queries on secondary indexes fields?

query/filter/key_builder.go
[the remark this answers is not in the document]
himank: great

query/filter/key_builder.go
himank: sg

query/filter/filter.go
garrensmith: We can support it in the in-memory filter part but not part of what we use to build the range keys. Or at least I can't think of a way of making that work.

himank: Sure, we can figure it out iteratively.

himank (approved):

adilansari (approved):

PR #830, opened 2023-02-16
fixes for secondary indexes and integration tests

garrensmith:
## Describe your changes

chore: Enable secondary index writes for tests

fix: Ignore missing fields for arrrays fields

fix: Refactor the way the iterator was chosen for reading the docs.
There was a subtle bug because the variables were declared at the top.

fix: Add a separate table for secondary indexes so that it doesn't
interfere with scanning the primary key index and none of the index
writes get sent to the search indexer

All tests should pass

### DIFF
--- a/internal/data.go
+++ b/internal/data.go
@@ -26,7 +26,7 @@ import (
 
 var (
 	UserTableKeyPrefix      = []byte("data")
-	SecondaryTableKeyPrefix = []byte("sidx")
+	SecondaryTableKeyPrefix = []byte("idx")
 	PartitionKeyPrefix      = []byte("part")
 	CacheKeyPrefix          = "cache"
 )
--- a/internal/data.go
+++ b/internal/data.go
@@ -25,9 +25,10 @@ import (
 )
 
 var (
-	UserTableKeyPrefix = []byte("data")
-	PartitionKeyPrefix = []byte("part")
-	CacheKeyPrefix     = "cache"
+	UserTableKeyPrefix      = []byte("data")
+	SecondaryTableKeyPrefix = []byte("sidx")
+	PartitionKeyPrefix      = []byte("part")
+	CacheKeyPrefix          = "cache"
 )
 
 var bh codec.BincHandle
--- a/schema/collection.go
+++ b/schema/collection.go
@@ -43,6 +43,8 @@ type DefaultCollection struct {
 	Name string
 	// EncodedName is the encoded name of the collection.
 	EncodedName []byte
+	// EncodedSecondaryName is the encoded name of the collection's Secondary Index.
+	EncodedSecondaryName []byte
 	// Fields are derived from the user schema.
 	Fields []*Field
 	// Indexes is a wrapper on the indexes part of this collection.
--- a/schema/collection.go
+++ b/schema/collection.go
@@ -43,8 +43,8 @@ type DefaultCollection struct {
 	Name string
 	// EncodedName is the encoded name of the collection.
 	EncodedName []byte
-	// EncodedSecondaryName is the encoded name of the collection's Secondary Index.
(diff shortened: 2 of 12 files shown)


server/config/options.go
```diff
@@ -226,8 +226,8 @@ var DefaultConfig = Config{
 		WriteEnabled: true,
 	},
 	SecondaryIndex: SecondaryIndexConfig{
-		ReadEnabled:  false,
-		WriteEnabled: false,
+		ReadEnabled:  true,
```
himank, line 229: We don't want to merge this with read enabled.

server/services/v1/database/base_runner.go
```diff
@@ -246,37 +246,37 @@ func (runner *BaseQueryRunner) getWriteIterator(ctx context.Context, tx transact
...
 	}
 
+	iterator, err := reader.FilteredRead(pkIterator, filter.NewWrappedFilter(filters))
+
+	metrics.SetWriteType("pkey")
 	return iterator, nil
 }
+
+func (runner *BaseQueryRunner) buildPrimaryKey(coll *schema.DefaultCollection) (keys.Key, error) {
+	return runner.encoder.EncodeKey(coll.EncodedName, coll.Indexes.PrimaryKey, []interface{}{})
```
himank, line 281: do we need to pass an empty slice at the end?

server/services/v1/database/base_runner.go
```diff
@@ -246,37 +246,37 @@ func (runner *BaseQueryRunner) getWriteIterator(ctx context.Context, tx transact
...
+		return nil, err
 	}
 
+	iterator, err := reader.FilteredRead(pkIterator, filter.NewWrappedFilter(filters))
+
+	metrics.SetWriteType("pkey")
 	return iterator, nil
 }
+
+func (runner *BaseQueryRunner) buildPrimaryKey(coll *schema.DefaultCollection) (keys.Key, error) {
```
himank, line 280: can we call it something like, `tableDataScanKey`

server/services/v1/database/query_runner.go
```diff
@@ -550,7 +558,7 @@ func (runner *StreamingQueryRunner) iterateOnKvStore(ctx context.Context, tx tra
 			// pass it to filterable
 			iter, err = reader.FilteredRead(iter, options.filter)
 		}
-	} else if iter, err = reader.ScanTable(options.table); err == nil {
+	} else if iter, err = reader.ScanPrimaryKeyTable(options.primaryKeyFrom); err == nil {
```
himank, line 561: I think calling it ScanTable is still fine as the scan table in this context is still valid.

server/services/v1/database/base_runner.go
garrensmith: Sadly yes. Otherwise it errors

JigarJoshi (approved):

internal/data.go
```diff
@@ -25,9 +25,10 @@ import (
 )
 
 var (
-	UserTableKeyPrefix = []byte("data")
-	PartitionKeyPrefix = []byte("part")
-	CacheKeyPrefix     = "cache"
+	UserTableKeyPrefix      = []byte("data")
+	SecondaryTableKeyPrefix = []byte("sidx")
```
himank, line 29: let's simply call it `idx`?

schema/collection.go
```diff
@@ -43,6 +43,8 @@ type DefaultCollection struct {
 	Name string
 	// EncodedName is the encoded name of the collection.
 	EncodedName []byte
+	// EncodedSecondaryName is the encoded name of the collection's Secondary Index.
+	EncodedSecondaryName []byte
```
himank, line 47: `EncodedTableIndexName`

himank (approved):

PR #842, opened 2023-02-17
fix: Handle floating point greater than 32 bit properly

himank:
## Describe your changes
1. This is fixing the floating point handling when the value can’t fit in 32-bit. The underlying indexing store is parsing the floating values in float(32bit) which means there are two problems:
  - The smallest floating point normal number needed to be less than ~ 1.1755e-38
  - The number has to be in the range -3.4e38/3.4e38

Outside of this range, an error is returned that was getting swallowed, resulting in empty documents returned for that query. I am fixing that and also handling queries for 64-bit floating-point numbers. 

2. This diff is also allowing collection names to start with `_`

### DIFF
--- a/query/filter/selector.go
+++ b/query/filter/selector.go
@@ -17,6 +17,7 @@ package filter
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 
 	"github.com/buger/jsonparser"
 	"github.com/tigrisdata/tigris/lib/date"
@@ -144,7 +145,21 @@ func (s *Selector) ToSearchFilter() []string {
 }
 
 func (s *Selector) IsIndexed() bool {
-	return !(s.Field.DataType == schema.ByteType || s.Matcher.GetValue().AsInterface() == nil)
+	switch {
+	case s.Field.DataType == schema.DoubleType:
+		v, ok := s.Matcher.GetValue().(*value.DoubleValue)
+		if !ok {
+			return false
+		}
+
+		if v.Double < value.SmallestNonZeroNormalFloat32 && v.Double > -value.SmallestNonZeroNormalFloat32 {
+			return false
+		}
+
+		return v.Double < math.MaxFloat32 && v.Double > -math.MaxFloat32
+	default:
+		return !(s.Field.DataType == schema.ByteType || s.Matcher.GetValue().AsInterface() == nil)
+	}
 }
 
 // String a helpful method for logging.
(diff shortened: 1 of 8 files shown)


store/search/ts.go
```diff
@@ -281,6 +281,24 @@ func (s *storeImpl) Search(_ context.Context, table string, query *qsearch.Query
...
+				Message string `json:"error"`
+			}
+
+			type errorsResult struct {
+				Res []errResult `json:"results"`
+			}
+
+			var errorsRes errorsResult
+			err = jsoniter.Unmarshal(res.Body, &errorsRes)
+			if err = jsoniter.Unmarshal(res.Body, &errorsRes); err == nil && len(errorsRes.Res) > 0 {
```
efirs, line 297: Is this duplication intentional?

query/filter/selector.go
```diff
@@ -144,7 +145,17 @@ func (s *Selector) ToSearchFilter() []string {
 }
 
 func (s *Selector) IsIndexed() bool {
-	return !(s.Field.DataType == schema.ByteType || s.Matcher.GetValue().AsInterface() == nil)
+	switch {
+	case s.Field.DataType == schema.DoubleType:
+		v := s.Matcher.GetValue().(*value.DoubleValue).Double
```
efirs, line 150: It's better to protect type assertion.

himank: the case is first checking if it is double, the building of value guarantees this as value is built from field type.

store/search/ts.go
himank: ah, good catch.

query/filter/selector.go
efirs: Linters would still complain.

himank: hmm

efirs (approved):

schema/fields_test.go
```diff
@@ -62,6 +62,10 @@ func TestFieldBuilder_Build(t *testing.T) {
 				builder:  &FieldBuilder{FieldName: "test", Type: "integer", Primary: &boolTrue},
 				expError: nil,
 			},
+			{
+				builder:  &FieldBuilder{FieldName: "_test", Type: "integer"},
```
efirs, line 66: Is this tests new behavior?

himank: no, it is just to test existing behavior. We already allow field name with `_`

PR #889, opened 2023-03-01
feat: Configure which fields to add to index

### LINKED ISSUE #875: Secondary indexes are specified in schema
Update secondary indexes to be specified in the schema:

- [ ] Store which fields are indexes
- [ ] Store if index is ready to be read
- [ ] Build function to build index from a table scan

garrensmith:
## Describe your changes
Secondary indexes are now configurable via the schema. Adding `index: true` to a top level field will add that field to the secondary index. The schema rules have been updated to make sure that only top level fields can be indexed.

All tests should pass. I've updated some of the update and delete test's so that the queries will use the secondary index

Fixes #875

### DIFF
--- a/schema/rules.go
+++ b/schema/rules.go
@@ -26,7 +26,7 @@ var (
 )
 
 var validators = []Validator{
-	&IndexSchemaValidator{},
+	&PrimaryIndexSchemaValidator{},
 	&FieldSchemaValidator{},
 }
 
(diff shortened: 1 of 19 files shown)


schema/rules.go
```diff
@@ -43,12 +44,43 @@ type SearchIndexValidator interface {
...
 
-type IndexSchemaValidator struct{}
+type PrimaryIndexSchemaValidator struct{}
 
-func (v *IndexSchemaValidator) Validate(existing *DefaultCollection, current *Factory) error {
+func (v *PrimaryIndexSchemaValidator) Validate(existing *DefaultCollection, current *Factory) error {
 	return existing.Indexes.PrimaryKey.IsCompatible(current.Indexes.PrimaryKey)
 }
 
+type SecondaryIndexSchemaValidator struct{}
```
himank, line 53: This is good and is for backward compatibility. Also, add a check in fieldBuilder `Build` method which is done when we receive the schema and build the fields.

himank (approved):

himank (approved):

PR #896, opened 2023-03-03
chore: lower value on flaky size test

garrensmith:
## Describe your changes
This test is quite flaky because size is an estimate. And at the smaller size it isn't that accurate. 
The test just needs to prove that we have some kind of size value. 

Tests should pass


rbarabas (approved):
👍

rbarabas:
Let's see if the tests will pass or... they'll be flaky!

PR #902, opened 2023-03-06
fix: only log unexpected errors in match

garrensmith:
## Describe your changes
We logging an error that is acceptable. There will be cases when the field we looking for does not exist in the
document or in the json timestamp object when we do a query. In those cases rather check first the field does not exist and return false. Only if it is an unexpected error do we log it. 

Test should pass


pboros (approved):

pboros:
I verified that this indeed fixes the logging issue.

PR #906, opened 2023-03-06
Push tigris-debug to Quay.io as well

terbepetra:
## Describe your changes
Modified the build to push debug `tigris-debug` image to Quay.io in addition to DockerHub, as we intend to serve all deployments from Quay.io

Added `debug image` label to run the `push-debug-image` job.
image

### DIFF
--- a/.github/workflows/publish-debug-image.yaml
+++ b/.github/workflows/publish-debug-image.yaml
@@ -7,6 +7,7 @@ on:
 
 env:
   DOCKER_REPOSITORY: "tigrisdata/tigris-debug"
+  QUAY_REPOSITORY: "quay.io/tigrisdata/tigris-debug"
 
 jobs:
   build-and-push-debug-image:
@@ -33,6 +34,13 @@ jobs:
           username: ${{ secrets.GH_DOCKER_ACCESS_USER }}
           password: ${{ secrets.GH_DOCKER_ACCESS_TOKEN }}
 
+      - name: Login to Quay.io
+        uses: docker/login-action@v1
+        with:
+          registry: quay.io
+          username: ${{ secrets.QUAY_REGISTRY_USER }}
+          password: ${{ secrets.QUAY_REGISTRY_PASSWORD }}
+
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@v4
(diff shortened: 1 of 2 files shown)


rbarabas:
> ## Describe your changes
> Modified `/metadata-action` and added Quay.io login in our `build-and-push-image` workflow so We will push `tigris-debug` image to Quay.io registry as well.
> 
> ## How best to test these changes
> Added `debug image` label to run the `push-debug-image` job. image

Nit: saying /metadata-action makes it sound like we have changed the Tigris API. Could we change the description to something like "Modified the build to push debug images to quay.io in addition to dockerhub, as we intend to serve all deployments from quay.io"? This would give people some background on _why_ we are doing it instead of describe the mechanics on how that is done which is visible through the PR.

.github/workflows/publish-debug-image.yaml
```diff
@@ -7,6 +7,7 @@ on:
 
 env:
   DOCKER_REPOSITORY: "tigrisdata/tigris-debug"
+  QUAY_IMAGE: "quay.io/tigrisdata/tigris-debug"
```
rbarabas, line 10: Isn't this a repository instead of an image?

rbarabas: I guess I see why we used image here from L49-L51. To avoid confusion, should we make the DOCKER_REPOSITORY variable similar (e.g. DOCKER_IMAGE) or vica versa?

.github/workflows/publish-debug-image.yaml
```diff
@@ -33,13 +34,21 @@ jobs:
...
+          password: ${{ secrets.QUAY_REGISTRY_PASSWORD }}
+
       - name: Docker meta
         id: meta
         uses: docker/metadata-action@v4
         with:
           # list of Docker images to use as base name for tags
           images: |
             ${{ env.DOCKER_REPOSITORY }}
+            ${{ env.QUAY_IMAGE }}
```
rbarabas, line 51: I'd make either the previous variable be called DOCKER_IMAGE or make this variable be called QUAY_REPOSITORY, whichever you think is more accurate. I realize that these are under the `images` spec so perhaps that would be more accurate?

terbepetra: Yes, you're right. Changing it to `QUAY_REPOSITORY`.

rbarabas (approved):

PR #917, opened 2023-03-07
Optionally disable schema validation

efirs:
Skip schema validation on write if `Tigris-Schema-Sign-Off : true` header is set.

### DIFF
--- a/server/services/v1/database/base_runner.go
+++ b/server/services/v1/database/base_runner.go
@@ -28,6 +28,7 @@ import (
 	"github.com/tigrisdata/tigris/server/config"
 	"github.com/tigrisdata/tigris/server/metadata"
 	"github.com/tigrisdata/tigris/server/metrics"
+	"github.com/tigrisdata/tigris/server/request"
 	"github.com/tigrisdata/tigris/server/transaction"
 	"github.com/tigrisdata/tigris/server/types"
 	"github.com/tigrisdata/tigris/store/search"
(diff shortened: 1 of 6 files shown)


server/services/v1/database/base_runner.go
```diff
@@ -183,9 +184,13 @@ func (runner *BaseQueryRunner) mutateAndValidatePayload(coll *schema.DefaultColl
 		return doc, err
 	}
 
-	if err = coll.Validate(deserializedDoc); err != nil {
-		// schema validation failed
-		return doc, err
+	needValidation := api.GetHeader(ctx, api.HeaderSchemaSignOff) != "true"
```
himank, line 187: Can you move this in a helper method in `request.go`?

server/services/v1/database/base_runner.go
```diff
@@ -168,7 +169,7 @@ func (runner *BaseQueryRunner) insertOrReplace(ctx context.Context, tx transacti
 	return ts, allKeys, err
 }
 
-func (runner *BaseQueryRunner) mutateAndValidatePayload(coll *schema.DefaultCollection, mutator mutator, doc []byte) ([]byte, error) {
+func (runner *BaseQueryRunner) mutateAndValidatePayload(ctx context.Context, coll *schema.DefaultCollection, mutator mutator, doc []byte) ([]byte, error) {
 	deserializedDoc, err := util.JSONToMap(doc)
```
himank, line 173: I am just thinking how to avoid this in case schema sign off is true, this is very expensive...(not for this diff though)

server/services/v1/database/base_runner.go
himank: By the way do we also send user-agent from client i.e. from which client request is issued?

server/services/v1/database/base_runner.go
efirs: If we remember that collection doesn't have defaults then we can do it.

himank: yes, that info we have - the other thing is stringsToInt64 for JS.

himank (approved):

PR #921, opened 2023-03-08
Account updated and deleted bytes towards quota

efirs:

### DIFF
--- a/server/services/v1/database/query_runner.go
+++ b/server/services/v1/database/query_runner.go
@@ -30,6 +30,7 @@ import (
 	"github.com/tigrisdata/tigris/server/config"
 	"github.com/tigrisdata/tigris/server/metadata"
 	"github.com/tigrisdata/tigris/server/metrics"
+	"github.com/tigrisdata/tigris/server/request"
 	"github.com/tigrisdata/tigris/server/transaction"
 	"github.com/tigrisdata/tigris/store/kv"
 	"github.com/tigrisdata/tigris/util"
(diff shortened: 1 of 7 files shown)


server/services/v1/database/query_runner.go
```diff
@@ -351,6 +355,8 @@ func (runner *DeleteQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 			}
 		}
 
+		request.IncrementWrittenBytes(ctx, len(row.Data.RawData))
```
ovaistariq, line 358: If I delete a document of size 2KB, will this cause it to be counted as two write units?

server/services/v1/database/query_runner.go
```diff
@@ -247,7 +248,8 @@ func (runner *UpdateQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 
 		newData := internal.NewTableDataWithTS(row.Data.CreatedAt, ts, merged)
 		newData.SetVersion(coll.GetVersion())
-		// as we have merged the data, it is safe to call replace
+
+		request.IncrementWrittenBytes(ctx, len(merged))
```
himank, line 252: We need to update it for read queries that are scanning a lot of documents.

ovaistariq: Yes, we will need to do two things here for the Update:
- count the number of bytes read to fetch the data to be updated (read units consumed)
- count the amount of data written back (write units consumed)

server/services/v1/database/query_runner.go
ovaistariq: Yes, we will need to do two things here for the Delete:
- count the number of bytes read to fetch the data to be deleted if we need to read. If it is a blind delete, then there is nothing read, so zero bytes (read units consumed)
- count the number of documents deleted as 1 write each if we are doing individual deletes
- if it is a blind delete then it is just 1 operation

ovaistariq:
Is this ready? Could you also document the change in the PR description.

efirs:
This is ready for update and delete, implemented on query_runner level. It's missing read part though.

@pboros suggested alternative approach using lower level, metrics KV wrapper. He is working on read part of it. If that approach is easier then we'll adopt it instead of this diff for the update and deletes too.

ovaistariq:
Sounds good

himank:
Closing as we don't need it.

PR #924, opened 2023-03-09
feat: Add background queue api

### LINKED ISSUE #910: Implement background queue api
Build queue api to add and remove items. This is the base of a background worker system to build secondary indexes in the background.

garrensmith:
## Describe your changes
Adds a background queue table and api to enqueue, peak, dequeue and complete items in the queue.
The queue using vesting time which is current time plus delay time to hide items that are not ready to be worked on or are currently being worked on.

Tests should all pass

Closes #910

### DIFF
--- a/server/metadata/queue.go
+++ b/server/metadata/queue.go
@@ -1,0 +1,241 @@
+// Copyright 2022-2023 Tigris Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metadata
+
+import (
+	"context"
+	"time"
+
+	jsoniter "github.com/json-iterator/go"
+	"github.com/rs/zerolog/log"
+	"github.com/tigrisdata/tigris/errors"
+	"github.com/tigrisdata/tigris/internal"
+	"github.com/tigrisdata/tigris/keys"
+	"github.com/tigrisdata/tigris/lib/uuid"
+	"github.com/tigrisdata/tigris/server/transaction"
+	"github.com/tigrisdata/tigris/store/kv"
+	ulog "github.com/tigrisdata/tigris/util/log"
+)
+
+const (
+	queueMetaValueVersion int32  = 1
+	itemsSpace            int64  = 1
+	maxId                 string = "$MAX"
+	maxPriority                  = 10
+)
+
+type QueueSubspace struct {
+	metadataSubspace
+}
+
+type QueueItem struct {
+	Id          string `json:"id"`
+	Priority    int64  `json:"priority"`
+	ErrorCount  uint8  `json:"error_count"`
+	LeaseId     string `json:"lease_id,omitempty"`
+	Data        []byte `json:"data"`
+	VestingTime int64  `json:"vesting_time"`
+}
+
+func NewQueueItem(priority int64, data []byte) *QueueItem {
+	return &QueueItem{
+		Id:         uuid.NewUUIDAsString(),
+		Priority:   priority,
+		ErrorCount: 0,
+		Data:       data,
+	}
+}
+
+func newQueueStore(nameRegistry *NameRegistry) *QueueSubspace {
+	return &QueueSubspace{
+		metadataSubspace{
+			SubspaceName: nameRegistry.QueueSubspaceName(),
+			KeyVersion:   []byte{byte(queueMetaValueVersion)},
+		},
+	}
+}
+
+func (q *QueueSubspace) getKey(vestingTime int64, priority int64, id string) keys.Key {
+	if id == maxId {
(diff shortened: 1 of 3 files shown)


server/metadata/queue.go
```diff
@@ -0,0 +1,241 @@
...
+	item.VestingTime = time.Now().Add(delay).UnixMilli()
+	key := q.getKey(item.VestingTime, item.Priority, item.Id)
+	tableData, err := q.encodeItem(item)
+	if err != nil {
+		return err
+	}
+	return tx.Replace(ctx, key, tableData, false)
+}
+
+func (q *QueueSubspace) Peak(ctx context.Context, tx transaction.Tx, max int) ([]QueueItem, error) {
```
efirs, line 91: Should this be "peek" instead?

server/metadata/queue.go
```diff
@@ -0,0 +1,241 @@
...
+	metadataSubspace
+}
+
+type QueueItem struct {
+	Id          string `json:"id"`
+	Priority    int64  `json:"priority"`
+	ErrorCount  uint8  `json:"error_count"`
+	LeaseId     string `json:"lease_id,omitempty"`
+	Data        []byte `json:"data"`
+	VestingTime int64  `json:"vesting_time"`
```
efirs, line 49: The type can time.Time, then we don't need `Time` suffix in the variable name.

server/metadata/queue.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	if item == nil {
+		return nil, errors.Internal("Queue Item was not found")
+	}
+
+	if err = q.Dequeue(ctx, tx, item); err != nil {
+		return nil, err
+	}
+
+	item.LeaseId = uuid.NewUUIDAsString()
```
himank, line 157: Ideally this should be passed by the owner?

server/metadata/queue.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (q *QueueSubspace) RenewLease(ctx context.Context, tx transaction.Tx, item *QueueItem, leaseTime time.Duration) error {
+	found, err := q.Find(ctx, tx, item)
+	if err != nil {
+		return err
+	}
+
+	if found.LeaseId != item.LeaseId {
+		return errors.Internal("another worker has claimed this item")
```
himank, line 185: So these errors doesn't need to be HTTP errors, we can have some errors defined for this class which can be used in this case. For example, this error is just throw away error i.e. if item is picked by some other error then ignore the item and continue. But in some cases the errors will be critical and we may need additional actions.

server/metadata/queue.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	return keys.NewKey(q.SubspaceName, q.KeyVersion, itemsSpace, vestingTime, priority, id)
+}
+
+func (q *QueueSubspace) Enqueue(ctx context.Context, tx transaction.Tx, item *QueueItem, delay time.Duration) error {
+	if len(item.Id) == 0 {
+		return errors.Internal("cannot enqueue without an id")
+	}
+	item.VestingTime = time.Now().Add(delay).UnixMilli()
```
himank, line 82: I am assuming the only advantage of vesting time is to reduce the contention between the queue adding items and workers picking it up?

server/metadata/queue.go
garrensmith: I would like to leave as is for now. I might change it to the owner once I implement the workers part.

server/metadata/queue.go
garrensmith: Vesting time means an item is never out of the queue and so it is never lost. When a user works on an item, the worker adds the vesting time - it could also be called an expired time at that point. And then if the item is not completed by the time vesting time == current time. Another worker can take over the task.

server/metadata/queue.go
```diff
@@ -0,0 +1,241 @@
...
+	key := q.getKey(item.VestingTime, item.Priority, item.Id)
+	tableData, err := q.encodeItem(item)
+	if err != nil {
+		return err
+	}
+	return tx.Replace(ctx, key, tableData, false)
```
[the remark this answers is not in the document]
garrensmith: Yes. We shouldn't have duplicates.

server/metadata/queue.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/tigrisdata/tigris/server/transaction"
+	"github.com/tigrisdata/tigris/store/kv"
+	ulog "github.com/tigrisdata/tigris/util/log"
+)
+
+const (
+	queueMetaValueVersion int32  = 1
+	itemsSpace            int64  = 1
+	maxId                 string = "$MAX"
+	maxPriority                  = 10
```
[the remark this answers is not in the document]
garrensmith: No. But I wanted a high number we would not reach

server/metadata/queue.go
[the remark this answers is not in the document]
himank: Also, do we need to cleanup the existing item if it is present?

server/metadata/queue.go
himank: Okay - the other thing is that as it is the same item and must be the same worker renewing the lease do we need to change the lease id?

server/metadata/queue.go
himank: Okay, now I see. So the worst case an item is not picked is the delay set by the worker.

server/metadata/queue.go
garrensmith: No the lease id should stay the same while a worker works on it.

himank: I thought it is renewLease, nvm.

PR #930, opened 2023-03-12
feat: Support search indexes on array of objects and free flow objects

himank:
## Describe your changes
This extends the search with following features,
- Search on array of objects (should be on top level)
- Search on free flow objects i.e. the objects with no schema (should be on top level)
- Search on primitive arrays is already supported. Primitive arrays can be on any level. This diff is adding some more tests related to it.

The following are the examples to enable this functionality,

- To enable a search index on an object without schema,

```
{
  "schema": {
    "title": "collection_a",
    "properties": {
      "field_a": {
        "searchIndex": true,
        "type": "object",
        "properties": {}
      }
    }
  }
}
```

- To enable a search index on an array of objects,
```      
{
  "schema": {
    "title": "collection_a",
    "properties": {
      "field_a": {
        "searchIndex": true,
        "type": "array",
        "items": {
          "type": "object",
          "properties": {}
        }
      }
    }
  }
}
```

- To enable a search index on an array of objects(object with schema),
```
{
  "schema": {
    "title": "collection_a",
    "properties": {
      "field_a": {
        "searchIndex": true,
        "type": "array",
        "items": {
          "type": "object",
          "properties": {
            "nested": {
              "type": "string"
            }
          }
        }
      }
    }
  }
}
```
 
 Note: `"searchIndex": true` is not needed for search indexes as default behavior is true. This is explicitly needed in case of collection schema. 
 
#931


schema/fields.go
```diff
@@ -337,12 +341,12 @@ type FieldBuilder struct {
...
 
-func (f *FieldBuilder) Build(forSearchIndex bool) (*Field, error) {
+func (f *FieldBuilder) Build(setSearchDefaults bool) (*Field, error) {
 	fieldType := ToFieldType(f.Type, f.Encoding, f.Format)
-	if forSearchIndex {
+	if setSearchDefaults {
 		// for search indexes, any field in schema is search indexable if it is not set explicitly.
 		// Similarly, we also tag it with sort if it is numeric.
-		if len(f.FieldName) > 0 && fieldType != ObjectType {
+		if len(f.FieldName) > 0 && (fieldType != ObjectType || (fieldType == ObjectType && len(f.Fields) == 0)) {
```
garrensmith, line 349: Can you move `(fieldType != ObjectType || (fieldType == ObjectType && len(f.Fields) == 0))` into its own function. Its a bit unwieldy to read

garrensmith (approved):
One small nit otherwise it looks good.

schema/fields.go
himank: sure

ovaistariq:
@himank could you add snippets of schema examples as well demonstrating this.

ovaistariq (dismissed):

efirs (approved):

PR #935, opened 2023-03-16
feat: Implement chunker to split/merge the data to support bigger payloads

himank:
## Describe your changes
FoundationDB only allows a maximum value size of 100KB. This diff is adding the functionality to have a layer that splits the value into multiple chunks and then store it in the underlying engine and similarly during reads merge these values back and return a single payload to the caller. This functionality is not tied to FDB directly but provide a generic mechanism to chunk and merge. 

This is not currently enabled. The following diffs will hook this class to actually start using it. The first step will be to do it for search indexes.

### DIFF
--- a/internal/data.go
+++ b/internal/data.go
@@ -53,6 +53,8 @@ const (
 	JsonEncoding    UserDataEncType = 2
 )
 
+var EmptyData = &TableData{}
+
 // CreateNewTimestamp is a method used to construct timestamp from unixNano. In search backend we store internal
 // timestamps as unixNano.
 func CreateNewTimestamp(nano int64) *Timestamp {
@@ -112,6 +114,21 @@ func NewTableDataWithVersion(data []byte, version int32) *TableData {
 	}
 }
 
+func (x *TableData) CloneWithAttributesOnly(newRawData []byte) *TableData {
+	return &TableData{
+		Ver:         x.Ver,
+		Encoding:    x.Encoding,
+		CreatedAt:   x.CreatedAt,
+		UpdatedAt:   x.UpdatedAt,
+		TotalChunks: x.TotalChunks,
+		RawData:     newRawData,
+	}
+}
+
+func (x *TableData) IsChunkedData() bool {
+	return x.TotalChunks != nil && *x.TotalChunks > 1
+}
+
 func (x *TableData) SetVersion(ver int32) {
 	x.Ver = ver
 }
@@ -147,6 +164,17 @@ func (x *TableData) TimestampsToJSON() ([]byte, error) {
 	return jsoniter.Marshal(data)
 }
 
+// ActualUserPayloadSize returns size of the user data. This is used by splitter to split the value if it is
+// greater than the maximum allowed by the underlying storage.
+func (x *TableData) ActualUserPayloadSize() int32 {
+	return int32(len(x.RawData))
+}
+
+func (x *TableData) SetTotalChunks(chunkSize int32) {
+	totalChunks := (x.ActualUserPayloadSize() + chunkSize - 1) / chunkSize
+	x.TotalChunks = &totalChunks
+}
+
 // Encode is used to encode data to the raw bytes which is used to store in storage as value. The first byte is storing
 // the type corresponding to this Data. This is important and used by the decoder later to decode back.
 func Encode(data *TableData) ([]byte, error) {
@@ -174,6 +202,7 @@ func Decode(b []byte) (*TableData, error) {
 	if len(b) == 0 {
 		return nil, errors.Internal("unable to decode table data is empty")
 	}
+
 	dataType := DataType(b[0])
 	return decodeInternal(dataType, b[1:])
 }
--- a/server/services/v1/database/secondary_indexer.go
+++ b/server/services/v1/database/secondary_indexer.go
@@ -192,7 +192,7 @@ func (q *SecondaryIndexer) Update(ctx context.Context, tx transaction.Tx, newTd 
 	}
 
 	for _, indexKey := range updateSet.addKeys {
-		if err := tx.Replace(ctx, indexKey, nil, false); err != nil {
+		if err := tx.Replace(ctx, indexKey, internal.EmptyData, false); err != nil {
 			return err
 		}
 	}
--- a/server/transaction/manager.go
+++ b/server/transaction/manager.go
@@ -41,7 +41,6 @@ type BaseTx interface {
 	GetTxCtx() *api.TransactionCtx
 	Insert(ctx context.Context, key keys.Key, data *internal.TableData) error
 	Replace(ctx context.Context, key keys.Key, data *internal.TableData, isUpdate bool) error
-	Update(ctx context.Context, key keys.Key, apply func(*internal.TableData) (*internal.TableData, error)) (int32, error)
 	Delete(ctx context.Context, key keys.Key) error
 	Read(ctx context.Context, key keys.Key) (kv.Iterator, error)
 	ReadRange(ctx context.Context, lKey keys.Key, rKey keys.Key, isSnapshot bool) (kv.Iterator, error)
@@ -50,7 +49,6 @@ type BaseTx interface {
 	SetVersionstampedKey(ctx context.Context, key []byte, value []byte) error
 	AtomicAdd(ctx context.Context, key keys.Key, value int64) error
 	AtomicRead(ctx context.Context, key keys.Key) (int64, error)
-	AtomicReadRange(ctx context.Context, lKey keys.Key, rKey keys.Key, isSnapshot bool) (kv.AtomicIterator, error)
 	RangeSize(ctx context.Context, table []byte, lKey keys.Key, rKey keys.Key) (size int64, err error)
 }
 
@@ -151,6 +149,7 @@ func (s *TxSession) start(ctx context.Context) error {
 	if s.kTx, err = s.kvStore.BeginTx(ctx); err != nil {
 		return err
 	}
+
 	s.state = sessionActive
 
 	return nil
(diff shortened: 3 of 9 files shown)


internal/data.go
```diff
@@ -90,28 +92,62 @@ func (x *Timestamp) UnixNano() int64 {
 
 // NewTableData returns a table data type by setting the ts to the current value.
 func NewTableData(data []byte) *TableData {
+	rawData := int32(len(data))
```
garrensmith, line 95: A nice longer term refactor here is to have one function:

```

func newTableDataFull(data []byte, createdAt *Timestamp, updatedAt *Timestamp, version int32) *TableData {
rawData := int32(len(data))
	return &TableData{
		CreatedAt: createdAt,
		UpdatedAt: updatedAt,
		RawData:   data,
		RawSize:   &rawData,
	}

} 

```
And then the other table data functions call that function to do the actual creation of the TableData.

server/services/v1/database/secondary_indexer.go
```diff
@@ -192,7 +192,7 @@ func (q *SecondaryIndexer) Update(ctx context.Context, tx transaction.Tx, newTd
 	}
 
 	for _, indexKey := range updateSet.addKeys {
-		if err := tx.Replace(ctx, indexKey, nil, false); err != nil {
+		if err := tx.Replace(ctx, indexKey, internal.EmptyData, false); err != nil {
```
garrensmith, line 195: Nice fix

server/transaction/chunk.go
```diff
@@ -0,0 +1,216 @@
...
+	chunkIdentifier = "__"
+)
+
+var chunkSize = KB99
+
+type chunkCB func(chunkNo int32, chunkData []byte) error
+
+// ChunkTx will be used to split the user payload if it is greater than 99KB. It adds some metadata in the zeroth chunk
+// so that it can rebuild the body again. The chunk number is only appended in the first chunk as we don't need to do
+// anything for zeroth chunk. The chunk number is appended at the end in the format "__<chunk number>".
```
garrensmith, line 40: I'm wondering if we should call it first chunk over `zeroth` chunck

server/transaction/chunk.go
```diff
@@ -0,0 +1,216 @@
...
+
+		if err := tx.Tx.Replace(ctx, key.Table(), kv.BuildKey(keyParts...), chunked, isUpdate); err != nil {
+			return err
+		}
+
+		return nil
+	})
+}
+
+// Delete doesn't need any special handling as it is deleting on the key level not on chunk level.
```
garrensmith, line 138: Just to confirm, the delete is going to delete a range based on the `PrimaryKey` so any chunk will be part of that range?

server/transaction/manager.go
```diff
@@ -79,18 +77,20 @@ func (c *SessionCtx) GetStagedDatabase() interface{} {
...
-	kvStore kv.KeyValueStore
+	kvStore  kv.KeyValueStore
+	chunking bool
 }
 
 func NewManager(kvStore kv.KeyValueStore) *Manager {
 	return &Manager{
-		kvStore: kvStore,
+		kvStore:  kvStore,
+		chunking: false,
```
garrensmith, line 87: I think we should make this a config option so we can run it with our tests and on dev

store/kv/fdb.go
```diff
@@ -416,6 +416,7 @@ func (t *ftx) Read(_ context.Context, table []byte, key Key) (baseIterator, erro
 		return nil, err
 	}
 
+	// the iterator should be change if we need to support reverse order because the keys may be chunked by the caller.
```
garrensmith, line 419: Sorry this comment doesn't quite make sense.

garrensmith (changes requested):
Looking good. This will be a nice improvement. A few comments and questions.

server/transaction/chunk.go
himank: that's right

server/transaction/chunk.go
himank: :) yeah, either way works. But mainly wanted to avoid confusion that the chunk number will not be `1` for this chunk, either it would be `nil` or `0`

server/transaction/manager.go
himank: that's the plan, we need to pass in the manager - that will be part of follow up diff when I'll hook it up with the API layer.

store/kv/fdb.go
himank: let me expand it, but basically we can't support reverse range by default as the chunker will be relying on some ordering to perform the merge. I have added a note here so that we don't forget to return a different iterator in that case.

internal/data.go
himank: sure

server/transaction/manager.go
```diff
@@ -41,7 +41,6 @@ type BaseTx interface {
 	GetTxCtx() *api.TransactionCtx
 	Insert(ctx context.Context, key keys.Key, data *internal.TableData) error
 	Replace(ctx context.Context, key keys.Key, data *internal.TableData, isUpdate bool) error
-	Update(ctx context.Context, key keys.Key, apply func(*internal.TableData) (*internal.TableData, error)) (int32, error)
```
himank, line 44: no usage of explicit Update, API layer reads and then merges.

ovaistariq (dismissed):

store/kv/chunk.go
```diff
@@ -0,0 +1,235 @@
...
+
+	"github.com/tigrisdata/tigris/internal"
+	"github.com/tigrisdata/tigris/server/config"
+)
+
+const (
+	KB   = 1000
+	KB99 = 99 * KB
+
+	chunkIdentifier = "__"
```
efirs, line 30: maybe something more human friendly, like `__chunk_`

store/kv/chunk.go
```diff
@@ -0,0 +1,235 @@
...
+
+	var buf bytes.Buffer
+	buf.Write(value.Data.RawData)
+	hasNext := false
+	for chunk := int32(1); chunk < *value.Data.TotalChunks; chunk++ {
+		var chunked KeyValue
+		if hasNext = it.Iterator.Next(&chunked); !hasNext {
+			break
+		}
+		if len(chunked.Key) < 2 {
```
efirs, line 211: move key check to the separate function.

store/kv/chunk.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+
+		if chunkNo, ok := chunked.Key[len(chunked.Key)-1].(int64); !ok || chunk != int32(chunkNo) {
+			it.err = fmt.Errorf("chunk number mismatch")
+			return false
+		}
+
+		buf.Write(chunked.Data.RawData)
+	}
+
```
efirs, line 228: You can check that number of chunk read is equal to the number of chunk in the first header.

store/kv/chunk_test.go
```diff
@@ -0,0 +1,171 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/tigrisdata/tigris/internal"
+	"github.com/tigrisdata/tigris/server/config"
+)
+
+func TestChunkStore(t *testing.T) {
```
efirs, line 26: Would be good to add some negative use cases.

store/kv/chunk.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// Read needs to return chunk iterator so that it can merge and returned merged chunk to caller.
+func (tx *ChunkTx) Read(ctx context.Context, table []byte, key Key) (Iterator, error) {
+	iterator, err := tx.KeyValueTx.Read(ctx, table, key)
+	if err != nil {
+		return nil, err
+	}
+
+	return &ChunkIterator{
```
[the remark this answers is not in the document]
himank: Hmm, it depends if it is returning an object.

store/kv/chunk.go
himank: I think it is small check, I won't add a function call overhead in this case.

store/kv/chunk.go
```diff
@@ -0,0 +1,235 @@
...
+	originalDoc := data.RawData
+	start := 0
+	chunk := int32(0)
+	for start < len(originalDoc) {
+		end := start + chunkSize
+		if end > len(originalDoc) {
+			end = len(originalDoc)
+		}
+
+		if err := cb(chunk, originalDoc[start:end]); err != nil {
```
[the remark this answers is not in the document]
himank: I think it probably would be more readable to have it separate.

store/kv/chunk.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &ChunkTx{
+		KeyValueTx: btx.(*KeyValueTx),
+	}, nil
+}
+
+func (tx *ChunkTx) setTotalChunks(data *internal.TableData) {
```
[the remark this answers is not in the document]
himank: yes, this doesn't belong here.

store/kv/chunk.go
himank: It will be added to all the chunked keys, better to optimize it on what we are adding.

store/kv/chunk.go
efirs: Compiler will inline if needed.
Function name will self document and simplify understanding.

himank: I mean it is single place usage why do need a separate function for this?

efirs: Just too many IFs, if we enable cyclomatic complexity linter it'll complain.
Also:
```
for ; chunk < *value.Data.TotalChunks; chunk++ {
		var chunked KeyValue
		if hasNext = it.Iterator.Next(&chunked); !hasNext {
			break
		}
 
        if !isKeyValid(chunked.Key, chunkNo) {
          return err
        }

		buf.Write(chunked.Data.RawData)
}
```
more readable.

himank: OK

efirs (approved):

PR #939, opened 2023-03-16
Auth caches refactor

JigarJoshi:
## Describe your changes

- Disabled token cache for gotrue implementation of oauth provider.
- Replaced LRU token validation cache with LRU+expiring token validation cache.
- Added a special header that allows us to bypass token validation cache.
- Added test that exercises token validation with and without cache.

Added tests.

### DIFF
--- a/api/server/v1/header.go
+++ b/api/server/v1/header.go
@@ -37,10 +37,11 @@ const (
 
 	HeaderPrefix = "Tigris-"
 
-	HeaderTxID          = "Tigris-Tx-Id"
-	HeaderTxOrigin      = "Tigris-Tx-Origin"
-	grpcGatewayPrefix   = "Grpc-Gateway-"
-	HeaderSchemaSignOff = "Tigris-Schema-Sign-Off"
+	HeaderTxID            = "Tigris-Tx-Id"
+	HeaderTxOrigin        = "Tigris-Tx-Origin"
+	grpcGatewayPrefix     = "Grpc-Gateway-"
+	HeaderSchemaSignOff   = "Tigris-Schema-Sign-Off"
+	HeaderBypassAuthCache = "Tigris-Bypass-Auth-Cache" // #nosec G101
 )
 
 func CustomMatcher(key string) (string, bool) {
--- a/server/middleware/auth.go
+++ b/server/middleware/auth.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/auth0/go-jwt-middleware/v2/jwks"
 	"github.com/auth0/go-jwt-middleware/v2/validator"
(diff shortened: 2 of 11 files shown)


server/middleware/auth.go
```diff
@@ -162,13 +166,18 @@ func authFunction(ctx context.Context, jwtValidators []*validator.Validator, con
 		return ctx, err
 	}
 
-	validatedToken, found := cache.Get(tkn)
-	log.Debug().Bool("validated token found", found).Interface("token", validatedToken)
-	if !found {
+	var validatedToken interface{}
+	if !BypassAuthCaches(ctx) {
```
himank, line 170: Can we move this logic in a separate function, then it would be easy to test whether a token is returned from the cache or not when the header is set by adding the token in the cache and then setting the header. 

```
validatedToken := getFromCache()
```

server/middleware/auth.go
```diff
@@ -162,13 +166,18 @@ func authFunction(ctx context.Context, jwtValidators []*validator.Validator, con
...
+		validatedToken, err = cache.Get(tkn)
+		if err != nil {
```
himank, line 172: if an item doesn't exist then does cache throws an error? Or only in case of legit errors?

api/server/v1/header.go
```diff
@@ -37,10 +37,11 @@ const (
...
 
-	HeaderTxID          = "Tigris-Tx-Id"
-	HeaderTxOrigin      = "Tigris-Tx-Origin"
-	grpcGatewayPrefix   = "Grpc-Gateway-"
-	HeaderSchemaSignOff = "Tigris-Schema-Sign-Off"
+	HeaderTxID             = "Tigris-Tx-Id"
+	HeaderTxOrigin         = "Tigris-Tx-Origin"
+	grpcGatewayPrefix      = "Grpc-Gateway-"
+	HeaderSchemaSignOff    = "Tigris-Schema-Sign-Off"
+	HeaderBypassAuthCaches = "Tigris-Bypass-Auth-Caches" // #nosec G101
```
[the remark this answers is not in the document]
JigarJoshi: renamed.

server/middleware/auth.go
JigarJoshi: Yes it returns an error if item doesn't exists in cache. I refactored this code as per your above comment and this is gone now.

server/middleware/auth.go
JigarJoshi: added test for this. Thanks

himank (dismissed):

himank (approved):

himank (approved):

PR #948, opened 2023-03-21
feat: Manage secondary indexes via metadata

### LINKED ISSUE #912: Store index metadata
Store the information around indexes in the metadata. 
Specifically, store which indexes are ready to be used for queries and which are in write mode only

garrensmith:
## Describe your changes

Manage secondary indexes via index metadata. All secondary indexes state are now tracked in the index metadata. The index can be in an `ACTIVE` state, meaning it can be used for querying. 

**One thing** I'm not sure about at the moment is that all secondary index information is stored under the encoding subspace. That should be ok. Changing it means we need to have two `indexStores`, one for the primary key and one for the secondary indexes

All current tests should pass

closes #912

### DIFF
--- a/schema/collection.go
+++ b/schema/collection.go
@@ -193,6 +193,12 @@ func (d *DefaultCollection) GetName() string {
 	return d.Name
 }
 
+// The subspace within a collection where the secondary index information
+// is stored.
+func (d *DefaultCollection) SecondaryIndexName() string {
+	return "skey"
+}
+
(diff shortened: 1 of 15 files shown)


himank:
> **One thing** I'm not sure about at the moment is that all secondary index information is stored under the encoding subspace. That should be ok. Changing it means we need to have two `indexStores`, one for the primary key and one for the secondary indexes
>

What if we store it as part of the collection metadata?

PR #949, opened 2023-03-21
feat: Support symmetric key token signature algorithm

JigarJoshi:
## Describe your changes
This PR adds support for symmetric key algorithm to verify token signature. This is required to decouple circular dependency amongst internal services.

Note: This PR changes the shape of config and needs attention with this release deployment.

Added test case for this.


docker/Dockerfile
```diff
@@ -36,8 +36,10 @@ RUN groupadd -r tigris && useradd -r -s /bin/false -g tigris tigris
 
 RUN mkdir -p /server /etc/tigrisdata/tigris /etc/foundationdb/
 
+ARG BUILD_PROFILE=""
```
rbarabas, line 39: Isn't there already an `env` based mechanism for this somewhere?

JigarJoshi: We need this at build time. Environment variables are available in container at runtime. For that reason we need build arg (i.e. build time env vars).

Let me know if I missed any existing mechanism to handle this.

server/middleware/auth.go
```diff
@@ -97,14 +97,33 @@ func BypassAuthCaches(ctx context.Context) bool {
...
+	for i, issuerCfg := range config.Auth.Issuers {
+		log.Info().Msg(issuerCfg.Issuer)
+		issuerURL, _ := url.Parse(issuerCfg.Issuer)
+		var keyFunc func(ctx context.Context) (interface{}, error)
+		switch issuerCfg.Algorithm {
+		case validator.RS256:
+			{
+				provider := jwks.NewCachingProvider(issuerURL, config.Auth.JWKSCacheTimeout)
+				keyFunc = provider.KeyFunc
+				break
```
himank, line 109: nit: You do not need braces and break in go inside cases.

server/middleware/auth.go
```diff
@@ -97,14 +97,33 @@ func BypassAuthCaches(ctx context.Context) bool {
...
+		case validator.HS256:
+			{
+				keyFunc = func(ctx context.Context) (interface{}, error) {
+					return []byte(issuerCfg.SharedSecret), nil
+				}
+				break
+			}
+		default:
+			{
+				panic("Unsupported Token signature algorithm")
```
himank, line 120: do we need to panic? Or return an error from here?

JigarJoshi: This is the code path that will get run during startup and not per request - so panic here is fine.

server/middleware/auth.go
JigarJoshi: made update.

server/config/config.go
```diff
@@ -86,6 +87,12 @@ func LoadConfig(config interface{}) {
 
 	// Deactivates lookup process for default server.yaml if
 	// configuration file is specified via the command line
+	if configFile != "" {
```
efirs, line 90:
```
if configFile == "" {
   configFile = os.Getenv(EnvTigrisConfigFile)
}
```

server/middleware/auth.go
himank: Okay, I thought it is in middleware and will be execute on per request.

himank (approved):

PR #954, opened 2023-03-22
fix: Reload version in case of index not found

himank:
## Describe your changes
Reload metadata version in case of search index not found

### DIFF
--- a/server/services/v1/search/session.go
+++ b/server/services/v1/search/session.go
@@ -64,8 +64,23 @@ func (sessions *SessionManager) Execute(ctx context.Context, runner Runner) (Res
 		return Response{}, errors.NotFound("tenant '%s' not found", namespace)
 	}
 
-	resp, err := runner.Run(ctx, tenant)
+	resp, err := sessions.execute(ctx, tenant, runner)
+	if err != nil && shouldRecheckTenantVersion(err) {
+		_ = sessions.TrackVersion(ctx, tenant)
+
+		resp, err = sessions.execute(ctx, tenant, runner)
+	}
+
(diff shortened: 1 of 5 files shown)


server/services/v1/search/session.go
```diff
@@ -64,10 +64,25 @@ func (sessions *SessionManager) Execute(ctx context.Context, runner Runner) (Res
 		return Response{}, errors.NotFound("tenant '%s' not found", namespace)
 	}
 
-	resp, err := runner.Run(ctx, tenant)
+	resp, err := sessions.execute(ctx, tenant, runner)
+	if err != nil && shouldRecheckTenantVersion(err) {
+		_ = sessions.TrackVersion(ctx, tenant)
+
+		resp, err = sessions.execute(ctx, tenant, runner)
```
efirs, line 71: Can this again return `metadata.ErrCodeProjectNotFound, metadata.ErrCodeSearchIndexNotFound` ?

himank: that's right and the `createApiError` at line 74 will wrap it up in user facing error.

adilansari (dismissed):

server/services/v1/search/session.go
efirs: Then should we run this in the loop, something like:
```
do {
 resp, err := sessions.execute(ctx, tenant, runner)
} while	(err != nil && shouldRecheckTenantVersion(err))
```

himank: We only need to check & reload once.

efirs (approved):

efirs (approved):

PR #958, opened 2023-03-23
fix: Fix unflatting object util method

JigarJoshi:
## Describe your changes
While unflatting the object, if the input map contains structure like this

"one" --> nil
"one.two" --> "v1"

then the method tries to access `nil` as the `map` and tries to insert ("two" in this case) into it and results into the error. This PR fixes that case. 

Added test case.


rbarabas (dismissed):
❤️

himank (approved):

PR #963, opened 2023-03-24
Allow maps in schema

efirs:
Allow sub-objects with arbitrary properties.

### DIFF
--- a/schema/collection.go
+++ b/schema/collection.go
@@ -121,8 +121,11 @@ func disableAdditionalPropertiesAndAllowNullable(required []string, properties m
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


schema/collection.go
```diff
@@ -122,7 +122,9 @@ func disableAdditionalPropertiesAndAllowNullable(required []string, properties m
 		}
 
 		if len(p.Properties) > 0 {
-			p.AdditionalProperties = false
+			if p.AdditionalProperties == nil {
```
himank, line 125: What about honoring this at line `147` as well? Then if it is passed we honor it top level as well?

himank (approved):

PR #972, opened 2023-03-28
Allow object to map conversion

efirs:
Consider structured object to map conversion compatible schema change, when additionalProperties explicitly set in the map.


ovaistariq (dismissed):

schema/fields.go
```diff
@@ -413,6 +415,8 @@ func (f *FieldBuilder) Build(setSearchDefaults bool) (*Field, error) {
 		SearchIndexed:   f.SearchIndex,
 		PrimaryKeyField: f.Primary,
 		AutoGenerated:   f.Auto,
+
```
himank, line 418: remove line?

himank (approved):

PR #974, opened 2023-03-28
feat: Manage secondary indexes via metadata

### LINKED ISSUE #912: Store index metadata

garrensmith:
## Describe your changes

Manage the state of secondary indexes in the collections metadata. Secondary indexes can be in 4 states:
    1. UNKNOWN = Have not fetched the index metadata yet so the state is unknown
	2. NOT_INDEXED = field is not indexed
	3. INDEX_WRITE_MODE = index is being built in the background and cannot be used for queries
	4. INDEX_ACTIVE = index can be used for queries

This allows Tigris to only use active and up to date indexes for querying.

There is also code clean up to differentiate between the primary key and the secondary index. The primary key index is kept in its own metadata subspace.

All tests should pass

closes #912

### DIFF
--- a/schema/collection.go
+++ b/schema/collection.go
@@ -48,7 +48,10 @@ type DefaultCollection struct {
 	// Fields are derived from the user schema.
 	Fields []*Field
 	// Indexes is a wrapper on the indexes part of this collection.
-	Indexes *Indexes
+	// Primary Key contains the fields used to make up the primary key
+	PrimaryKey *Index
+	// Secondary SecondaryIndexes for this collection
+	SecondaryIndexes *Indexes
 	// Validator is used to validate the JSON document. As it is expensive to create this, it is only created once
 	// during constructor of the collection.
 	Validator *jsonschema.Schema
@@ -85,6 +88,10 @@ type CollectionType string
 const (
 	DocumentsType CollectionType = "documents"
 )
+
+func (d *DefaultCollection) GetPrimaryKey() *Index {
+	return d.PrimaryKey
+}
 
 func disableAdditionalPropertiesAndAllowNullable(required []string, properties map[string]*jsonschema.Schema) {
 	for name, p := range properties {
@@ -168,7 +175,8 @@ func NewDefaultCollection(id uint32, schVer int, factory *Factory, schemas Versi
 		SchVer:                   schVer,
 		Name:                     factory.Name,
 		Fields:                   factory.Fields,
-		Indexes:                  factory.Indexes,
+		PrimaryKey:               factory.PrimaryKey,
+		SecondaryIndexes:         factory.Indexes,
 		Validator:                validator,
 		Schema:                   factory.Schema,
 		QueryableFields:          queryableFields,
@@ -198,6 +206,12 @@ func (d *DefaultCollection) GetName() string {
 	return d.Name
 }
 
+// The subspace within a collection where the secondary index information
+// is stored.
+func (d *DefaultCollection) SecondaryIndexName() string {
+	return "skey"
+}
+
 func (d *DefaultCollection) GetVersion() int32 {
 	return int32(d.SchVer)
 }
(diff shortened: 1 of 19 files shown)


server/metadata/tenant.go
```diff
@@ -1251,40 +1250,46 @@ func (tenant *Tenant) createCollection(ctx context.Context, tx transaction.Tx, d
...
-			meta, err = tenant.metaStore.CreateIndex(ctx, tx, idx.Name, tenant.namespace.Id(), database.id, c.id)
-			if err != nil {
-				return err
-			}
-			c.addIndex(idx.Name, meta)
+	primaryKeyIndex := schFactory.PrimaryKey
+	meta, ok := cHolder.primaryIdxMeta[primaryKeyIndex.Name]
+	if !ok {
+		// these are the new indexes present in the new collection
+		meta, err = tenant.metaStore.CreatePrimaryIndex(ctx, tx, primaryKeyIndex.Name, tenant.namespace.Id(), database.id, cHolder.id)
```
himank, line 1260: this should not happen now as it is backward incompatible change, let's return a bad request error that updating primary key is not allowed.

schema/collection.go
```diff
@@ -198,6 +206,12 @@ func (d *DefaultCollection) GetName() string {
 	return d.Name
 }
 
+// The subspace within a collection where the secondary index information
+// is stored.
+func (d *DefaultCollection) SecondaryIndexName() string {
```
himank, line 211: as you are renaming it can we call it `SecondaryIndexKeyword`?

schema/schema.go
```diff
@@ -127,8 +127,10 @@ type Factory struct {
 	Name string
 	// Fields are derived from the user schema.
 	Fields []*Field
-	// Indexes is a wrapper on the indexes part of this collection. At this point the dictionary encoded value is not
+	// Primary Key points to the fields used for the primary index. At this point the dictionary encoded value is not
 	// set for these indexes which is set as part of collection creation.
+	PrimaryKey *Index
+	// Indexes is a wrapper on the indexes part of this collection.
```
himank, line 133: `Indexes is a wrapper on the secondary indexes of this collection.`

schema/fields.go
```diff
@@ -276,16 +276,40 @@ var SupportedFieldProperties = container.NewHashSet(
 
 // Indexes is to wrap different index that a collection can have.
 type Indexes struct {
-	PrimaryKey     *Index
-	SecondaryIndex *Index
+	All []*Index
```
himank, line 279: So All returns all the secondary indexes, right?

server/metadata/collection.go
```diff
@@ -55,6 +57,72 @@ func (c *CollectionSubspace) getKey(nsID uint32, dbID uint32, name string) keys.
 	return keys.NewKey(c.SubspaceName, c.KeyVersion, UInt32ToByte(nsID), UInt32ToByte(dbID), collectionKey, name, keyEnd)
 }
 
+func (c *CollectionSubspace) Create(ctx context.Context, tx transaction.Tx, nsID uint32, dbID uint32, name string, id uint32, indexes []*schema.Index,
+) (*CollectionMetadata, error) {
+	for _, index := range indexes {
+		index.State = schema.INDEX_ACTIVE
```
himank, line 63: nit: add a comment that these are indexes created at collection creation time so they are set as active.

server/metadata/collection.go
```diff
@@ -55,6 +57,72 @@ func (c *CollectionSubspace) getKey(nsID uint32, dbID uint32, name string) keys.
...
+		index.IdxType = schema.SECONDARY_INDEX
```
himank, line 64: this doesn't need to be set here? Can be moved where you are creating these Indexes.

schema/fields.go
garrensmith: Yes. correct

server/metadata/tenant.go
garrensmith: Something like:
```
if !ok {
		return errors.InvalidArgument("cannot create a new primary key name")
	}
```

himank (approved):

PR #983, opened 2023-03-30
fix: Some fixes in the read API

himank:


garrensmith:
Overall looks good. Do we need to add a test for any of that?

ovaistariq (approved):

PR #984, opened 2023-03-30
feat: Add indexing to endpoints

garrensmith:
## Describe your changes
Adds an explain to the query endpoint.
Adds indexes to the describe collection endpoint

I'm not so sure about having the explain be part of the query endpoint. It will result in a response like this:
```
{"result": "explain": {"collection":"test_collection","index":"secondary_index","filter":"{\"int_value\":10}"}}}
```
I'm thinking we should change it to be its own endpoint. We could then support it like this:
```
  collection.findMany({filter...}, {explain: true}) 
```
Which then uses the `explain` endpoint or:
```
 collection.findMany({filter}).explain()
```
Which then uses the explain endpoint


himank:
I think it is a good idea, let's move explain to its own endpoint. One other thing is that it is also unary so doesn't make sense to attach it to read for now.

server/services/v1/database/query_runner.go
```diff
@@ -755,3 +757,65 @@ func (runner *StreamingQueryRunner) writeTimestamp(buf *bytes.Buffer, key string
...
+		return Response{}, ctx, err
+	}
+
+	return Response{
+		Response: explainResp,
+	}, ctx, nil
+}
+
+const (
+	SEARCH    = "search"
```
himank, line 796: Let's remove SEARCH as this will go away once we move to the secondary index.

server/services/v1/database/query_runner.go
```diff
@@ -755,3 +757,65 @@ func (runner *StreamingQueryRunner) writeTimestamp(buf *bytes.Buffer, key string
...
+	PRIMARY   = "scan"
+	SECONDARY = "secondary index"
+)
+
+func buildExplainResp(options readerOptions, coll *schema.DefaultCollection, filter []byte) (*api.ExplainResponse, error) {
```
himank, line 801: Don't we need to have one for SCAN as well, i.e., when a full table scan is triggered?

garrensmith: That is on line 819.

himank (dismissed):

himank (approved):

PR #988, opened 2023-03-31
feat: Vector search feature in search queries

himank:
## Describe your changes
Introducing the vector search. The API will remain the same i.e. Search but inside it there is an option "vector" that needs to be sent something like 
```
{
  "vector": {
        "vec_field": [-0.12, -0.23, -0.56, -0.83]
   }
}
```
where `vec_field` is the vector field.

To declare the vector field the following syntax can be used,
```
"vec_field": {"type":"array","format":"vector","dimensions":4}
```
i.e. the format is "vector" and "dimensions" is the dimension of this vector.


api/.DS_Store
efirs, line 1: remove

query/search/search.go
```diff
@@ -119,10 +120,29 @@ func (q *Query) ToSearchGroupBy() string {
 	return groupBy
 }
 
+func (q *Query) ToSearchVector() string {
+	if len(q.VectorS.VectorF) == 0 {
+		return ""
```
garrensmith, line 125: Will returning an empty string have any effect on the search query?

query/search/vector.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	var err error
+	var g VectorSearch
+	err = jsonparser.ObjectEach(input, func(k []byte, v []byte, jsonDataType jsonparser.ValueType, offset int) error {
+		if err != nil {
+			return err
+		}
+
+		if string(k) == "top_k" {
```
garrensmith, line 43: Can you make `"top_k"` a top leve var/const

go.mod
```diff
@@ -39,7 +41,7 @@ require (
 	github.com/spf13/viper v1.15.0
 	github.com/stretchr/testify v1.8.2
 	github.com/tigrisdata/tigris-client-go v1.0.0-beta.25
-	github.com/typesense/typesense-go v0.6.2
+	github.com/tigrisdata/typesense-go v0.6.2-beta.6
```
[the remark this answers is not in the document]
ovaistariq: Because their go library was last released in December, it’s behind in features and doesn’t support vector search feature for example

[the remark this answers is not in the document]
garrensmith: Ok that makes sense.

query/search/search.go
himank: The caller is checking it and not filling in the params if it is an empty string.

go.mod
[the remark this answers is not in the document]
himank: That's right, we are mainly using the forked version so that we can control their open API spec and regenerate the client. I also found out that their latest open API spec will break us as it has some incompatible changes i.e. the API changes are not tied to the server version changes so one another reason to fork it.

ovaistariq (dismissed):

efirs (approved):

adilansari (approved):

PR #993, opened 2023-03-31
feat: Namespace API list/info/describe consolidation. 

JigarJoshi:
## Describe your changes
Implemented a management API to get the namespace information and consolidated list/info/describe into single API.

Added tests for it.

### DIFF
--- a/api/server/v1/tx.go
+++ b/api/server/v1/tx.go
@@ -51,11 +51,10 @@ const (
 	DescribeDatabaseMethodName   = apiMethodPrefix + "DescribeDatabase"
 	DescribeCollectionMethodName = apiMethodPrefix + "DescribeCollection"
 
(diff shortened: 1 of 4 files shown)


efirs:
Would it be easier to add a filter to list namespaces API?

JigarJoshi:
> Would it be easier to add a filter to list namespaces API?

Yes but namespaceinfo api can extend to serve other information that we may need.

efirs:
ok. It just we have get/list/describe and it a bit inconsistent with other APIs.

ovaistariq:
Should we extend list namespaces to return additional information when an option is passed?

ovaistariq:
Or may be make the namespace info part of the describe endpoint.

JigarJoshi:
we internally had discussion and decided to consolidate describe/info/list into single RPC method.

himank (approved):

PR #1001, opened 2023-04-04
feat: Order secondary indexes via field type

### LINKED ISSUE #976: Store field type information in the secondary index
We can have fields that do not have a strict type. This change allows supporting such fields.

garrensmith:
## Describe your changes
Add a field type to each index field so that fields in an index will be ordered by type. The ordering follows the same order as BSON:

MinKey (internal type)
Null
Numbers (ints, longs, doubles, decimals)
Symbol, String
Object
Array
BinData
ObjectId
Boolean - false then true
Date
Timestamp
Regular Expression
MaxKey (internal type)

All tests should pass

closes #976

### DIFF
--- a/schema/collection.go
+++ b/schema/collection.go
@@ -93,10 +93,6 @@ func (d *DefaultCollection) GetPrimaryKey() *Index {
 	return d.PrimaryKey
 }
 
-func (d *DefaultCollection) GetPrimaryQueryableFields() []*QueryableField {
-	return NewQueryableFieldsBuilder().BuildQueryableFields(d.PrimaryKey.Fields, nil)
-}
-
 func disableAdditionalPropertiesAndAllowNullable(required []string, properties map[string]*jsonschema.Schema) {
 	for name, p := range properties {
 		isRequired := false
--- a/schema/collection.go
+++ b/schema/collection.go
@@ -93,6 +93,10 @@ func (d *DefaultCollection) GetPrimaryKey() *Index {
 	return d.PrimaryKey
 }
 
+func (d *DefaultCollection) GetPrimaryQueryableFields() []*QueryableField {
+	return NewQueryableFieldsBuilder().BuildQueryableFields(d.PrimaryKey.Fields, nil)
+}
+
 func disableAdditionalPropertiesAndAllowNullable(required []string, properties map[string]*jsonschema.Schema) {
 	for name, p := range properties {
 		isRequired := false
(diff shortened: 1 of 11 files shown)


schema/collection.go
```diff
@@ -93,6 +93,10 @@ func (d *DefaultCollection) GetPrimaryKey() *Index {
 	return d.PrimaryKey
 }
 
+func (d *DefaultCollection) GetPrimaryQueryableFields() []*QueryableField {
```
himank, line 96: I am not seeing the usage of it, is it needed?

value/secondary_value_order.go
```diff
@@ -0,0 +1,75 @@
...
+	case schema.ObjectType:
+		return 30
+	case schema.BoolType:
+		if val != nil && val.String() == "true" {
+			return 36
+		}
+		return 35
+	case schema.DateTimeType:
+		return 40
+	case schema.MaxType:
```
himank, line 69: how will this be used? If type is not found can we default to MaxOrder?

garrensmith: Max Type is used for querying whenever there is no end range we use this

schema/collection.go
garrensmith: Ah yes. I was planning to remove the generics in key_builder and use this. I'll remove and add it later

value/secondary_value_order.go
himank: So they have exposed it as well?

value/secondary_value_order.go
```diff
@@ -0,0 +1,75 @@
...
+func ToSecondaryOrder(dataType schema.FieldType, val Value) int {
+	switch dataType {
+	case schema.NullType:
+		return 5
+	case schema.DoubleType, schema.Int32Type, schema.Int64Type:
+		return 10
+	case schema.StringType, schema.UUIDType:
+		return 20
+	case schema.ObjectType:
+		return 30
```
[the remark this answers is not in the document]
himank: Okay, may be then a gap of `5` to be consistent with the current gaps?

himank (dismissed):

value/secondary_value_order.go
garrensmith: Yes you can do  `{field: {"lte": new MaxKey()}` or `{field: {gte: new MinKey()}}`  in js

himank: Interesting.

himank (approved):

PR #1006, opened 2023-04-06
feat: Delete namespace implementation

JigarJoshi:
## Describe your changes
Allow cluster admins to delete namespace. This is required for internal clean up and this needs to be followed up with a restart of server for now as the memory state needs to be updated across all the nodes.

### DIFF
--- a/server/metadata/dictionary.go
+++ b/server/metadata/dictionary.go
@@ -172,6 +172,21 @@ func (r *reservedSubspace) reload(ctx context.Context, tx transaction.Tx) error 
 	return it.Err()
 }
 
+func (r *reservedSubspace) unReserveNamespace(ctx context.Context, tx transaction.Tx, namespaceId string) error {
+	if len(namespaceId) == 0 {
+		return errors.InvalidArgument("namespaceId is empty")
+	}
+
+	key := keys.NewKey(r.ReservedSubspaceName(), namespaceKey, namespaceId, keyEnd)
+
+	err := tx.Delete(ctx, key)
+
+	log.Debug().Err(err).Str("key", key.String()).Str("value", namespaceId).
+		Msg("un-reserving namespace")
+
+	return err
+}
+
 func (r *reservedSubspace) reserveNamespace(ctx context.Context, tx transaction.Tx, namespaceId string,
 	namespaceMetadata NamespaceMetadata,
 ) error {
@@ -309,6 +324,11 @@ func (k *Dictionary) ReserveNamespace(ctx context.Context, tx transaction.Tx, na
 	namespaceMetadata NamespaceMetadata,
 ) error {
 	return k.reservedSb.reserveNamespace(ctx, tx, namespaceId, namespaceMetadata)
+}
+
+// UnReserveNamespace deletes the namespace.
+func (k *Dictionary) UnReserveNamespace(ctx context.Context, tx transaction.Tx, namespaceId string) error {
+	return k.reservedSb.unReserveNamespace(ctx, tx, namespaceId)
 }
 
 func (k *Dictionary) GetNamespaces(ctx context.Context, tx transaction.Tx,
--- a/server/metadata/tenant.go
+++ b/server/metadata/tenant.go
@@ -234,13 +234,13 @@ func (m *TenantManager) GetNamespaceId(namespaceName string) (uint32, error) {
 
 // GetTenant is responsible for returning the tenant from the cache. If the tenant is not available in the cache then
 // this method will attempt to load it from the database and will update the tenant manager cache accordingly.
-func (m *TenantManager) GetTenant(ctx context.Context, namespaceName string) (*Tenant, error) {
+func (m *TenantManager) GetTenant(ctx context.Context, namespaceId string) (*Tenant, error) {
 	var (
 		tenant *Tenant
 		err    error
 	)
 
-	if tenant = m.getTenantFromCache(namespaceName); tenant != nil {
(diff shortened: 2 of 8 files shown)


ovaistariq (approved):

server/metadata/dictionary.go
```diff
@@ -172,6 +172,28 @@ func (r *reservedSubspace) reload(ctx context.Context, tx transaction.Tx) error
...
+func (r *reservedSubspace) unReserveNamespace(ctx context.Context, tx transaction.Tx, namespaceId string) error {
+	if len(namespaceId) == 0 {
+		return errors.InvalidArgument("namespaceId is empty")
+	}
+
+	if err := r.reload(ctx, tx); ulog.E(err) {
+		return err
+	}
+
+	r.RLock()
```
himank, line 184: This is not needed.

server/metadata/dictionary.go
```diff
@@ -172,6 +172,28 @@ func (r *reservedSubspace) reload(ctx context.Context, tx transaction.Tx) error
 	return it.Err()
 }
 
+func (r *reservedSubspace) unReserveNamespace(ctx context.Context, tx transaction.Tx, namespaceId string) error {
+	if len(namespaceId) == 0 {
+		return errors.InvalidArgument("namespaceId is empty")
+	}
+
+	if err := r.reload(ctx, tx); ulog.E(err) {
```
himank, line 180: No need to call reload for this.

server/metadata/tenant.go
```diff
@@ -1051,6 +1051,33 @@ func (tenant *Tenant) ListProjects(_ context.Context) []string {
...
+// This needs to be followed up with a "restart" of server to clear memory state.
+// Be careful calling this.
+func (m *TenantManager) DeleteTenant(ctx context.Context, tx transaction.Tx, tenantToDelete *Tenant) error {
+	tenantToDelete.RLock()
+	defer tenantToDelete.RUnlock()
+	projects := tenantToDelete.ListProjects(ctx)
+
+	deletedProjects := true
+	for _, project := range projects {
+		_, err := tenantToDelete.DeleteProject(ctx, tx, project)
```
himank, line 1064: `DeleteProject` needs a lock, so it will be a deadlock if you use `defer` above.

server/metadata/tenant.go
```diff
@@ -1051,6 +1051,33 @@ func (tenant *Tenant) ListProjects(_ context.Context) []string {
 	return projects
 }
 
+// DeleteTenant is used to delete tenant and all the content within it.
+// This needs to be followed up with a "restart" of server to clear memory state.
+// Be careful calling this.
+func (m *TenantManager) DeleteTenant(ctx context.Context, tx transaction.Tx, tenantToDelete *Tenant) error {
+	tenantToDelete.RLock()
+	defer tenantToDelete.RUnlock()
+	projects := tenantToDelete.ListProjects(ctx)
```
himank, line 1060: ListProjects is taking a read lock on the tenant, so you don't need to grab a lock at the start of this method.

server/metadata/tenant.go
```diff
@@ -1051,6 +1051,33 @@ func (tenant *Tenant) ListProjects(_ context.Context) []string {
...
+
+	deletedProjects := true
+	for _, project := range projects {
+		_, err := tenantToDelete.DeleteProject(ctx, tx, project)
+		if err != nil {
+			log.Error().Err(err).Msgf("Failed to delete project named %s", project)
+			deletedProjects = false
+			err = tx.Rollback(ctx)
+			if err != nil {
+				log.Error().Err(err).Msgf("Failed to rollback tx while deleting project named %s", project)
```
himank, line 1070: What is the advantage of `deletedProjects` flag? Should we rollback and return?

server/metadata/tenant.go
```diff
@@ -1051,6 +1051,33 @@ func (tenant *Tenant) ListProjects(_ context.Context) []string {
...
+	defer tenantToDelete.RUnlock()
+	projects := tenantToDelete.ListProjects(ctx)
+
+	deletedProjects := true
+	for _, project := range projects {
+		_, err := tenantToDelete.DeleteProject(ctx, tx, project)
+		if err != nil {
+			log.Error().Err(err).Msgf("Failed to delete project named %s", project)
+			deletedProjects = false
+			err = tx.Rollback(ctx)
```
himank, line 1068: Do we need to roll back here or just return an error and let the caller roll back as caller is responsible for starting the transaction?

server/metadata/dictionary.go
JigarJoshi: good catch. fixed.

server/metadata/tenant.go
JigarJoshi: good catch. fixed.

server/metadata/dictionary.go
JigarJoshi: good catch. fixed.

server/metadata/tenant.go
JigarJoshi: good catch. fixed.

server/metadata/tenant.go
JigarJoshi: it was carried over when this code was placed outside of tenant manager. fixed thanks.

server/metadata/tenant.go
JigarJoshi: you are right. This was copied code when this logic was placed within the caller. fixed it.

himank (approved):

PR #1012, opened 2023-04-07
[WIP] Sharded atomics

efirs:
On write increment random shard of N.
On read sum up the value from all the shards.

This will be used by statistics calculation, such as data size.

### DIFF
--- a/store/kv/atomic.go
+++ b/store/kv/atomic.go
@@ -1,0 +1,78 @@
+// Copyright 2022-2023 Tigris Data, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package kv
+
+import (
+	"context"
+	"math/rand"
+)
+
+const (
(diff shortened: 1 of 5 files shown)


store/kv/atomic.go
```diff
@@ -0,0 +1,78 @@
...
+	n := rand.Intn(s.numShards) //nolint:gosec
+
+	key = append(key, AtomicSuffixSeparator)
+	key = append(key, n)
+
+	return s.kv.AtomicAdd(ctx, table, key, inc)
+}
+
+// AtomicRead calculates the value by summing all the shards.
+func (s *shardedAtomics) AtomicRead(ctx context.Context, table []byte, key Key) (int64, error) {
```
himank, line 55: When we will be diffing and updating that can be done on any random shard, so we need one more API to randomly read from any shard.

store/kv/atomic.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"math/rand"
+)
+
+const (
+	DefaultNumShards      = 1024
+	AtomicSuffixSeparator = "__A__"
+)
+
+type ShardedAtomics interface {
```
himank, line 27: As we are introducing this interface, should we move `AtomicAdd`/`AtomicRead` from the kv interface to a `NonShardedAtomics` interface? Or remove them.

store/kv/atomic.go
efirs: Individual shard value is no better then picking a random number,
I don't see a value accessing it individually.

store/kv/atomic.go
efirs: Not a bad idea. I think how to do it properly.

store/kv/atomic.go
himank: Never mind, I was still thinking that we are not using atomics, and then we need to read in a transaction. We don't need the API to read from the random shard, this is good.

PR #1013, opened 2023-04-07
feat: Usage reporter

adilansari:
## Describe your changes
- Usage reporter for periodically publishing metrics to billing service

- Unit tests
- Running tigris locally with proper metronome config

## Prerequisites
- [x] https://github.com/tigrisdata/tigris-infra/pull/831

TIG-1204

### DIFF
--- a/server/main.go
+++ b/server/main.go
@@ -125,7 +125,7 @@ func mainWithCode() int {
 	mx.RegisterServices(&cfg.Server, kvStoreForDatabase, searchStore, tenantMgr, txMgr, forSearchTxMgr)
 
 	// metrics is already initialized, we can start reporting usage data
-	ur, err := billing.NewUsageReporter(metrics.GlobalSt, tenantMgr)
+	ur, err := billing.NewUsageReporter(metrics.GlobalSt, tenantMgr, billing.NewProvider())
 	if !ulog.E(err) {
 		ur.Start()
 	}
--- a/server/main.go
+++ b/server/main.go
@@ -27,13 +27,13 @@ import (
 	"github.com/tigrisdata/tigris/server/muxer"
 	"github.com/tigrisdata/tigris/server/quota"
 	"github.com/tigrisdata/tigris/server/request"
+	"github.com/tigrisdata/tigris/server/services/v1/billing"
 	"github.com/tigrisdata/tigris/server/tracing"
 	"github.com/tigrisdata/tigris/server/transaction"
 	"github.com/tigrisdata/tigris/store/kv"
 	"github.com/tigrisdata/tigris/store/search"
 	"github.com/tigrisdata/tigris/util"
 	ulog "github.com/tigrisdata/tigris/util/log"
-	"github.com/tigrisdata/tigris/server/services/v1/billing"
 )
 
 func main() {
--- a/server/main.go
+++ b/server/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/tigrisdata/tigris/store/search"
 	"github.com/tigrisdata/tigris/util"
 	ulog "github.com/tigrisdata/tigris/util/log"
+	"github.com/tigrisdata/tigris/server/services/v1/billing"
 )
 
 func main() {
@@ -122,6 +123,13 @@ func mainWithCode() int {
 
 	mx := muxer.NewMuxer(cfg)
 	mx.RegisterServices(&cfg.Server, kvStoreForDatabase, searchStore, tenantMgr, txMgr, forSearchTxMgr)
+
+	// metrics is already initialized, we can start reporting usage data
+	ur, err := billing.NewUsageReporter(metrics.GlobalSt, tenantMgr)
+	if !ulog.E(err) {
+		ur.Start()
+	}
+
 	port := cfg.Server.Port
 	if cfg.Server.Type == config.RealtimeServerType {
 		port = cfg.Server.RealtimePort
--- a/server/metadata/dictionary.go
+++ b/server/metadata/dictionary.go
@@ -251,7 +251,6 @@ func (r *reservedSubspace) updateNamespace(ctx context.Context, tx transaction.T
 	err = tx.Replace(ctx, key, internal.NewTableDataWithVersion(payload, nsMetaValueVersion), true)
 	log.Debug().Err(err).Str("key", key.String()).Uint32("value", namespaceMetadata.Id).Msg("updated namespace metadata")
 
-	// todo: Check for race condition
 	return r.reload(ctx, tx)
 }
 
--- a/server/metadata/dictionary.go
+++ b/server/metadata/dictionary.go
@@ -229,6 +229,32 @@ func (r *reservedSubspace) reserveNamespace(ctx context.Context, tx transaction.
 	return err
 }
 
+func (r *reservedSubspace) updateNamespace(ctx context.Context, tx transaction.Tx, namespaceMetadata NamespaceMetadata) error {
+	if len(namespaceMetadata.StrId) == 0 {
+		return errors.InvalidArgument("namespaceId is empty")
+	}
+	if namespaceMetadata.Id < 1 {
+		return errors.InvalidArgument("id should be greater than 0, received %d", namespaceMetadata.Id)
+	}
+	if err := r.reload(ctx, tx); ulog.E(err) {
+		return err
+	}
+	if _, ok := r.idToNamespaceStruct[namespaceMetadata.Id]; !ok {
+		return errors.NotFound("no namespace exists for id %s", namespaceMetadata.StrId)
+	}
+
+	key := keys.NewKey(r.ReservedSubspaceName(), namespaceKey, namespaceMetadata.StrId, keyEnd)
+	payload, err := jsoniter.Marshal(namespaceMetadata)
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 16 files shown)


server/main.go
```diff
@@ -122,6 +123,13 @@ func mainWithCode() int {
 
 	mx := muxer.NewMuxer(cfg)
 	mx.RegisterServices(&cfg.Server, kvStoreForDatabase, searchStore, tenantMgr, txMgr, forSearchTxMgr)
+
+	// metrics is already initialized, we can start reporting usage data
+	ur, err := billing.NewUsageReporter(metrics.GlobalSt, tenantMgr, billing.NewProvider())
+	if !ulog.E(err) {
+		ur.Start()
```
adilansari, line 130: starting the background worker in main

server/services/v1/billing/metronome.go
```diff
@@ -119,16 +119,31 @@ func (m *Metronome) pushBillingEvents(ctx context.Context, events []biller.Event
...
-	if resp.StatusCode() != http.StatusOK {
-		return errors.Internal("metronome failure: %s", resp.Body)
+	// todo: let page size be a const
+	pageSize := 100
+	pages := len(events) / pageSize
+	if len(events)%pageSize > 0 {
+		pages += 1
 	}
 
+	for p := 0; p < pages; p++ {
```
adilansari, line 129: Batching events before putting them on wire

server/metrics/global_status.go
```diff
@@ -29,24 +29,28 @@ var ignoredFieldsForWrites = [][]byte{
...
+}
+
 type GlobalStatus struct {
 	mu          sync.Mutex
 	activeChunk *TenantStatusTimeChunk
 }
 
 type TenantStatusTimeChunk struct {
-	startTime time.Time
+	StartTime time.Time
```
pboros, line 42: The idea is that these can only be written with a lock. Making these public loses that.

server/services/v1/billing/reporter.go
```diff
@@ -0,0 +1,127 @@
...
+	ctx        context.Context
+}
+
+func NewUsageReporter(gs metrics.UsageProvider, tm metadata.NamespaceMetadataMgr, billing Provider) (*UsageReporter, error) {
+	if gs == nil || tm == nil {
+		return nil, errors.Internal("usage reporter cannot be initialized")
+	}
+
+	return &UsageReporter{
+		glbState:   gs,
```
pboros, line 45: For the global status, there should be only one existing per server instance. Because there is only a global one, there is no need to pass it. Do we have a good way to handle server level singletons like that? This was one of the areas that I didn't like in the global status code. Metric providers have a similar problem.

pboros:
Have some minor comments/questions, overall it looks great!

server/main.go
```diff
@@ -122,6 +123,13 @@ func mainWithCode() int {
 
 	mx := muxer.NewMuxer(cfg)
 	mx.RegisterServices(&cfg.Server, kvStoreForDatabase, searchStore, tenantMgr, txMgr, forSearchTxMgr)
+
+	// metrics is already initialized, we can start reporting usage data
+	ur, err := billing.NewUsageReporter(metrics.GlobalSt, tenantMgr, billing.NewProvider())
```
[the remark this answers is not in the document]
adilansari: Its a pointer `GlobalSt       *GlobalStatus`

server/metrics/global_status.go
adilansari: Another option I can think of is `Flush()` returning a copy instead of the original object

server/services/v1/billing/reporter.go
adilansari: I don't think there is a framework as such that can do that. The `NewGlobalStatus()` method can do a nil check to only initialize `var GlobalSt *GlobalStatus` once.

server/metrics/global_status.go
pboros: Flush is returning a copy and zeroing out the original object while holding a lock. So if a request finishes in the meantime, we log the stats for those in the new structure, and the reporter will get it in the next flush.

server/services/v1/billing/reporter.go
himank: `sync.Once` can be used to guarantee that it is only initialized once https://pkg.go.dev/sync#Once

server/metadata/tenant.go
```diff
@@ -301,6 +316,39 @@ func (m *TenantManager) GetTenant(ctx context.Context, namespaceId string) (*Ten
...
+
+	if _, found := m.tenants[meta.StrId]; !found {
+		return errors.NotFound("no tenant found with id %s", meta.StrId)
+	}
+
+	defer func() {
+		if err == nil {
+			tenant := m.tenants[meta.StrId]
+			if err = tx.Commit(ctx); err == nil && tenant != nil {
+				tenant.namespace = NewTenantNamespace(meta.StrId, meta)
```
himank, line 341: Instead of this, should we just update inside the same tenant object?

server/services/v1/billing/reporter.go
```diff
@@ -0,0 +1,127 @@
...
+	if len(chunk.Tenants) == 0 {
+		log.Info().Msg("no tenant data to report")
+		return nil
+	}
+
+	trxnSuffix := strconv.FormatInt(chunk.EndTime.Unix(), 10)
+	events := make([]*UsageEvent, 0, len(chunk.Tenants))
+
+	for namespaceId, stats := range chunk.Tenants {
+		nsMeta := r.tenantMgr.GetNamespaceMetadata(r.ctx, namespaceId)
```
himank, line 80: Instead of this, you have an explicit method in tenantMgr that returns `accounts`, and that method will try to read from the tenant cache, and if it is empty, then try to read from the database. This will then avoid triggering the UpdateNamespaceMetadata below if the metadata is updated, but the worker view is stale.

server/metadata/dictionary.go
```diff
@@ -229,6 +229,32 @@ func (r *reservedSubspace) reserveNamespace(ctx context.Context, tx transaction.
...
+	key := keys.NewKey(r.ReservedSubspaceName(), namespaceKey, namespaceMetadata.StrId, keyEnd)
+	payload, err := jsoniter.Marshal(namespaceMetadata)
+	if err != nil {
+		return err
+	}
+	err = tx.Replace(ctx, key, internal.NewTableDataWithVersion(payload, nsMetaValueVersion), true)
+	log.Debug().Err(err).Str("key", key.String()).Uint32("value", namespaceMetadata.Id).Msg("updated namespace metadata")
+
+	// todo: Check for race condition
+	return r.reload(ctx, tx)
```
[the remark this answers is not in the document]
himank: As reload is reading from the database so, even if the transaction fails, it still only has the state of whatever the database has.

server/metadata/tenant.go
```diff
@@ -232,6 +238,49 @@ func (m *TenantManager) GetNamespaceId(namespaceName string) (uint32, error) {
...
+		// getTenant
+		tenant, err := m.GetTenant(ctx, namespaceId)
+		if err != nil {
+			log.Error().Err(err).Msgf("cannot find tenant with id %s", namespaceId)
+			continue
+		}
+		// update accounts if tenant does not have it
+		if tenant.namespace.Metadata().Accounts.Metronome == nil {
+			tenant.Lock()
+			tenant.namespace = NewTenantNamespace(namespaceId, metadata)
```
himank, line 267: just update the accounts?

server/metadata/tenant.go
```diff
@@ -301,6 +350,41 @@ func (m *TenantManager) GetTenant(ctx context.Context, namespaceId string) (*Ten
...
+	if _, found := m.tenants[meta.StrId]; !found {
+		return errors.NotFound("no tenant found with id %s", meta.StrId)
+	}
+
+	defer func() {
+		if err == nil {
+			tenant := m.tenants[meta.StrId]
+			if err = tx.Commit(ctx); err == nil && tenant != nil {
+				tenant.Lock()
+				tenant.namespace = NewTenantNamespace(meta.StrId, meta)
```
himank, line 376: Just update the accounts or the metadata?

server/services/v1/billing/metronome.go
```diff
@@ -119,16 +119,31 @@ func (m *Metronome) pushBillingEvents(ctx context.Context, events []biller.Event
...
+
+		page := events[p*pageSize : high]
+
+		// content encoding - gzip?
+		resp, err := m.client.IngestWithResponse(ctx, page)
+		if err != nil {
+			return err
+		}
+		if resp.StatusCode() != http.StatusOK {
+			return errors.Internal("metronome failure: %s", resp.Body)
```
himank, line 143: return `resp.StatusCode` as well along with the body

server/metadata/tenant.go
adilansari: `tenant.namespace` is an interface, we cannot replace the metadata:

https://github.com/tigrisdata/tigris/blob/bb8cad820de6c98b3456ebc3dd1bf35dbe6cee59/server/metadata/namespace.go#L27-L34

himank: We can have`SetMetadata(md NamespaceMetadata)` on it.

server/metrics/global_status.go
himank: There can be an interface here(`TenantTimeChunk`) that has getters, and then that interface can be mocked easily in the reporter.

himank (approved):

himank (approved):

PR #1020, opened 2023-04-12
[WIP] Statistics KV layer

efirs:


store/kv/kv_store.go
```diff
@@ -86,67 +95,101 @@ func (tx *KeyValueTx) ReadRange(ctx context.Context, table []byte, lkey Key, rke
...
+
 	return NewKeyValueIterator(ctx, iter), nil
 }
 
-type KeyValueIterator struct {
-	ctx context.Context
-	baseIterator
-	err error
-}
+func customMetadataDecoder(d *msgpack.Decoder) (any, error) {
```
himank, line 102: Can you move this to the internal package? As we have encoding/decoding logic on table data in that package.

PR #1025, opened 2023-04-13
Use proto for document metadata encoding

efirs:
This makes encoding ~3 times faster.
Decoding is made O(1), so no special metadata-only decoding API is needed.

This is forward incompatible change.


himank:
Very Cool!

PR #1026, opened 2023-04-13
Explicit schema versioning

efirs:
Allow zero-downtime new schema version rollout.
Or run two versions of the applications simultaneously. This change only enables compatible schema support.

There is no functional change if explicit versioning is not used.

Explicit schema version is a single number, which determines database schema state. This state of collections schema should be applied transactionally.

The schema version is set as a "version" field in the collection schema.

Versioning can be started from any number greater than maximum of implicit versions of all collections in the database.

Once schema versioning is started, version can be incremented sequentially only. Now holes are allowed.

Historical schema version is allowed to proceed if the schema body is the same as stored in the metadata store.


himank (approved):

ovaistariq:
This is not enabled by default, right? Because we need to think about how to prevent data loss due to incompatible schema change

efirs:
This not enabled by default, but also incompatible schema changes is not enabled.
This is only allows smooth compatible schema change rollout.

himank:
This enables pushing the schema version from the client instead of automatically generating it on the server during DDL. The schema support is still backward compatible changes only.

ovaistariq:
👍👍

PR #1037, opened 2023-04-14
feat: Return matched token for nested fields in case of array of objects

himank:
## Describe your changes
We were returning only the parent name(array of objects) in case the search matches that field. This diff will return an exact nested field that matches in the case of an array of objects. The convention will be the same `parent.nested`

### DIFF
--- a/server/search/hits.go
+++ b/server/search/hits.go
@@ -22,6 +22,10 @@ import (
 	tsApi "github.com/tigrisdata/typesense-go/typesense/api"
 )
 
+const (
+	matchedTokensKey = "matched_tokens"
+)
+
 type Hits struct {
 	hits  []*Hit
 	index int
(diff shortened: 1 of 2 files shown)


server/search/hits.go
```diff
@@ -112,3 +104,83 @@ func NewSearchHit(tsHit *tsApi.SearchResultHit) *Hit {
...
+					}
+
+					// now this means we can have an array of objects, we iterate one level
+					// to build matched fields
+					buildMatchedForArrObjects(name, mp, fields)
+				}
+			}
+		}
+	}
+}
```
efirs, line 152: newline missing?

ovaistariq (approved):

server/search/hits.go
himank: Saw it now, will fix it in next diff.

efirs: I wonder how it passed the lints.
We need to harden it more :)

himank: I don't think lint complains about the extra line? Yes, we need to add that check as well.

PR #1077, opened 2023-04-21
refactor: Always enable chunk store but honor the enable/disable flag during writes

himank:
## Describe your changes
ChunkStore will always be enabled, but whether we chunk the data or not will depend on the config flag. This allows us to decouple the reading of data because reading should be dependent on the attributes present inside the value, **not** on the store type. This means chunking can be disabled at any time.


server/config/options.go
```diff
@@ -505,6 +508,8 @@ type SchemaConfig struct {
 
 // FoundationDBConfig keeps FoundationDB configuration parameters.
 type FoundationDBConfig struct {
+	// Chunking allows us to persist bigger payload in storage.
+	Chunking    bool   `mapstructure:"chunking" yaml:"chunking" json:"chunking"`
```
efirs, line 512: I was thinking to add special KVConfig.

efirs: Where we can put chunking, compression, stats, which is actually independent from FDB.

efirs (approved):

server/config/options.go
himank: Makes sense, added with chunking.

efirs (approved):

Issue #873, opened 2023-02-27
Add support for `$ne` query comparison operator


AlekSi:
https://www.tigrisdata.com/docs/concepts/database/filters/

Currently, it is impossible to filter by a condition like `v != 42` because neither the "not equal" comparison operator nor the "not" logical operator is implemented.

roxanne-mackinnon:
I am currently working on this issue.

PR #1089, opened 2023-04-24
added support for $ne operator

roxanne-mackinnon:
## Describe your changes
Added support for $ne comparison operator in query/filter, to test not-equal
Added tests to query/filter/comparison_test.go
/claim #873 for algora bounty

### DIFF
--- a/server/metadata/tenant.go
+++ b/server/metadata/tenant.go
@@ -1506,7 +1506,7 @@ func (tenant *Tenant) updateCollection(ctx context.Context, tx transaction.Tx, d
 	return nil
 }
 
-// Old collections may not have any indexes stored in metadata. This will update that
+// Old collections may not have any indexes stored in metadata. This will update that.
 func (tenant *Tenant) UpgradeCollectionIndexes(ctx context.Context, tx transaction.Tx, db *Database, coll *schema.DefaultCollection) error {
 	builder := schema.NewFactoryBuilder(false)
 	schFactory, err := builder.Build(coll.Name, coll.Schema)
--- a/server/metrics/datadog.go
+++ b/server/metrics/datadog.go
@@ -112,23 +112,23 @@ func FormDatadogQueryNoMeta(namespace string, noMeta bool, req *api.QueryTimeSer
 	}
 
 	if config.GetEnvironment() != "" {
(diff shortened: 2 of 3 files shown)


himank:
Overall, it is the right direction. But change it like this,
- `$ne` will implement `LikeMatcher` interface
-  Then this matcher will be used by `LikeFilter,` which you'll add in the second case in `filter.go` at line 359.
The reason for doing it this way is that we don't want to implement keys from this filter and mainly only want to use it for filtering.

Once this is done, please add a test in `test/v1/server/document_test.go` where we have other Read API integration tests.

roxanne-mackinnon:
> Overall, it is the right direction. But change it like this,
> 
>     * `$ne` will implement `LikeMatcher` interface
> 
>     * Then this matcher will be used by `LikeFilter,` which you'll add in the second case in `filter.go` at line 359.
>       The reason for doing it this way is that we don't want to implement keys from this filter and mainly only want to use it for filtering.
> 
> 
> Once this is done, please add a test in `test/v1/server/document_test.go` where we have other Read API integration tests.

If I'm understanding correctly, do you mean that `$ne` should only implement LikeMatcher, or should it also implement ValueMatcher? Also, is the purpose of `$ne` to do exact string comparison, as opposed to `$not` which is more like 'does not contain'?

himank:
> > Overall, it is the right direction. But change it like this,
> 
> > 
> 
> >     * `$ne` will implement `LikeMatcher` interface
> 
> > 
> 
> >     * Then this matcher will be used by `LikeFilter,` which you'll add in the second case in `filter.go` at line 359.
> 
> >       The reason for doing it this way is that we don't want to implement keys from this filter and mainly only want to use it for filtering.
> 
> > 
> 
> > 
> 
> > Once this is done, please add a test in `test/v1/server/document_test.go` where we have other Read API integration tests.
> 
> 
> 
> If I'm understanding correctly, do you mean that `$ne` should only implement LikeMatcher, or should it also implement ValueMatcher? Also, is the purpose of `$ne` to do exact string comparison, as opposed to `$not` which is more like 'does not contain'?

That's right only LikeMatcher. Correct $ne is for exact not equality comparison.

himank:
Looks good, can you rebase it, and I'll merge it.

PR #1100, opened 2023-04-26
[Do not merge]

adilansari:
- demo where mocks are generated in a separate path and not alongside interface


adilansari:
Problem:

```shell
$ ➔  go test github.com/tigrisdata/tigris/server/services/v1/billing

# github.com/tigrisdata/tigris/server/services/v1/billing
package github.com/tigrisdata/tigris/server/services/v1/billing
        imports github.com/tigrisdata/tigris/mocks/server/services/v1/billing: import cycle not allowed in test
FAIL    github.com/tigrisdata/tigris/server/services/v1/billing [setup failed]
FAIL
```

PR #1118, opened 2023-05-01
feat: log the number of documents for a slow call

pboros:
## Describe your changes

Log the number of documents for slow calls

Unit tests

TIG-1406

### DIFF
--- a/server/metrics/measurement.go
+++ b/server/metrics/measurement.go
@@ -60,6 +60,7 @@ type Measurement struct {
 	startedAt       time.Time
 	stoppedAt       time.Time
 	projectCollTags map[string]string
+	nDocs           int64
 }
 
 type MeasurementCtxKey struct{}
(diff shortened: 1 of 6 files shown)


server/metrics/measurement.go
```diff
@@ -91,6 +93,10 @@ func (m *Measurement) IncrementCount(scope tally.Scope, tags map[string]string,
 	scope.Tagged(tags).Counter(counterName).Inc(count)
 }
 
+func (m *Measurement) SetNDocs(value int64) {
+	m.nDocs += value
```
himank, line 97: is it thread safe?

store/kv/measure.go
```diff
@@ -71,6 +71,7 @@ func (i *KeyValueIteratorWithMetrics) Next(value *KeyValue) bool {
 
 	if reqStatus != nil && value.Data != nil {
 		reqStatus.AddReadBytes(int64(value.Data.Size()))
+		reqStatus.AddResultDocs(1)
```
himank, line 74: Can we do something similar for Read API as well?

server/metrics/measurement.go
pboros: Yes. A measurement is created at the beginning of the request and the scope of one measurement struct is the request.

himank (approved):

PR #1121, opened 2023-05-02
Allow user to explicitly specify fields as nullable

efirs:
* This fixes external data validation using our schema.
* This would allow proper reconstruction of language schema.
* We would not need to cleanup nulls during import process. Just specify additional "null" as type.

### DIFF
--- a/schema/factory_builder.go
+++ b/schema/factory_builder.go
@@ -16,6 +16,7 @@ package schema
 
 import (
 	"bytes"
+	"strings"
 
 	"github.com/buger/jsonparser"
 	jsoniter "github.com/json-iterator/go"
@@ -51,7 +52,7 @@ func (fb *FactoryBuilder) deserializeArray(items *FieldBuilder, current *[]*Fiel
 
 	// object array type
 	if len(items.Properties) > 0 {
-		if items.Type != jsonSpecObject {
+		if items.JSONType() != jsonSpecObject {
 			return errors.InvalidArgument("Properties only allowed for object type")
 		}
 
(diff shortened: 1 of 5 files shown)


test/v1/server/collection_test.go
```diff
@@ -18,6 +18,7 @@ package server
 
 import (
 	"fmt"
+	"github.com/tigrisdata/tigris/schema"
```
himank, line 21: goimports

schema/fields.go
```diff
@@ -413,10 +417,66 @@ func FindIndex(indexes []*Index, name string) *Index {
...
+type FieldMultiType struct {
+	Type       []string
+	isNullable bool
+}
+
+func (f *FieldMultiType) append(val string) {
+	if val == "null" {
+		f.isNullable = true
+	} else {
+		f.Type = append(f.Type, val)
```
himank, line 429: Type can only be `1` as you have `isNullable` property to support "null". Then do we need to have `Type` as an array, or can it be just a string?

schema/factory_builder.go
```diff
@@ -96,10 +97,14 @@ func (fb *FactoryBuilder) deserializeProperties(properties jsoniter.RawMessage,
 		dec := jsoniter.NewDecoder(bytes.NewReader(v))
 		dec.UseNumber()
 		if err = dec.Decode(&builder); err != nil {
+			if strings.Contains(err.Error(), ErrInvalidType.Error()) ||
```
himank, line 100: do we need contains as you're returning an exact error?

efirs: I tried errors.Is and it doesn't work by some reason.
The exact error looks like this:
```
schema.FieldBuilder.TypeRaw: unmarshalerDecoder: only one non-null type allowed for the field, error found in #10 byte of ...|,\"object\"]}|..., bigger context ...|{\"type\":[\"null\",\"string\",\"object\"]}|...
```

himank: Actually yes, the unmarshaler will change the error.

himank (approved):

Issue #893, opened 2023-03-03
Improve the experience of working with nested arrays


garrensmith:
If I have a model like this:

```ts
export class Comment {
  @Field()
  content: string;

  @Field()
  name: string;

  @Field({ timestamp: "createdAt" })
  createdAt?: Date;
}

@TigrisCollection("post")
export class Post {
  @PrimaryKey(TigrisDataTypes.UUID, { order: 1, autoGenerate: true })
  id?: string;

  @Field({ timestamp: "createdAt" })
  createdAt?: Date;

  @Field({ timestamp: "updatedAt" })
  updatedAt?: Date;

  @Field()
  title: string;

  @Field()
  content?: string;

  @Field({ default: false })
  published?: boolean;

  @Field(TigrisDataTypes.ARRAY, { elements: Comment })
  comments: Array<Comment>;
}
```

If I want to add another comment to the post document, I have to read the whole document and write it back:

```ts
const post = await postCollection.findOne({
    filter: { title: "my blog post" },
  });

post.comments.push({
    content: "nice post",
    name: "bob",
  });

 postCollection.insertOrReplaceOne(post)

```

I think it would be much nicer to have something similar to [MongoDB $push](https://www.mongodb.com/docs/manual/reference/operator/update/push/)

```ts
postCollection.updateOne({
    filter: { id: post.id },
    fields: {$push: {comments: {content: "nice", name: "bob"}}},
  });

```

akhill10:
Can I contribute to this? 
- As a starting point we can support inserting a single element and later add modifiers like $each, $sort and $slice on this $push field operator ( or should we do that along with this $push ? ) 

Proposed Solution:
1. Validate and Mutate Payload: If the request contains a push field operator, then mutate and validate the payload with `newInsertPayloadMutator` in the UpdateQueryRunner.
2. Add $push field operator: `MergeAndGet` in FieldOperatorFactory contains this push option as well, where we append this new element into the existing array and also check for primaryKeyMutation.  If the field is absent, then add the array field with the value as its only element.

ovaistariq:
@akhill10 this would be a good start.

himank:
@akhill10 You only need to do (2) i.e. 
  - In `MergeAndGet,` check if this operator is present in `FieldOperators`
  - if yes, then the factory will have a method called `Push` (factory.Push) where we will have a logic of appending to an array, and if the array is empty, then an array with this element only. 

The syntax of the operator to start with would be,
[...]

PR #1131, opened 2023-05-04
[WIP] feat: push field operator for Update

akhill10:
## Describe your changes
- Added $push operator field for Update

#893

### DIFF
--- a/query/update/operator.go
+++ b/query/update/operator.go
@@ -36,6 +36,7 @@ const (
 	Decrement FieldOPType = "$decrement"
 	Multiply  FieldOPType = "$multiply"
 	Divide    FieldOPType = "$divide"
+	Push      FieldOPType = "$push"
 )
 
 // BuildFieldOperators un-marshals request "fields" present in the Update API and returns a FieldOperatorFactory
@@ -46,7 +47,6 @@ func BuildFieldOperators(reqFields []byte) (*FieldOperatorFactory, error) {
 	if err := jsoniter.Unmarshal(reqFields, &decodedOperators); log.E(err) {
 		return nil, err
 	}
-
 	operators := make(map[string]*FieldOperator)
 	for op, val := range decodedOperators {
 		switch op {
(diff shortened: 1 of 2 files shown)


server/services/v1/database/query_runner.go
```diff
@@ -218,6 +254,14 @@ func (runner *UpdateQueryRunner) Run(ctx context.Context, tx transaction.Tx, ten
 		}
 	}
 
+	if fieldOperator, ok := factory.FieldOperators[string(update.Push)]; ok {
+		// mutate if push payload needs set default values and also need to convert numeric fields from string to int64
```
himank, line 258: We can safely assume we don't need to set defaults inside an Array. This will make this operation less expensive.

akhill10: Okay, reverted the change. Not setting default values inside an array.

server/services/v1/database/query_runner.go
```diff
@@ -178,6 +178,37 @@ func updateDefaultsAndSchema(db string, branch string, collection *schema.Defaul
...
+	}
+
+	// converting a single element to array for mutation.
+	for key, value := range deserializedDoc {
+		var newValue []any
+		newValue = append(newValue, value)
+		deserializedDoc[key] = newValue
+	}
+
+	mutator := newInsertPayloadMutator(coll, ts.ToRFC3339())
```
himank, line 195: As this is for update only, you can create a `updateMutator`.

server/services/v1/database/query_runner.go
```diff
@@ -178,6 +178,37 @@ func updateDefaultsAndSchema(db string, branch string, collection *schema.Defaul
...
+
+	// this will mutate map, so we need to serialize this map again
+	if err = mutator.stringToInt64(deserializedDoc); err != nil {
+		return doc, err
+	}
+
+	if mutator.isMutated() {
```
himank, line 202: just above this, we need to first validate schema as well like this,
```
	if request.NeedSchemaValidation(ctx) {
		if err = coll.Validate(deserializedDoc); err != nil {
			// schema validation failed
			return doc, err
		}
	}
```

himank:
Looking great, just have a couple more comments.

PR #1165, opened 2023-05-10
feat: Index metadata toggle and write mode index

garrensmith:
## Describe your changes
Configure secondary indexes to not index the timestamp metadata. Configure the secondary_indexer to only index indexes that are not in active mode. This is for the background worker to reduce the number of indexes being written to.

All tests should pass

### DIFF
--- a/server/services/v1/database/secondary_indexer.go
+++ b/server/services/v1/database/secondary_indexer.go
@@ -138,16 +138,19 @@ type SecondaryIndexerImpl struct {
 	// Used for unit tests because it can index deeper into a schema than we currently expose to the user
 	// This can be removed once indexing into arrays and objects is exposed to the user
 	indexAll bool
+	// only index fields that are not active. For background indexing
+	indexWriteModeOnly bool
 	// Spare indexes do not index missing fields
 	sparse bool
 }
 
-func newSecondaryIndexerImpl(coll *schema.DefaultCollection) *SecondaryIndexerImpl {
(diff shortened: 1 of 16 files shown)


ovaistariq (approved):

server/services/v1/database/secondary_indexer.go
```diff
@@ -440,20 +443,26 @@ func (q *SecondaryIndexerImpl) buildTableRows(tableData *internal.TableData) ([]
...
 		},
 		{
 			tableData.UpdatedAt,
 			schema.UpdatedAt,
+			true,
 		},
 	}
 
 	rows := make([]IndexRow, 0, len(timeStamps))
+	if !q.coll.SecondaryIndexMetadata() {
```
himank, line 463: Can we assume now that Replace API doesn't need to read anything if there is no index defined by the user and the index metadata is false?

server/services/v1/database/secondary_indexer.go
```diff
@@ -463,6 +472,20 @@ func (q *SecondaryIndexerImpl) buildTSRows(tableData *internal.TableData) ([]Ind
 			dt = schema.DateTimeType
 		}
 
+		if q.indexWriteModeOnly {
+			for _, idx := range q.coll.SecondaryIndexes.All {
+				if idx.Name == schema.ReservedFields[ts.field] {
+					if idx.State == schema.INDEX_ACTIVE {
+						ts.shouldIndex = false
```
himank, line 479: I am not following this logic, can you add some comments?

himank (approved):

himank (dismissed):

himank (approved):

PR #1166, opened 2023-05-10
feat: Implement reverse order(descending sort) with chunking iterator

himank:
## Describe your changes
As we also support returning results in descending order, we need to implement that functionality in the chunking iterator to read in reverse, merge the chunks and return.


efirs (approved):

efirs:
Looks very cool!

efirs (approved):

PR #1190, opened 2023-05-15
fix: check metadata version before reads

garrensmith:
## Describe your changes
Check the metadata version before reads in case an index schema has been updated.


server/services/v1/api.go
```diff
@@ -344,7 +344,9 @@ func (s *apiService) Read(r *api.ReadRequest, stream api.Tigris_ReadServer) erro
 			InstantVerTracking: true,
 		})
 	} else {
-		_, err = s.sessions.ReadOnlyExecute(stream.Context(), s.runnerFactory.GetStreamingQueryRunner(r, stream, &queryMetrics, accessToken), database.ReqOptions{})
+		_, err = s.sessions.ReadOnlyExecute(stream.Context(), s.runnerFactory.GetStreamingQueryRunner(r, stream, &queryMetrics, accessToken), database.ReqOptions{
+			InstantVerTracking: true,
```
himank, line 348: `ReadOnlyExecute` is already doing that, are you seeing any issue?

PR #1193, opened 2023-05-15
refactor: Only index that are needed in indexing store

himank:
## Describe your changes
We only index fields in the indexing store that are tagged in the model

### DIFF
--- a/schema/search.go
+++ b/schema/search.go
@@ -458,25 +458,12 @@ func (s *ImplicitSearchIndex) buildSearchSchema(searchStoreName string) {
 	tsFields := make([]tsApi.Field, 0, len(s.QueryableFields))
 
 	for _, f := range s.QueryableFields {
-		// the implicit search index by default index all the fields that are indexable and same applies to facet/sort.
-		shouldIndex := SupportedSearchIndexableType(f.DataType, f.SubType)
-		shouldFacet := DefaultFacetableType(f.DataType)
-		shouldSort := DefaultSortableType(f.DataType)
-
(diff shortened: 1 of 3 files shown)


garrensmith (changes requested):
This will crash collection creation. If a user creates a schema with no search fields. Then in `tenant.go` where we create the search collection it will throw an error. 

The best way to test this is to create a collection with no search fields and then update the collection with a field that needs to search indexed.

schema/search.go
```diff
@@ -579,9 +556,9 @@ func (s *ImplicitSearchIndex) GetSearchDeltaFields(existingFields []*QueryableFi
 		tsFields = append(tsFields, tsApi.Field{
```
garrensmith, line 556: We adding to the tsFields here but all the values would be `false` if we have no search fields. What would then happen for typesense? 
I was testing this on the background search branch and what I did was not add the field to the `tsFields` if there was no search field But this caused the crashes in collection creation

himank: All the `fields` false and `optional` true mean storing in search but not indexing. This is needed so we don't need to read from the database when we get search results. You should not change that because that will break the search as then search results won't have full data.

garrensmith (approved):
Looks good. It might be worth checking if we have an integration test with no search fields just to make sure this doesn't throw errors

himank:
> Looks good. It might be worth checking if we have an integration test with no search fields just to make sure this doesn't throw errors

Yes, many of the collection tests are not adding a search index.

PR #1202, opened 2023-05-17
fix: catch background worker panics

garrensmith:
## Describe your changes
Catch panics from background workers

### DIFF
--- a/server/services/v1/workers/worker.go
+++ b/server/services/v1/workers/worker.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"runtime/debug"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


server/services/v1/workers/worker.go
```diff
@@ -101,10 +102,22 @@ func (w *Worker) jitterSleep() {
 }
 
 func (w *Worker) Start() {
+	defer catchWorkerPanics()
 	w.jitterSleep()
 	w.Loop()
 }
 
+func catchWorkerPanics() {
```
himank, line 110: This is good to have, but is it still happening? Or is this just in case to catch any bugs in the future?

himank (approved):

server/services/v1/workers/worker.go
garrensmith: This is defensive to catch any bugs in the future.

server/services/v1/workers/worker.go
```diff
@@ -101,10 +102,22 @@ func (w *Worker) jitterSleep() {
...
+	if r := recover(); r != nil {
+		stackTrace := string(debug.Stack())
+		if err, ok := r.(error); ok {
+			log.Err(err).Str("stacktrace", stackTrace).Msg("worker panic")
+		} else {
+			log.Error().Str("stacktrack", stackTrace).Msgf("worker panic %v", err)
```
pboros, line 116: Send the error as a separate argument with Err(), so it's searchable, rather than putting it in the message as text.

garrensmith: @pboros I do that above if the value from recover can be converted into an error. Otherwise we have to put it in the msg

pboros: Ah ok, makes sense.

pboros (approved):

ovaistariq (approved):

PR #1211, opened 2023-05-18
chore: Add configurable log for collection search filter

himank:
## Describe your changes
Add configurable log for collection search filter

### DIFF
--- a/server/services/v1/database/search_runner.go
+++ b/server/services/v1/database/search_runner.go
@@ -18,12 +18,14 @@ import (
 	"context"
 	"math"
 
+	"github.com/rs/zerolog/log"
 	api "github.com/tigrisdata/tigris/api/server/v1"
 	"github.com/tigrisdata/tigris/errors"
 	"github.com/tigrisdata/tigris/query/filter"
 	"github.com/tigrisdata/tigris/query/read"
 	qsearch "github.com/tigrisdata/tigris/query/search"
 	"github.com/tigrisdata/tigris/schema"
+	"github.com/tigrisdata/tigris/server/config"
 	"github.com/tigrisdata/tigris/server/metadata"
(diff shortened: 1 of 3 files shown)


server/services/v1/database/search_runner.go
```diff
@@ -63,6 +65,14 @@ func (runner *SearchQueryRunner) ReadOnly(ctx context.Context, tenant *metadata.
 	if err != nil {
 		return Response{}, ctx, err
 	}
+	if config.DefaultConfig.Search.LogFilter {
+		log.Error().
```
ovaistariq, line 69: Wouldn't this be info log?

JigarJoshi: I wish Go's logging libraries were as mature as Java ;) 's logging libraries.

JigarJoshi (approved):

server/services/v1/database/search_runner.go
```diff
@@ -63,6 +65,14 @@ func (runner *SearchQueryRunner) ReadOnly(ctx context.Context, tenant *metadata.
...
+			Str("tenant", tenant.Name()).
+			Str("project", db.Name()).
+			Str("collection", collection.Name).
+			Str("filter", string(runner.req.Filter)).
```
garrensmith, line 73: We have to be careful with logging users filters. This could become a compliance issue.

server/services/v1/database/search_runner.go
himank: The log level is currently set as `Error` to reduce the info logs. We need to change that to Warn and move logs like these to War as the info needs more cleanup.

server/services/v1/database/search_runner.go
himank: This is a temporary log to know if the user is using collection search with non-indexed filters.

PR #1212, opened 2023-05-18
feat: more accurate unit metrics

pboros:
## Describe your changes

* Read unit should not be counted if the read source is an internal metadata operation
* Count written search bytes and search write units
* If read api falls back to search, it should emit search units instead of read units

Check billing metrics after deployment, tested locally.

TIG-1528


garrensmith (approved):

garrensmith (approved):

server/metadata/key_generator.go
```diff
@@ -16,6 +16,7 @@ package metadata
 
 import (
 	"context"
+	"github.com/tigrisdata/tigris/server/metrics"
```
himank, line 19: goimports.

PR #1214, opened 2023-05-18
fix: Count single write unit for each search deletes, minimum threshold for compression configurable, and fixing go imports

himank:
## Describe your changes
- Count single write for search deletes
- Make minimum threshold for compression configurable
- Fixing go imports

### DIFF
--- a/server/services/v1/database/search_reader.go
+++ b/server/services/v1/database/search_reader.go
@@ -16,12 +16,12 @@ package database
 
 import (
 	"context"
-	"github.com/tigrisdata/tigris/server/metrics"
 
 	api "github.com/tigrisdata/tigris/api/server/v1"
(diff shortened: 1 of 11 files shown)


server/services/v1/database/search_reader.go
```diff
@@ -16,12 +16,13 @@ package database
 
 import (
 	"context"
-	"github.com/tigrisdata/tigris/server/metrics"
+
```
efirs, line 20: Is it two empty lines?

himank: Okay, make test only shifted the imports but didn't remove that line.

efirs (approved):

pboros (approved):

PR #1239, opened 2023-05-24
fix: Do not reload metadata if nothing changed in describeCollection

himank:
## Describe your changes
Do not reload metadata if nothing changed in describeCollection

### DIFF
--- a/server/transaction/manager.go
+++ b/server/transaction/manager.go
@@ -59,18 +59,41 @@ type Tx interface {
 	Rollback(ctx context.Context) error
 }
 
+type stagedDatabase interface {
+	IsMetadataChange() bool
+}
+
 // SessionCtx is used to store any baggage for the lifetime of the transaction. We use it to stage the database inside
 // a transaction when the transaction is performing any DDLs.
 type SessionCtx struct {
-	db any
+	db                    stagedDatabase
+	noMetadataChanged     bool
+	metadataChangeSession bool
 }
 
-func (c *SessionCtx) StageDatabase(db any) {
+func (c *SessionCtx) MetadataChangeSession(mcs bool) {
+	c.metadataChangeSession = mcs
+}
+
+func (c *SessionCtx) StageDatabase(db stagedDatabase) {
 	c.db = db
 }
 
-func (c *SessionCtx) GetStagedDatabase() any {
+func (c *SessionCtx) GetStagedDatabase() stagedDatabase {
 	return c.db
+}
+
+func (c *SessionCtx) MarkNoMetadataStateChanged() {
+	if c.db != nil {
+		// possible an interactive transaction
+		return
+	}
+
+	c.noMetadataChanged = true
+}
+
+func (c *SessionCtx) IsMetadataStateChanged() bool {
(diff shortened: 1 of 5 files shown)


ovaistariq (approved):

server/transaction/manager.go
```diff
@@ -59,20 +59,44 @@ type Tx interface {
...
 }
 
-func (c *SessionCtx) GetStagedDatabase() any {
+func (c *SessionCtx) GetStagedDatabase() stagedDatabase {
 	return c.db
 }
 
+func (c *SessionCtx) MarkNoMetadataStateChanged() {
+	ptrFalse := false
+	c.noMetadataChanged = &ptrFalse
```
efirs, line 88: Why this need to be a pointer?

server/transaction/manager.go
```diff
@@ -59,20 +59,44 @@ type Tx interface {
...
+}
+
+func (c *SessionCtx) IsMetadataStateChanged() bool {
+	incrMetadataVersion := (c.db == nil && c.metadataChangeSession) || (c.db != nil && c.db.IsMetadataChange())
+	if c.noMetadataChanged != nil {
+		incrMetadataVersion = *c.noMetadataChanged
```
efirs, line 94: noMetadataChanged == true will return incrMetadataVersion == true?

server/transaction/manager.go
```diff
@@ -59,20 +59,44 @@ type Tx interface {
...
+	IsMetadataChange() bool
+}
+
 // SessionCtx is used to store any baggage for the lifetime of the transaction. We use it to stage the database inside
 // a transaction when the transaction is performing any DDLs.
 type SessionCtx struct {
-	db any
+	db                    stagedDatabase
+	noMetadataChanged     *bool
+	metadataChangeSession bool
```
efirs, line 71: Why we need two variables?

server/transaction/manager.go
himank: it should never be true but let me add a guard against it.

server/transaction/manager.go
himank: `metadataChangeSession` is whether the session is requesting metadata change, I have added it here so that logic of checking whether we need to increment version or not will be at one place.

server/transaction/manager.go
efirs: Why we cannot just set c.metadataChangeSession to true here?

himank: What will the benefit of setting it to true?

server/transaction/manager.go
```diff
@@ -59,20 +59,48 @@ type Tx interface {
...
+	if c.db != nil {
+		// possible an interactive transaction
+		return
+	}
+
+	c.noMetadataChanged = true
+}
+
+func (c *SessionCtx) IsMetadataStateChanged() bool {
+	incrMetadataVersion := (c.db == nil && c.metadataChangeSession) || (c.db != nil && c.db.IsMetadataChange())
```
efirs, line 96: return ((c.db == nil && c.metadataChangeSession) || (c.db != nil && c.db.IsMetadataChange())) && !c.noMetadataChanged

efirs (approved):

PR #1241, opened 2023-05-26
refactor: refactor metronome metrics with consistent tags

pboros:
## Describe your changes

Refactor metronome metrics to be in-line with the rest and make sure tags won't be different in the same scope.
Right now the new metrics are additional to the old, one, needs to be deleted.

Check the metronome alerts in dev.

TIG-1549

### DIFF
--- a/server/metrics/metrics.go
+++ b/server/metrics/metrics.go
@@ -16,6 +16,7 @@ package metrics
 
 import (
 	"io"
+	"sync"
 	"time"
 
 	"github.com/rs/zerolog/log"
@@ -41,6 +42,7 @@ var (
 	SchemaMetrics         tally.Scope
 	MetronomeMetrics      tally.Scope
 	GlobalSt              *GlobalStatus
+	once                  sync.Once
 )
 
 func getVersion() string {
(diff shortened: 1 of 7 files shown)


server/metrics/metrics.go
```diff
@@ -93,6 +94,9 @@ func SchemaUpdateRepaired(project string, branch string, collection string) {
 }
 
 func InitializeMetrics() func() {
```
himank, line 96: How can this called twice?

server/metrics/metronome.go
```diff
@@ -15,44 +15,50 @@
...
-	MetronomeListInvoices = MetronomeMetrics.SubScope("list_invoices")
-	MetronomeGetInvoice = MetronomeMetrics.SubScope("get_invoice")
-	MetronomeGetUsage = MetronomeMetrics.SubScope("get_usage")
-	MetronomeGetCustomer = MetronomeMetrics.SubScope("get_customer")
 }
 
-func GetResponseCodeTags(code int) map[string]string {
+func getMetronomeTagKeys() []string {
+	return []string{
+		"env",
```
adilansari, line 42: do we need env?

server/metrics/metrics.go
```diff
@@ -93,6 +94,9 @@ func SchemaUpdateRepaired(project string, branch string, collection string) {
...
+	if metricsInitialized {
```
himank, line 97: What about instead of this do,
```
var once sync.Once

InitializeMetrics() {
    once.Do(func() {})
}
```

server/metrics/metronome.go
```diff
@@ -15,44 +15,50 @@
 package metrics
 
 import (
+	"github.com/tigrisdata/tigris/server/config"
```
himank, line 18: goimports

server/metrics/metronome.go
pboros: Yes, and it it's not here, it will be stripped by standardizeTags

server/metrics/metrics.go
pboros: I added a mutex, which is basically the same as what this does internally. We need to pass `io.Closer` outside.

himank: You can have io.Closer outside of `Do()`

himank (approved):

himank (approved):

PR #1251, opened 2023-06-01
fix: Break up tenants reloading on independent transactions

himank:
## Describe your changes
A single transaction is hitting the 5-second limit now, so we need to split reload on independent transactions.

### DIFF
--- a/server/services/v1/api.go
+++ b/server/services/v1/api.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/fullstorydev/grpchan/inprocgrpc"
 	"github.com/go-chi/chi/v5"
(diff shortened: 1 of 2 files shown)


server/services/v1/api.go
```diff
@@ -77,17 +78,18 @@ func newApiService(kv kv.TxStore, searchStore search.Store, tenantMgr *metadata.
...
 	}
 
+	var err error
 	ctx := context.TODO()
-	tx, err := u.txMgr.StartTx(ctx)
-	if ulog.E(err) {
-		log.Fatal().Err(err).Msgf("error starting server: starting transaction failed")
+	for i := 0; i < 3; i++ {
+		if err = tenantMgr.Reload(ctx); ulog.E(err) {
+			time.Sleep(100 * time.Second)
```
ovaistariq, line 85: Should we reduce the sleep duration to may be 30 seconds

ovaistariq (approved):

pboros (dismissed):

rbarabas (approved):
LGTM, per tenant reload trx

rbarabas (approved):
LGTM, per tenant reload trx

efirs (approved):

PR #1261, opened 2023-06-05
fix: Do not reload metadata on search delete index API

efirs:


server/services/v1/search/session.go
```diff
@@ -97,8 +98,10 @@ func (sessions *SessionManager) TxExecute(ctx context.Context, runner TxRunner,
 		return Response{}, errors.NotFound("tenant '%s' not found", namespace)
 	}
 
-	if err = sessions.TrackVersion(ctx, tenant); err != nil {
-		return Response{}, err
+	if !option.DoNotReloadMetadata {
+		if err = sessions.TrackVersion(ctx, tenant); err != nil {
```
adilansari, line 102: Should we reload by default and put skip behind flag?

adilansari (dismissed):

server/services/v1/search/session.go
efirs: Fixed.

server/services/v1/search.go
```diff
@@ -120,7 +121,8 @@ func (s *searchService) DeleteIndex(ctx context.Context, req *api.DeleteIndexReq
 	runner.SetDeleteIndexReq(req)
 
 	resp, err := s.sessions.TxExecute(ctx, runner, search.SessionOptions{
-		IncVersion: true,
+		IncVersion:          false,
```
himank, line 124: We need to make code change for this. Should we just have a single flag for this?

himank (approved):

PR #1269, opened 2023-06-08
fix: Break the tenant reload transaction further

JigarJoshi:
## Describe your changes
Break the tenant reload transaction further. batch 10 search index per transaction.

Test it on dev.


rbarabas (approved):

himank:
> reload() closing tx internally has caused lots of trouble in all parts of the code.

Yes, we need to break the reload further at a more granular level

PR #1279, opened 2023-06-20
fix: removing test search key

adilansari:


server/config/options.go
```diff
@@ -388,7 +388,6 @@ var DefaultConfig = Config{
 		Compression:       false,
 		IgnoreExtraFields: false,
 		LogFilter:         false,
-		AuthKey:           "ts_test_key",
```
pboros, line 391: Will this break tests?

adilansari: No. It was added accidentally in previous commit.

pboros (approved):

